PR #4, opened 2023-06-15
Initial Goal domain model. Simple POKOs with no business logic

mattwills-moj-digital:


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/goal/Step.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import java.time.LocalDate
+import java.util.*
+
+data class Step(
+  val reference: UUID,
+  val title: String,
+  val targetDate: LocalDate,
+  val status: StepStatus,
+  val number: Int,
```
nathanrussell-moj-digital, line 11: `sequenceNumber` ?

nathanrussell-moj-digital (commented):
A few minor comments/questions, but looking good 👍 

I wonder also if, given that the "domain" is the most important thing (above all else arguably), we should add some doc somewhere? I'm not thinking we go overboard, but perhaps some class level kdoc on the "root" class for a given area of the domain (`Goal` in this case), that just explains what a Goal is and what it comprises perhaps ?
I'm just thinking of other devs as and when they work in our codebase 🤔

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/goal/Goal.kt
```diff
@@ -0,0 +1,19 @@
...
+import java.time.LocalDate
+import java.util.UUID
+
+data class Goal(
+  val reference: UUID,
+  val title: String,
+  val reviewDate: LocalDate,
+  val type: GoalType,
+  val steps: List<Step>,
+  val status: GoalStatus,
```
nathanrussell-moj-digital, line 13: Not for this PR, but I think this type of thing will come out as we work through the dev and establish our patterns in this respect, but I'm now back to wondering if the all args constructor should be private, and we have other public constructors that provide values for certain fields (like we played around with yesterday)
I'm specifically thinking of `status` ,,,, when would I want to create (from a business domain point of view) a `Goal` with a status other than `NOT_STARTED` ?

I guess this will come out when we start thinking about the domain service class, and who/where we expect a `Goal` to be created 🤔

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/goal/Goal.kt
```diff
@@ -0,0 +1,19 @@
...
+  val notes: String?,
+  val createdBy: String,
+  val createdDate: Instant,
```
nathanrussell-moj-digital, line 16: `createdDate` (and `lastUpdatedDate`) or `createdDateTime`? or `creaatedAt` / `lastUpdatedAt` ?
🤔 Naming is hard!

mattwills-moj-digital:
> A few minor comments/questions, but looking good 👍
> 
> I wonder also if, given that the "domain" is the most important thing (above all else arguably), we should add some doc somewhere? I'm not thinking we go overboard, but perhaps some class level kdoc on the "root" class for a given area of the domain (`Goal` in this case), that just explains what a Goal is and what it comprises perhaps ? I'm just thinking of other devs as and when they work in our codebase 🤔

We definitely need to add kdoc - I was planning to 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/goal/Goal.kt
mattwills-moj-digital: Perhaps we should default status to `NOT_STARTED`

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/goal/Goal.kt
mattwills-moj-digital: Opted for the latter 👍

nathanrussell-moj-digital (approved):
LGTM 👍

PR #5, opened 2023-06-16
Added basic timeline domain classes and basis service

nathanrussell-moj-digital:
This PR adds the basic domain classes for Timeline, and also a basis domain service so as to start establishing our patterns in this respect

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/timeline/Link.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/timeline/Link.kt
@@ -1,6 +1,20 @@
 package uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.timeline
 
+/**
+ * A [Link] represents a link to the source event in the source system for a given [TimelineEvent].
+ *
+ * For example a given [TimelineEvent] may be pertaining to something that happened in Application-A; but it may be
+ * that Application-B is retrieving and processing the [Timeline]. The [Link] can be used to provide a link from
+ * Application-B back to the specific event in Application-B
+ *
+ * When we talk about 'links' we typically think of web anchor tag links, and it is acknowledged that will be the
+ * primary use case.
+ * The domain is meant to be the business domain and abstracted from any technology or presentation layer. The concept
+ * of a 'link' is difficult to model without thinking about web presentations, and certainly that is how the business
+ * think about it. The field `location` is deliberately named to abstract away from a web href, but it is acknowledged
+ * this will be the primary use case.
+ */
 data class Link(
   val title: String,
-  val href: String,
+  val location: String,
 )
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/timeline/Link.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/timeline/Link.kt
@@ -5,7 +5,7 @@ package uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.timeline
  *
  * For example a given [TimelineEvent] may be pertaining to something that happened in Application-A; but it may be
  * that Application-B is retrieving and processing the [Timeline]. The [Link] can be used to provide a link from
- * Application-B back to the specific event in Application-B
+ * Application-B back to the specific event in Application-A
  *
  * When we talk about 'links' we typically think of web anchor tag links, and it is acknowledged that will be the
  * primary use case.
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/timeline/Link.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/timeline/Link.kt
@@ -1,0 +1,6 @@
+package uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.timeline
+
+data class Link(
+  val title: String,
+  val href: String,
+)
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/timeline/Timeline.kt
```diff
@@ -0,0 +1,34 @@
...
+class Timeline(
+  val prisonNumber: String,
+  events: List<TimelineEvent> = emptyList(),
+) {
+
+  // _events is the real property and is deliberately private to prevent external access and mutation.
+  // events is a pseudo property that returns the current _events field as a List (ie. non-mutable)
+  private val _events: MutableList<TimelineEvent>
+  val events: List<TimelineEvent>
+    get() = _events.toList()
```
nathanrussell-moj-digital, line 18: I'm not entirely happy with this approach to prevent someone using the property directly and mutating it. Having the "real" property as private, and then a "pseudo" property seemed like an OK'ish approach (in the absence of finding anything native in the language to help); but I'm not entirely comfortable with the underscore prefix.

Happy for any comments or suggestions here

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/timeline/service/TimelinePersistenceService.kt
```diff
@@ -0,0 +1,21 @@
+package uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.timeline.service
+
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.timeline.Timeline
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.timeline.TimelineEvent
+
```
nathanrussell-moj-digital, line 5: "PersistenceService" is the terminology I introduced in OneTimeCode; though I did think of a better name for it the other day (and I've now forgotten it !!!)

src/test/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/testfixtures/domain/timeline/TimeLineBuilder.kt
```diff
@@ -0,0 +1,34 @@
+package uk.gov.justice.digital.hmpps.educationandworkplanapi.testfixtures.domain.timeline
```
nathanrussell-moj-digital, line 1: This file really needs to live in the `testFixtures` source root, rather than in the `test` source root in a package called `testfixtures` !!!

There's some weird conflict between the standard gradle plugin `java-test-fixtures`, and something in the DPS Spring Boot gradle plugin. I fell into a bit of a time sink rabbit hole last night trying to get to the bottom of it !!!, so for now we'll have it here in the `test` source root, and we can look to work out what is going on some other time

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/timeline/Timeline.kt
mattwills-moj-digital: If I've understood this correctly, you don't want to return the real List because that would allow someone to mutate it (a problem a naive developer often overlooks, because they think simply making the fields private is enough to make something immutable).

Another approach (which I think was recommended by Joshua Bloch, but it's a long time since I read his book!), is to return a copy of the list. Could you do that instead?

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/timeline/Link.kt
```diff
@@ -0,0 +1,6 @@
+package uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.timeline
+
+data class Link(
```
mattwills-moj-digital, line 3: I wasn't sure what this is initially (some kdoc would be good!), but after looking at the wireframe, I guess it represents the links to "View details" or "view assessment" etc?

I think I understand why you need the href here (since the event in question belongs in different domain and we might not know (or want to know) how to derive it?), but I think this needs some further thought/discussion/design.

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/timeline/Timeline.kt
nathanrussell-moj-digital: > If I've understood this correctly, you don't want to return the real List because that would allow someone to mutate it
Yes, exactly. And the order/sequence of the list is the important thing which we want to maintain in the domain

But thinking about it, there might be a different approach - the key thing we're interested in is the order of the list of events, not necessarily preventing mutation. The idea of preventing mutation was to ensure we were in control of the list, and therefore the order. But I wonder if there's a different way to do this which would be easier to read and grok .... let me have a play ...

nathanrussell-moj-digital: I've just committed a change to this class to simplify the events list and the part around ensuring its an ordered list

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/timeline/service/TimelinePersistenceService.kt
nathanrussell-moj-digital: I've remembered it now .... how about `PersistenceAdapter` ? Because thats actually what it does (adapts the persistence concern from the app - which could be jpa, or mongo, or flat file, etc - into the domain)
"Adapter" also aligns with some of the terminology from the Hexagonal Architecture that we are basing this on (ie. "ports and adapters")

So ... shall we use the term "persistence adapter", in which case this interface becomes `TimelinePersistenceAdapter` ?

What do you think?

mattwills-moj-digital: `PersistenceBridge` 😉 

Just kidding, LGTM.

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/timeline/Link.kt
nathanrussell-moj-digital: Added kdoc 👍

mattwills-moj-digital (approved):
LGTM 👍

PR #6, opened 2023-06-19
Enable jar task to fix test-fixtures

nathanrussell-moj-digital:
This PR fixes the test fixtures.
The problem was that `java-test-fixtures` is dependent on the `jar` task at build time (to access the `main` classes from the Spring Boot plain jar).
However, `dps-gradle-spring-boot` disables the `jar` task because it creates the Spring Boot plain jar, which in turn means there are 2 jars in the lib folder once the Spring Boot fat jar has been created. This means the `Dockerfile` has indeterminate behaviour as to which jar file to copies into the container.

This PR fixes the problem by enabling the `jar` task (to get `java-test-fixtures` working again); and then deleting the plain jar (and the test fixtures jar) before the Spring Boot fat jar is created.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -50,11 +50,11 @@ tasks.named("assemble") {
   // `assemble` task assembles the classes and dependencies into a fat jar
   // Beforehand we need to remove the plain jar and test-fixtures jars if they exist
   doFirst {
-    files(
+    delete(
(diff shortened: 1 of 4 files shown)


mattwills-moj-digital (approved):
Nice work 👍

mattwills-moj-digital (approved):

PR #11, opened 2023-06-19
Add OpenAPI specification and gradle model generation

mattwills-moj-digital:
This PR adds the Open API specification within its own file under src/main/resources/openapi. This is opposed to using Swagger annotations in the Controller classes for the following reasons:

1. It avoids the Controller classes becoming cluttered with numerous Swagger annotations, which often make it difficult to see the underlying code.
2. Being in its own file (which is nicely rendered within IntelliJ or swagger editor), it's easy to view/edit the specification.
3. Since it is not placed within annotations, the resulting specification is typically more detailed, with much richer prose.
4. It enables the API model objects to be easily generated.

The often cited drawback of this approach is that, with the exception of the model objects, the code and spec can easily diverge from each other. However, we will address this by ensuring any changes to the implementation of the API (e.g. the Controller classes) is backed up by a suitable change to the Open API spec (including a bump to its version).

---
This PR also adds the concept of an `ActionPlan` to the domain model, partly because the concept of a Prisoner's "Action Plan" could contain more than just a list of Goals in future.

---

I will create subsequent PRs to add:
1. Error responses.
2. Serving the swagger spec from the API.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -25,6 +25,7 @@ dependencies {
   implementation("org.springframework.boot:spring-boot-starter-data-jpa")
 
   implementation("org.springdoc:springdoc-openapi-ui:1.7.0")
+  implementation("org.springframework.boot:spring-boot-starter-validation")
 
   implementation("io.github.microutils:kotlin-logging:3.0.5")
 
@@ -111,14 +112,17 @@ tasks.register<JacocoReport>("combineJacocoReports") {
 }
 
 tasks.register<GenerateTask>("buildEducationAndWorkPlanModel") {
-  generatorName.set("kotlin")
+  validateSpec.set(true)
+  generatorName.set("kotlin-spring")
   inputSpec.set("$projectDir/src/main/resources/openapi/EducationAndWorkPlanAPI.yml")
   outputDir.set("$buildDir/generated")
   modelPackage.set("uk.gov.justice.digital.hmpps.educationandworkplanapi.model")
   configOptions.set(
     mapOf(
       "dateLibrary" to "java8",
-      "serializationLibrary" to "jackson"
+      "serializationLibrary" to "jackson",
+      "useBeanValidation" to "true",
+      "useSpringBoot3" to "true"
     )
   )
   globalProperties.set(
--- a/src/main/resources/openapi/EducationAndWorkPlanAPI.yml
+++ b/src/main/resources/openapi/EducationAndWorkPlanAPI.yml
@@ -51,7 +51,7 @@ paths:
         required: true
     post:
       summary: Creates a Goal.
-      description: Creates a Goal and saves it to the Prisoner's Action Plan (which will be created if it doesn't already exist).
+      description: Creates a Goal and saves it to the Prisoner's Action Plan (the latter will be created if it doesn't already exist).
       tags:
         - Action Plan - Single Goal
       responses:
@@ -139,6 +139,7 @@ components:
         - title
         - reviewDate
         - status
+        - steps
         - createdBy
         - createdAt
         - updatedBy


build.gradle.kts
```diff
@@ -1,6 +1,11 @@
+import org.jlleitschuh.gradle.ktlint.tasks.KtLintCheckTask
+import org.jlleitschuh.gradle.ktlint.tasks.KtLintFormatTask
+import org.openapitools.generator.gradle.plugin.tasks.GenerateTask
+
 plugins {
   id("uk.gov.justice.hmpps.gradle-spring-boot") version "5.2.0"
-  kotlin("plugin.spring") version "1.8.21"
+  id("org.openapi.generator") version "6.6.0"
+  kotlin("plugin.spring") version "1.8.22"
   kotlin("plugin.jpa") version "1.8.21"
```
nathanrussell-moj-digital, line 9: Very minor, but if you are bumping `plugin.spring` could you also bump `plugin.jpa`?
I have no evidence to support this, but I believe these core kotlin plugins should be keep in sync in respect of their versions (though like I said I've not read that anyway, or have any other evidence - just seems "logical" to me?)

build.gradle.kts
```diff
@@ -25,6 +32,8 @@ dependencies {
   // Integration test dependencies
   integrationTestImplementation("com.h2database:h2")
 
+  implementation("org.springdoc:springdoc-openapi-ui:1.7.0")
```
nathanrussell-moj-digital, line 35: Minor. and I accept this is a subjective style thing .... but is it worth grouping all the dependency groups together? IE. put all the main implementations together, all the runtime deps together, all the implement test deps together, all the test deps together etc?
If so, would this line be better off somewhere near line 26 ?

src/main/resources/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -0,0 +1,278 @@
...
+  # --------------------------------------------------------------------------------
+  #
+
+  #
+  # Goals
+  # --------------------------------------------------------------------------------
+  '/prisoners/{prisonNumber}/action-plan':
+    parameters:
+      - name: prisonNumber
+        description: The ID of the Prisoner (confusingly not the prison) who may have created (or is in the process of creating) an "Action Plan", consisting of one or more Goals.
```
nathanrussell-moj-digital, line 27: I'd be tempting to remove the "(confusingly not the prison)" - its only confusing to newbies to HMPPS. I suspect other devs who might read/use this spec know this 😁

src/main/resources/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -0,0 +1,278 @@
...
+        schema:
+          type: string
+        example: 'A1234BC'
+        in: path
+        required: true
+    post:
+      summary: Creates a Goal.
+      description: Creates a Goal and saves it to the Prisoner's Action Plan (which will be created if it doesn't already exist).
```
nathanrussell-moj-digital, line 54: What if it does already exist? How do we tell if it exists or not (equality)? Do we care? or do we just create it?

src/main/resources/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -0,0 +1,278 @@
...
+        goals:
+          type: array
+          description: A List of at least one or more Goals.
+          items:
+            $ref: '#/components/schemas/GoalResponse'
+      required:
+        - prisonNumber
+        - goals
+    GoalResponse:
+      title: Goal
```
nathanrussell-moj-digital, line 90: `title` is wrong?

src/main/resources/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -0,0 +1,278 @@
...
+          example: '2023-08-19'
+        status:
+          $ref: '#/components/schemas/StepStatus'
+        sequenceNumber:
+          type: integer
+          description: The number (position) of the Step within the overall Goal.
+          example: 1
+
+    CreateGoalRequest:
+      title: Goal
```
nathanrussell-moj-digital, line 171: `title` is wrong?

src/main/resources/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -0,0 +1,278 @@
...
+          description: A List of at least one Step.
+          items:
+            $ref: '#/components/schemas/CreateStepRequest'
+        notes:
+          type: string
+          description: Some additional notes related to the Goal.
+          example: Pay close attention to Peter's behaviour.
+      required:
+        - title
+        - reviewDate
```
nathanrussell-moj-digital, line 195: Should `steps` be required here, given that we know we cannot create a Goal without any steps?

src/main/resources/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -0,0 +1,278 @@
...
+        reviewDate:
+          type: string
+          format: date
+          description: An ISO-8601 date representing when the Goal is up for review. Defaults to 6 months.
+          example: '2023-12-19'
+        steps:
+          type: array
+          description: A List of at least one Step.
+          items:
+            $ref: '#/components/schemas/CreateStepRequest'
```
nathanrussell-moj-digital, line 188: Is there some kind of openapi notation we can use here to say the list must  be populated with at least 1 step ?

src/main/resources/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -0,0 +1,278 @@
...
+          example: Book first aid course
+        targetDate:
+          type: string
+          format: date
+          description: A date by which the prisoner should aim to complete the step.
+          example: '2023-12-19'
+        sequenceNumber:
+          type: integer
+          description: The number (position) of the Step within the overall Goal.
+          example: 1
```
nathanrussell-moj-digital, line 213: Any required properties for `CreateStepRequest` ?
Also, sequence number .... I wonder if we just simply add it to the bottom of the list in the domain, and let the domain class(es) sequence it and assign it the correct sequence number?

src/main/resources/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -0,0 +1,278 @@
...
+      enum:
+        - ACTIVE
+        - COMPLETE
+        - ARCHIVED
+    StepStatus:
+      title: StepStatus
+      description: Enum describing the various statuses of a Step's lifecyle.
+      type: string
+      enum:
+        - NOT_STARTED
```
nathanrussell-moj-digital, line 228: 👍 for using uppercase enums in the REST API 👍 
(Unlike our previous project where they standardised on lowercase kebab case, which resulted in generated model enums having the word `MINUS` in them !!! - eg: `not-started` would be generated as `NOT_MINUS_STARTED` !!!)

nathanrussell-moj-digital (commented):
Sorry for so many comments! Many of which are a bit nit-picky! Sorry!

My thoughts though are that, whilst I agree with you as to the motivations and reasons for writing a spec vs. generating a spec from annotations, if we are going to do it we need to do it really really well! We don't to give other devs/teams the opportunity to say we shouldn't have done it.

Also, something we are going to need to work out .... not only does the current HMPPS standard approach generate the spec, they also serve it as a static resource via the spring doc plugins etc - eg: https://activities-api-dev.prison.service.justice.gov.uk/swagger-ui/index.html?configUrl=/v3/api-docs/swagger-config

We are going to need to workout how to configure the relevant spring doc class/config to point at our static resource in the resources folder. Wonder if its worth looking at this in this PR? (because if we cant do it we are not likely to win this one .... 🤔 )

build.gradle.kts
mattwills-moj-digital: Oops, I should have done that. Missed it in the merge conflict 👍

src/main/resources/openapi/EducationAndWorkPlanAPI.yml
mattwills-moj-digital: Funnily enough, I had the same thought and nearly removed it :)

src/main/resources/openapi/EducationAndWorkPlanAPI.yml
mattwills-moj-digital: I assume we just create it. We could have dozens of goals that are effectively duplicates, but I don't think it would ever be used that way.

src/main/resources/openapi/EducationAndWorkPlanAPI.yml
mattwills-moj-digital: Oops, I clearly should have reviewed this myself before creating the PR! It was late in the day but that's no excuse 🙄

src/main/resources/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -0,0 +1,278 @@
...
+      description: A request to create an individual Goal that a Prisoner aims to complete as part of their Action Plan. The goal's status will be set to 'ACTIVE'.
+      properties:
+        title:
+          type: string
+          description: A title explaining the aim of the goal.
+          example: Improve communication skills
+        reviewDate:
+          type: string
+          format: date
+          description: An ISO-8601 date representing when the Goal is up for review. Defaults to 6 months.
```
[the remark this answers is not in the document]
mattwills-moj-digital: I reckon so, unless we default it?

src/main/resources/openapi/EducationAndWorkPlanAPI.yml
mattwills-moj-digital: I've tidied up several things in the spec, including this 👍

src/main/resources/openapi/EducationAndWorkPlanAPI.yml
nathanrussell-moj-digital: Agreed. Which was why I was wondering why we even mention it - ie. whats the relevance/significance of "which will be created if it doesn't already exist" ?
We are going to simply create it?  no?

src/main/resources/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -0,0 +1,291 @@
...
+          type: string
+          format: date-time
+          description: An ISO-8601 timestamp representing when the Goal was last updated. This will be the same as the created date if it has not yet been updated.
+          example: '2023-06-19T09:39:44Z'
+      required:
+        - goalReference
+        - title
+        - reviewDate
+        - status
+        - createdBy
```
nathanrussell-moj-digital, line 142: I know you've now added the minItems value to the steps array 👍 , but I think we also need steps in the required array as well?
IE. its a mandatory property AND it needs to have at least 1 item in the array
(rather than its an optional property, but if its provided it needs to have at least 1 item in the array)

src/main/resources/openapi/EducationAndWorkPlanAPI.yml
mattwills-moj-digital: I was referring to the ActionPlan. I'll try to make it clearer.

nathanrussell-moj-digital (approved):
Thanks for addressing the comments
LGTM 👍

src/main/resources/openapi/EducationAndWorkPlanAPI.yml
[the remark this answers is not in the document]
mattwills-moj-digital: I removed that wording, but I think some kind of requirement like that could be coming our way.

PR #13, opened 2023-06-20
Serve OpenAPI spec 

mattwills-moj-digital:
This PR serves the Open API specification file, which lives under `src/main/resources/static/openapi/`, via the standard HMPPS swagger endpoint: `/swagger-ui/index.html?configUrl=/v3/api-docs/swagger-config`.

It also updates the spec itself to include an `ErrorResponse` object where appropriate.


src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -57,6 +59,10 @@ paths:
       responses:
         '201':
           $ref: '#/components/responses/Goal'
+        '400':
+          $ref: '#/components/responses/400ErrorResponse'
+        '404':
```
nathanrussell-moj-digital, line 64: Minor, and perhaps I'm missing something, but why would POSTing to create a goal result in a 404?
Especially when we say in the description "the latter will be created if it doesn't already exist" ?

nathanrussell-moj-digital (approved):
Love it; ATOB 👍 😁

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
mattwills-moj-digital: In case the `{prisonNumber}` in the path isn't known 👍

nathanrussell-moj-digital: Hmmmmm ..... is that a thing? .... how might we know if the prison number is valid or not? .... 🤔  ... is that the responsibility of our service? .... 🤔

mattwills-moj-digital: I haven't fully thought this through, as I'm not familiar with how things work here, but I was thinking our service might want to check the prisonNumber exists via NOMIS, or some other system.

Or to put it another way, what happened in our previous project when an attempt was made, for example, to create a PaperApplication for an ERO that doesn't exist?

PR #19, opened 2023-06-21
Fix startup problems when deployed

nathanrussell-moj-digital:
The PR fixes the startup problems when the app is deployed

The problem was related to spring security in that it could not find a JwtDecoder bean. This appears to be because I had not configured where `hmpps-auth` was (ie. it is `hmpps-auth` that will issue the JWT and therefore issues the jwks to decode it)

To (hopefully) prove that it works I've put together the local `docker-compose` file complete the a db and relevant config. Therefore we can start the app locally with `docker-compose up` (we will build upon and improve this in the future, but it allows us to start the app locally in the immediate term)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM --platform=$BUILDPLATFORM eclipse-temurin:19-jre-jammy AS builder
+FROM --platform=$BUILDPLATFORM eclipse-temurin:19-jdk-jammy AS builder
```
nathanrussell-moj-digital, line 1: I had to change the `Dockerfile` to use a JDK as the base image (rather than a JRE)
This is because we are using `mapstruct`, which in turn uses `kapt`, which in turn requires a JDK at build time!
Hopefully we wont get frowned upon for doing this .... 😢

src/main/resources/application.yml
```diff
@@ -8,6 +8,12 @@ spring:
   codec:
     max-in-memory-size: 10MB
 
+  security:
```
nathanrussell-moj-digital, line 11: This appears to be the magic sauce that was missing. The env var value will be pulled in at runtime from the various helm values files (above)

src/main/resources/banner.txt
```diff
@@ -1,13 +1,15 @@
...
-___ ____ _  _ ___  _    ____ ___ ____ 
- |  |___ |\/| |__] |    |__|  |  |___ 
- |  |___ |  | |    |___ |  |  |  |___ 
-                                      
-_  _ ____ ___ _    _ _  _ 
-|_/  |  |  |  |    | |\ | 
-| \_ |__|  |  |___ | | \| 
-                          
-Please change me by generating your own ASCII art and placing in banner.txt
+
```
nathanrussell-moj-digital, line 1: Its about time we did this! 😁

README.md
```diff
@@ -2,43 +2,42 @@
...
 [![CircleCI](https://circleci.com/gh/ministryofjustice/hmpps-education-and-work-plan-api/tree/main.svg?style=svg)](https://circleci.com/gh/ministryofjustice/hmpps-education-and-work-plan-api)
 [![Docker Repository on Quay](https://quay.io/repository/hmpps/hmpps-education-and-work-plan-api/status "Docker Repository on Quay")](https://quay.io/repository/hmpps/hmpps-education-and-work-plan-api)
-[![API docs](https://img.shields.io/badge/API_docs_-view-85EA2D.svg?logo=swagger)](https://hmpps-education-and-work-plan-api-dev.hmpps.service.justice.gov.uk/webjars/swagger-ui/index.html?configUrl=/v3/api-docs)
+[![API docs](https://img.shields.io/badge/API_docs_-view-85EA2D.svg?logo=swagger)](https://hmpps-education-and-work-plan-api-dev.hmpps.service.justice.gov.uk/swagger-ui/index.html?configUrl=/v3/api-docs)
```
mattwills-moj-digital, line 5: 👍

docker-compose.yml
```diff
@@ -12,6 +13,24 @@ services:
     environment:
       - SERVER_PORT=8080
       - SPRING_PROFILES_ACTIVE=dev
+      - HMPPS_AUTH_URL=https://sign-in-dev.hmpps.service.justice.gov.uk/auth
```
mattwills-moj-digital, line 16: Minor - I wonder if there's a docker alternative, so we're not relying on that service.

mattwills-moj-digital (approved):

docker-compose.yml
nathanrussell-moj-digital: Yes, there totally is. We can bring hmpps-auth in as a docker container and run it locally 👍 
Its just that I wanted to get something up quickly in this case - we can defo add to this over time 👍

PR #21, opened 2023-06-22
ActionPlanRepository and JPA PersistenceAdapter to retrieve an ActionPlan

mattwills-moj-digital:
This PR should enable us to retrieve an `ActionPlan` all the way via the REST controller (subject to integration tests which I'll add in a subsequent PR).

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/JpaActionPlanPersistenceAdapter.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/JpaActionPlanPersistenceAdapter.kt
@@ -1,0 +1,20 @@
+package uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa
+
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.mapper.ActionPlanEntityMapper
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.repository.ActionPlanRepository
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.goal.ActionPlan
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.goal.ActionPlanNotFoundException
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.goal.service.ActionPlanPersistenceAdapter
+
+@Component
+class JpaActionPlanPersistenceAdapter(
+  private val actionPlanRepository: ActionPlanRepository,
+  private val actionPlanMapper: ActionPlanEntityMapper,
+) : ActionPlanPersistenceAdapter {
+  override fun getActionPlan(prisonNumber: String): ActionPlan {
+    val entity = actionPlanRepository.findByPrisonNumber(prisonNumber)
+      ?: throw ActionPlanNotFoundException("No Action Plan found for prisoner [$prisonNumber]")
+    return actionPlanMapper.fromEntityToDomain(entity)
+  }
+}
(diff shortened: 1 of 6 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/JpaActionPlanPersistenceAdapterTest.kt
```diff
@@ -0,0 +1,47 @@
...
+    given(actionPlanMapper.fromEntityToDomain(any())).willReturn(expectedActionPlanDomain)
+
+    // When
+    val actual = persistenceAdapter.getActionPlan(prisonNumber)
+
+    // Then
+    assertThat(actual).isEqualTo(expectedActionPlanDomain)
+    verify(actionPlanRepository).findByPrisonNumber(prisonNumber)
+    verify(actionPlanMapper).fromEntityToDomain(actionPlanEntity)
+  }
```
nathanrussell-moj-digital, line 46: Do we need a test for when the action plan cannot be found, and the adapter throws the exception?

(Though I'm not convinced about the persistence adapter throwing exceptions .... we might need to think that one through)

mattwills-moj-digital: Good point. I was thinking we want an exception to be thrown as early as possible, but I guess there could be a use case where `null` is acceptable (for example when checking if the entity exists). I'll move it to the domain service and throw the exception there instead.

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/JpaActionPlanPersistenceAdapter.kt
```diff
@@ -0,0 +1,20 @@
...
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.goal.service.ActionPlanPersistenceAdapter
+
+@Component
+class JpaActionPlanPersistenceAdapter(
+  private val actionPlanRepository: ActionPlanRepository,
+  private val actionPlanMapper: ActionPlanEntityMapper,
+) : ActionPlanPersistenceAdapter {
+  override fun getActionPlan(prisonNumber: String): ActionPlan {
+    val entity = actionPlanRepository.findByPrisonNumber(prisonNumber)
+      ?: throw ActionPlanNotFoundException("No Action Plan found for prisoner [$prisonNumber]")
```
nathanrussell-moj-digital, line 17: I'm not sure about the persistence adapter throwing exceptions. This is what we have in the interface (I know it was my doing originally ... 🤔 )
```
/**
 * Persistence Adapter for [ActionPlan] instances.
 *
 * Implementations should use the underlying persistence of the application in question, eg: JPA, Mongo, Dynamo, Redis etc
 *
 * Implementations should not throw exceptions. These are not part of the interface and are not checked or handled by
 * [ActionPlanService].
 */
interface ActionPlanPersistenceAdapter {

  fun getActionPlan(prisonNumber: String): ActionPlan
}
```

I think the problem I was trying to prevent is that the persistence adapter impl (which is in the `app` codebase) is called by the service (which is in the `domain`) codebase.

Allowing the persistence adapter (app) to throw exceptions means that the service (domain) needs to handle them.

Perhaps this is OK in this case, where the exception is a `domain` exception ..... but I think I was trying to prevent persistence adapters throwing anything - especially exceptions pertaining to the underlying persistence technology ....

src/test/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/JpaActionPlanPersistenceAdapterTest.kt
mattwills-moj-digital: Moved to `ActionPlanService` (and corresponding test added).

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/JpaActionPlanPersistenceAdapter.kt
mattwills-moj-digital: I agree. See my other comment.

nathanrussell-moj-digital (approved):
LGTM 👍

PR #41, opened 2023-07-27
RR-161 - Return the goals in chronological order

mattwills-moj-digital:
As per the ACs in RR-106, the goals should be ordered chronologically (oldest first) and it makes sense to do this in the API.


nathanrussell-moj-digital (approved):
LGTM 👍

PR #43, opened 2023-07-27
RR-3 - Edit goal swagger spec

mattwills-moj-digital:
Swagger definition for a new endpoint to edit a Goal.

This endpoint is an "all or nothing" type affair - the whole Goal needs to be provided, or any missing data (e.g. a single Step) will be removed.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/mapper/StepResourceMapper.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/mapper/StepResourceMapper.kt
@@ -3,7 +3,7 @@ package uk.gov.justice.digital.hmpps.educationandworkplanapi.app.resource.mapper
 import org.mapstruct.Mapper
 import org.mapstruct.Mapping
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.goal.Step
-import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.CreateStepRequest
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.StepRequest
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.StepResponse
 import java.util.UUID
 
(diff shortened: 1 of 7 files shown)


src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -259,6 +291,65 @@ components:
...
+      required:
+        - goalReference
+        - title
+        - steps
+    UpdateStepRequest:
+      title: UpdateStepRequest
+      type: object
+      description: A request to create a Step within a Goal.
+      properties:
+        stepReference:
```
nathanrussell-moj-digital, line 332: LGTM 👍 
My only thought/question would be whether `stepReference` should be a mandatory field or not ... 🤔 
I'm thinking about a future story as part of "editing a goal", where the screens allow for adding steps as part of "editing a goal". In this case we wont have a `stepReference` (its a new step); so I'm wondering if the API needs to support it in this way and therefore the implementation will need to work out what is new/what is edited/what is deleted as part of working out the diff

What do you think?

mattwills-moj-digital: That's a good point, I'll change it 👍

src/testFixtures/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/resource/model/StepRequestBuilder.kt
```diff
@@ -18,8 +19,21 @@ fun anotherValidCreateStepRequest(
...
-): CreateStepRequest =
-  CreateStepRequest(
+): StepRequest =
+  StepRequest(
+    title = title,
+    targetDateRange = targetDateRange,
+    sequenceNumber = sequenceNumber,
+  )
+
+fun aValidUpdateStepRequest(
```
mattwills-moj-digital, line 29: I've added this to show how we could use this builder for creating and updating a Step.

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/mapper/StepResourceMapper.kt
```diff
@@ -13,9 +13,10 @@ import java.util.UUID
   ],
 )
 interface StepResourceMapper {
+  // TODO - RR-3 - this will need to change when creating vs updating a Step
```
nathanrussell-moj-digital, line 16: 👍

nathanrussell-moj-digital (approved):
LGTM 👍

PR #48, opened 2023-08-01
RR-3 - Updated goal steps to ensure and enforce sequence number ordering

nathanrussell-moj-digital:
This PR updates the domain `Goal` class to enforce that a `Goal`s `Step`s are correctly ordered based on their `sequenceNumber` ; and that when adding a new `Step` it is added in the correct position in the ordered list of `Step`s based on it's `sequenceNumber`

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/goal/Goal.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/goal/Goal.kt
@@ -32,9 +32,16 @@ class Goal(
 
   /**
    * Returns the Goal's [Step]s, ordered by their sequence number (ascending).
+   * The sequence number is enforced to be sequential starting from 1 with no gaps.
    */
   val steps: MutableList<Step>
-    get() = field.also { steps -> steps.sortBy { it.sequenceNumber } }
+    get() = field.also { steps ->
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/goal/Goal.kt
```diff
@@ -32,9 +32,16 @@ class Goal(
 
   /**
    * Returns the Goal's [Step]s, ordered by their sequence number (ascending).
+   * The sequence number is enforced to be sequential starting from 1 with no gaps.
    */
   val steps: MutableList<Step>
-    get() = field.also { steps -> steps.sortBy { it.sequenceNumber } }
+    get() = field.also { steps ->
```
nathanrussell-moj-digital, line 38: Previously the `steps` getter used to return the steps ordered by their sequence number (each `Step` has a `sequenceNumber` field)
This did not account for gaps within the sequence numbers (ie. 1, 2, 4); and did not account for duplicates (ie. you could have 2 steps with sequence number 1, in which case the order was indeterminate)

The change here orders them based on their current sequence number, then updates the sequence number of each `Step` starting sequentially from 1

Now the `Step`s will always be ordered with a sequential sequence

chrimesdev:
LGTM

PR #54, opened 2023-08-15
RR-229 - Add new fields to Action Plan

mattwills-moj-digital:
This PR add `reference` and `reviewDate` to the `ActionPlan` domain, entity and DB table.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/entity/ActionPlanEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/entity/ActionPlanEntity.kt
@@ -71,8 +71,8 @@ class ActionPlanEntity(
     /**
      * Returns new un-persisted [ActionPlanEntity] for the specified prisoner with an empty collection of [GoalEntity]s
      */
-    fun newActionPlanForPrisoner(prisonNumber: String, reviewDate: LocalDate?): ActionPlanEntity =
-      ActionPlanEntity(reference = UUID.randomUUID(), prisonNumber = prisonNumber, reviewDate = reviewDate, goals = mutableListOf())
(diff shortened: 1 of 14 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/entity/ActionPlanEntity.kt
```diff
@@ -63,8 +71,8 @@ class ActionPlanEntity(
     /**
      * Returns new un-persisted [ActionPlanEntity] for the specified prisoner with an empty collection of [GoalEntity]s
      */
-    fun newActionPlanForPrisoner(prisonNumber: String): ActionPlanEntity =
-      ActionPlanEntity(prisonNumber = prisonNumber, goals = mutableListOf())
+    fun newActionPlanForPrisoner(prisonNumber: String, reviewDate: LocalDate?): ActionPlanEntity =
+      ActionPlanEntity(reference = UUID.randomUUID(), prisonNumber = prisonNumber, reviewDate = reviewDate, goals = mutableListOf())
```
nathanrussell-moj-digital, line 75: Slightly uncomfortable with this in that we have a persistence concern effectively deciding how to create a domain object .... 🤔 

IE. the reference and review date of an ActionPlan are actually domain concerns I think (though I know our mappers that map from the REST API to Domain objects for Goals and Steps do this currently .... 🤔 )

The purist in me is nagging at me! 🤣 - if we changed our DB layer to mongo or dynamo or something non-JPA (not that we would) , we would need to move/re-implement that logic .... feels wrong ... 🤔 

Lets go with it for now, but it is eating me alive! 🤣

nathanrussell-moj-digital (approved):
LGTM, though I have left a comment/concern about who is responsible for generating the reference .... 🤔

nathanrussell-moj-digital (approved):

PR #55, opened 2023-08-16
RR-228 - Endpoint to create an Action Plan with a reviewDate

mattwills-moj-digital:
This PR adds the swagger definition and endpoint to an **Action Plan**, with a list of one or more **Goals**.

As part of this change, I've added the new `reference` and `reviewDate` fields to the GET Action Plan response.

It is the end result of a series of previously reviewed and merged PRs into this branch.

### DIFF
--- a/src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
+++ b/src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
@@ -37,7 +37,7 @@ paths:
         - Action Plan
       responses:
         '201':
-          $ref: '#/components/responses/ActionPlan'
+          description: The Action Plan was created successfully.
         '400':
           $ref: '#/components/responses/400ErrorResponse'
         '404':
@@ -71,7 +71,7 @@ paths:
         - Action Plan - Single Goal
       responses:
         '201':
-          $ref: '#/components/responses/Goal'
+          description: The Goal was created successfully.
         '400':
           $ref: '#/components/responses/400ErrorResponse'
         '404':
@@ -282,7 +282,8 @@ components:
           example: '2023-12-19'
         goals:
           type: array
-          description: A List of at least one or more Goals.
+          minItems: 1
+          description: A List of at least one Goal.
           items:
             $ref: '#/components/schemas/CreateGoalRequest'
       required:
--- a/src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
+++ b/src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   title: Education and Work Plan API
-  version: '1.1.3'
+  version: '1.1.4'
   description: Education and Work Plan API
   contact:
     name: Matt Wills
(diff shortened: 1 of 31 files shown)


src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -30,6 +30,21 @@ paths:
         example: 'A1234BC'
         in: path
         required: true
+    post:
+      summary: Creates an Action Plan.
```
nathanrussell-moj-digital, line 34: I know we have it everywhere .... but I'm falling out of love with the term "action" plan ... 🤔 
It's not a term the business use. Initially they were quite insistent that there was no "plan" and were keen it was not part of the service name and/or content. I think we felt a "plan" was a convenient collective noun for a number of Goals, and for some reason called it an "action plan"

Now the business have acknowledged there is a plan, I wonder if we need to try to understand a little more about what they call it.

(I'm not suggesting we do it in this PR - or even at all perhaps? - I'm just raising it as a discussion point)

Also (similar), I'm not keen on the root path any more! 😢 
`/action-plans/{prisonNumber}` is plural and suggests a prisoner can have many action plans. We know that is not true, and they can only have 1 or zero (which would be a 404) 😬

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -244,6 +270,29 @@ components:
         - status
         - sequenceNumber
 
+    CreateActionPlanRequest:
+      title: CreateActionPlanRequest
+      type: object
+      description: A request to create an  Action Plan with at least one or more Goals that a Prisoner aims to complete.
+      properties:
+        prisonNumber:
```
nathanrussell-moj-digital, line 278: Not that I mind one way or the other, but do we need prison number in the request body? Its in the URI path .... 🤔

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -244,6 +270,29 @@ components:
...
+          type: string
+          description: The ID of the prisoner
+          example: 'A1234BC'
+        reviewDate:
```
nathanrussell-moj-digital, line 282: I think we need to think this one through and understand it a little better ....

The proposed screens allow for entering a time range (0-3, 3,-6 months etc), or None, or a specific date.
Presumably we are thinking the UI typescript code (UI server side) will take the value from the user form object - eg: `3-6_MONTHS`, and work out an actual date, which will be used in this request object .

That works .... but .... will the CIAG ever need to see the original value again? Does it appear on a screen anywhere? Might it appear on an edit type screen at some point, where they CIAG can change the plan review date?
In this case we would need to know the original value (presumably?) so we can present a web screen that says "0-6 months".
I dont think we'll be able to reverse engineer / derive the original time frame, just from a date .... 🤔

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
mattwills-moj-digital: I've removed it from the request body 👍

nathanrussell-moj-digital (approved):

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
mattwills-moj-digital: Me too - I don't love "action" plan either, but I think it's near impossible to come up with a name that everyone's happy with, which the business will also stick to! 😄 

As discussed, the root path should definitely be plural, as we could in theory have an endpoint to return all Action Plans. In fact, after the meeting with the CIAG team earlier, it sounds like we'll need one that returns all the Action Plans for a Prison - e.g. perhaps a using a query param of `?prisonId` or something similar. 👍

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
mattwills-moj-digital: As discussed, I'm making a separate change to add an enum with 5 different values - e.g. `THREE_MONTHS`, `NO_REVIEW_DATE`, `SPECIFIC_DATE` - to cover the various options.

mattwills-moj-digital: I ended up getting rid of that enum! See the description of this PR: https://github.com/ministryofjustice/hmpps-education-and-work-plan-api/pull/68

nathanrussell-moj-digital (approved):
On the assumption that I've already reviewed all this in the various "feeder" PRs .... 😁 

AToAB 👍

PR #57, opened 2023-08-17
RR-237 - Configure appinsights; send event when goals are created

nathanrussell-moj-digital:
This PR configures azure Application Insights; and sends a tracking event for when a Goal has been created


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/config/TelemetryService.kt
```diff
@@ -0,0 +1,49 @@
...
+  /*
+  The TelemetryClient is instantiated and injected at runtime by ApplicationInsights java agent via AppInsightsConfigManager
+  in the dps-gradle-spring-boot-plugin (https://github.com/ministryofjustice/dps-gradle-spring-boot/blob/spring-boot-3/src/main/kotlin/uk/gov/justice/digital/hmpps/gradle/configmanagers/AppInsightsConfigManager.kt#L35)
+  where the jar is available at runtime courtesy of the Dockerfile. This means that in reality there will only be a TelemetryClient
+  bean in a deployed environment (ie. a deployed docker container).
+  In non-docker runtimes (eg: running in the IDE, or running the integration tests), this bean will not exist; hence the
+  nullable field and IDE warning.
+   */
+  private val telemetryClient: TelemetryClient?,
+) {
```
nathanrussell-moj-digital, line 24: Other services create a TelemetryClient bean to address this problem (eg: https://github.com/ministryofjustice/prison-api/blob/57bd1a2094efc613c0f4558f4c443ef9f2cd1338/src/main/java/uk/gov/justice/hmpps/prison/web/config/ApplicationInsightsConfiguration.kt) but 
* I don't think their documentation explains it well enough
* We don't need a telemetry client when running locally or integration tests (and we probably dont want one - dont want to be sending loads of events as part of our integration tests etc!)
* We don't like the idea of having production code that is not used in production

Even though my kdoc is a little wordy, I hope it explains it better than providing a dummy bean like the other services do.

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/GoalController.kt
```diff
@@ -39,7 +41,9 @@ class GoalController(
     goalService.createGoal(
       prisonNumber = prisonNumber,
       goal = goalResourceMapper.fromModelToDomain(request),
-    )
+    ).apply {
+      telemetryService.trackGoalCreateEvent(this)
+    }
```
nathanrussell-moj-digital, line 46: This approach of injecting the `TelemetryService` into the controller will do for now, to prove it works and serve as an example, but I think we can do better moving forwards (this approach will get messy moving forwards)

I did consider injecting the `TelemetryService` into the `GoalService`, but the `GoalService` is a domain class/concern, and I really don't think we want the domain being dependent on things in the app (like this `TelemetryService` bean)
I did consider creating the `TelemetryService` in the domain, but that felt wrong, as sending telemetry events etc does not feel like a concern of the domain.
Hence I added it to the controller here.

I think moving forwards we can perhaps look at an AOP approach, where the `TelemetryService` is not injected into either the controller or the domain. That feels like a better separation perhaps, but we can look at that later

src/testFixtures/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/JwtAccessTokenBuilder.kt
```diff
@@ -59,6 +60,7 @@ fun buildAccessToken(
         "auth_source" to "nomis",
         "user_uuid" to UUID.randomUUID(),
         "name" to displayName,
+        "client_id" to clientId,
       ),
```
nathanrussell-moj-digital, line 64: Our JWT tokens that our integration tests use need a client_id claim because the interceptor above reads that claim and adds it to the custom dimensions
(the real JWT we get from hmpps-auth will contain a client_id claim)

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/config/ClientTrackingConfiguration.kt
```diff
@@ -0,0 +1,57 @@
...
+class ClientTrackingConfiguration(private val clientTrackingInterceptor: ClientTrackingInterceptor) : WebMvcConfigurer {
+  override fun addInterceptors(registry: InterceptorRegistry) {
+    registry.addInterceptor(clientTrackingInterceptor).addPathPatterns("/**")
+  }
+}
+
+/**
+ * An MVC [HandlerInterceptor] to add the request's authentication username and clientId to the ApplicationInsights
+ * customDimensions.
+ */
```
mattwills-moj-digital, line 28: Nice 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/GoalController.kt
mattwills-moj-digital: I think AOP would be ideal for this 👍

mattwills-moj-digital (approved):

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/config/TelemetryServiceTest.kt
```diff
@@ -0,0 +1,49 @@
...
+import org.mockito.Mock
+import org.mockito.junit.jupiter.MockitoExtension
+import org.mockito.kotlin.verify
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.goal.GoalStatus
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.goal.aValidGoal
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.goal.aValidStep
+import java.util.UUID
+
+@ExtendWith(MockitoExtension::class)
+class TelemetryServiceTest {
```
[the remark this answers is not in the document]
nathanrussell-moj-digital: Good spot; yes, its a unit test so was meant to go in `test` 👍 
Done 👍

mattwills-moj-digital (approved):
GS!

PR #59, opened 2023-08-17
RR-237 - Send event using Span

nathanrussell-moj-digital:


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/TelemetryService.kt
```diff
@@ -1,7 +1,9 @@
-package uk.gov.justice.digital.hmpps.educationandworkplanapi.app.config
+package uk.gov.justice.digital.hmpps.educationandworkplanapi.app.service
```
nathanrussell-moj-digital, line 1: I moved the `TelemetryService` from the `config` package to the `service` package (because thats where it belongs 😁 )

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/TelemetryService.kt
```diff
@@ -18,6 +20,10 @@ class TelemetryService(
 
   fun trackGoalCreateEvent(goal: Goal) {
     telemetryClient.trackEvent(GOAL_CREATE_EVENT, goal.goalCreateEventCustomDimensions())
+    with(Span.current()) {
```
nathanrussell-moj-digital, line 23: This is the new code (thats not covered by unit tests at the moment) - its a hunch that we might be able to send events using the underlying `opentelemetry` API (the `Span` class) , in which case it might use all the custom dimensions that were already set on the `Span` via the interceptor.
At the moment I am sending both events - one via the `TelemetryClient` class which we know works (albeit without the custom dimensions), and this way as well. I've given this approach (using the `Span`) a subtly different event name (for now) so we can identify it in the app insights query tool

mattwills-moj-digital (approved):

PR #65, opened 2023-08-18
RR-230 - Improvements to ITs for create/retrieve ActionPlan

mattwills-moj-digital:
This PR adds some additional checks and changes to the create/retrieve Action Plan integration tests, which I realised were missing.

### DIFF
--- a/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/GetActionPlanTest.kt
+++ b/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/GetActionPlanTest.kt
@@ -10,8 +10,11 @@ import uk.gov.justice.digital.hmpps.educationandworkplanapi.aValidTokenWithViewA
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.IntegrationTestBase
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.bearerToken
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.ActionPlanResponse
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.CreateActionPlanRequest
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.CreateGoalRequest
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.ErrorResponse
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.ReviewDateCategory
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.aValidCreateActionPlanRequest
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.aValidCreateGoalRequest
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.assertThat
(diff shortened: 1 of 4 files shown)


src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/GetActionPlanTest.kt
```diff
@@ -122,6 +135,8 @@ class GetActionPlanTest : IntegrationTestBase() {
     val actual = response.responseBody.blockFirst()
     assertThat(actual)
       .isForPrisonNumber(prisonNumber)
+      .hasReviewDateCategory(ReviewDateCategory.SIX_MONTHS)
+      .hasNoReviewDate()
```
mattwills-moj-digital, line 139: The controversial bit! I'm not sure if I should automatically populate `reviewDate` if the `reviewDateCategory` is set to 3, 6 or 12 months 🤔 I've chosen not to as there are no requirements around this, but I'll try and find out more from the BAs.

mattwills-moj-digital: mmm...the more I think about it, the more I think I should set the `reviewDate`, as I can't see any downsides to this approach. We'll still know which option the user selected (i.e. to populate the UI, or for reporting purposes).

In contrast, I can foresee problems with not setting it! 😂

nathanrussell-moj-digital: Hmmm, I'm not so sure
My thinking is that you currently have it correct, and your test is perfect as its our safety net for and and when we might get asked to change the behaviour.

If the business tell us to calculate the `reviewDate` from the value of the enum, then thats fine. But I dont think we should 2nd guess that as a requirement. I think we should only do it if its documented / communicated as a business requirement and expected behaviour.

src/testFixtures/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/resource/model/ActionPlanResponseAssert.kt
```diff
@@ -21,6 +22,36 @@ class ActionPlanResponseAssert(actual: ActionPlanResponse?) :
...
+
+  fun hasNoReviewDate(): ActionPlanResponseAssert {
+    isNotNull
+    with(actual!!) {
+      if (reviewDate != null) {
+        failWithMessage("Expected reviewDate to be null, but was $reviewDate")
+      }
+    }
+    return this
+  }
```
nathanrussell-moj-digital, line 53: ❤️ 
I absolutely love our assertJ assertion classes 👍 ❤️

nathanrussell-moj-digital (approved):
LGTM (as you currently have it, rather than deriving the `reviewDate`)

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/GetActionPlanTest.kt
mattwills-moj-digital: Let me talk you through this alternative approach: https://github.com/ministryofjustice/hmpps-education-and-work-plan-api/pull/66

PR #66, opened 2023-08-18
RR-230 populate review date

mattwills-moj-digital:
This PR populates an Action Plan's `reviewDate` based on the provided `reviewDateCategory`. This feels like a better approach than having no `reviewDate` at all - e.g. when the user selects "3 months" on the UI.

After further discussion with the business, the radio buttons are a quick way for the user to say in 3/6/12 months time, rather than having to manually enter a date each time.

image

This PR also adds some additional checks and changes to the create/retrieve Action Plan integration tests, which I realised were missing.


mattwills-moj-digital:
I've decided to remove `reviewDateCategory` as I think we're in danger of making this more complicated than it needs to be! As things stand, I reckon it will be unlikely we'll need to redisplay the previously chosen option (e.g. "3 months") when editing an Action Plan (and if we do then we can always add it back into the API, but let's not try and second guess that now).

My understanding is that the radio buttons to choose 3/6/12 months are just a convenient to set the `reviewDate` quickly, so the user doesn't have to manually type it in each time. In either case, we're still setting the `reviewDate` and I don't think we will care what the user previously choose when come to edit it, perhaps months later. We might choose to default one of the radio buttons, but that's different to rendering the previously chosen value. This is fundamentally different to a Step's `targetDateRange`, which needs to be redisplayed each time.

Refer to this PR instead: https://github.com/ministryofjustice/hmpps-education-and-work-plan-api/pull/68

PR #73, opened 2023-08-31
POC to show how we might capture the CIAG's current prison (activeCaseLoadId)

nathanrussell-moj-digital:
This PR is draft and is not intended to be merged (yet / in it's current form), and is to serve as a discussion document.

This PR is a POC to show how we might capture the CIAG's current prison (activeCaseLoadId) and store it against the Goal when we save it to the database.

It's not 100% complete yet, in that I've not done the database DDL changes yet to add the new columns, so therefore the CIAGs location isn't actually persisted yet. And there are a few tests missing!

### How it works
The way it works is that is hooks into Spring Security (which already do anyway). Currently we have a custom `Principal` object called `DpsPrincipal` that contains their username (eg: `NRUSSELL_GEN`) and their display name (eg: `Nathan Russell`). The `DpsPrinicpal` is instantiated by `AuthAwareTokenConverter`, where both fields are taken from the JWT (because they are claims in the JWT 👍 )
We then subsequently use these fields at the JPA layer via the annotations `@CreatedBy` and `@LastModifiedBy` (standard annotations) and our custom annotations `@CreatedByDisplayName` and `@LastModifiedByDisplayName`. These last 2 annotations are processed by `DisplayNameAuditingEntityListener` which sets the field values from the fields in the `DpsPrincipal`

What I'm proposing here is that our `DpsPrincipal` contains a new field - `activeCaseLoadId` and that is populated by `AuthAwareTokenConverter`. The active caseload ID is unfortunately not a claim in the JWT, so `AuthAwareTokenConverter` needs to use a new client class `HmppsAuthClient` to call HMPPS Auth's `/api/user/me` endpoint. This returns a small object that contains the field `activeCaseLoadId`

Now that `DpsPrincipal` contains the active case load ID, it can be used at the JPA layer via some new annotations and a new entity listener class.

Thats basically it 👍 

### Things missing / considerations
* There are a handful of unit tests missing - its a POC after all 😁 (though integration test support is provided for via the introduction of wiremock)
* The DDL to add the new columns to the `Goal` table is missing; I've not done that yet. But I have got as far as adding the new fields to the entity class with the new annotations; but I've annotated them as `@Transient` so that JPA does not try to write them to a column that doesn't exist yet 😁 
* Performance .... hmmmm ... I think we're going to need to put a caching abstraction over the new `HmppsAuthClient`. This is because now every single request, even a `GET` request, will go off to HMPPS Auth to get the active caseload ID. 
Using this approach, if we don't implement caching, we are going to be making a lot of calls to HMPPS Auth, and the reality is that the data for each user wont change frequently. So caching it makes sense.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -21,6 +21,7 @@ ext["mapstruct.version"] = "1.5.5.Final"
 ext["postgresql.version"] = "42.6.0"
 ext["kotlin.logging.version"] = "3.0.5"
 ext["springdoc.openapi.version"] = "2.1.0"
+ext["wiremock.version"] = "3.0.0-beta-10"
 
 allOpen {
   annotations(
@@ -67,6 +68,7 @@ dependencies {
 
   // Integration test dependencies
   integrationTestImplementation("com.h2database:h2")
+  integrationTestImplementation("com.github.tomakehurst:wiremock:${property("wiremock.version")}")
   integrationTestImplementation(testFixtures(project("domain:goal")))
   integrationTestImplementation(testFixtures(project("domain:timeline")))
 
--- a/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/IntegrationTestBase.kt
+++ b/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/IntegrationTestBase.kt
@@ -6,6 +6,7 @@ import org.springframework.boot.test.context.SpringBootTest
 import org.springframework.boot.test.context.SpringBootTest.WebEnvironment.RANDOM_PORT
 import org.springframework.test.context.ActiveProfiles
 import org.springframework.test.web.reactive.server.WebTestClient
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.config.WiremockService
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.repository.ActionPlanRepository
 import java.security.KeyPair
 
@@ -23,8 +24,19 @@ abstract class IntegrationTestBase {
   @Autowired
   lateinit var keyPair: KeyPair
 
+  @Autowired
+  lateinit var wireMockService: WiremockService
+
   @BeforeEach
   fun clearDatabase() {
     actionPlanRepository.deleteAll() // Will also remove all Goals and Steps due to cascade
+  }
+
+  @BeforeEach
+  fun resetWireMock() {
+    with(wireMockService) {
+      resetAllStubsAndMappings()
+      stubHmppsAuthUserMe()
+    }
   }
 }
--- a/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/config/WiremockConfiguration.kt
+++ b/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/config/WiremockConfiguration.kt
@@ -1,0 +1,47 @@
+package uk.gov.justice.digital.hmpps.educationandworkplanapi.app.config
+
+import com.github.tomakehurst.wiremock.WireMockServer
+import com.github.tomakehurst.wiremock.core.WireMockConfiguration.options
+import com.github.tomakehurst.wiremock.http.Request
+import com.github.tomakehurst.wiremock.http.Response
+import mu.KotlinLogging
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.boot.test.util.TestPropertyValues
+import org.springframework.context.ConfigurableApplicationContext
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+
+private val logger = KotlinLogging.logger {}
+
+@Configuration
+class WiremockConfiguration {
+
+  @Bean
+  fun wireMockServer(
+    applicationContext: ConfigurableApplicationContext,
+    @Value("\${logWiremockRequests:false}") logWiremockRequests: Boolean,
+  ): WireMockServer =
+    WireMockServer(
+      options()
+        .dynamicPort(),
+    ).apply {
+      if (logWiremockRequests) {
+        addMockServiceRequestListener { request: Request, _: Response ->
+          val formattedHeaders = request.headers.all().joinToString("\n") {
+            "${it.key()}: ${it.values().joinToString(", ")}"
+          }
+          val logMessage = StringBuilder()
+            .appendLine("Request sent to wiremock:")
+            .appendLine("${request.method} ${request.absoluteUrl}")
+            .appendLine(formattedHeaders)
+            .appendLine()
+            .appendLine(request.bodyAsString)
+          logger.info { logMessage }
+        }
+      }
+      start()
+      TestPropertyValues.of(
+        "api.hmpps-auth.url=http://localhost:${this.port()}/auth",
+      ).applyTo(applicationContext)
+    }
+}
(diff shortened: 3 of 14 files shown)


build.gradle.kts
```diff
@@ -21,6 +21,7 @@ ext["mapstruct.version"] = "1.5.5.Final"
 ext["postgresql.version"] = "42.6.0"
 ext["kotlin.logging.version"] = "3.0.5"
 ext["springdoc.openapi.version"] = "2.1.0"
+ext["wiremock.version"] = "3.0.0-beta-10"
```
nathanrussell-moj-digital, line 24: Unfortunately the non-beta stream of wiremock (version 2.xx) no longer works with Spring Boot 3.
You have to use a version `3.0.0-beta-2` or above 😢 
`3.0.0-beta-10` is the latest and [based on the frequency of beta releases](https://mvnrepository.com/artifact/com.github.tomakehurst/wiremock) this is clearly active development so I'd like to think there is a formal v3 release soon 🤞

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/IntegrationTestBase.kt
```diff
@@ -23,8 +24,19 @@ abstract class IntegrationTestBase {
...
   @BeforeEach
   fun clearDatabase() {
     actionPlanRepository.deleteAll() // Will also remove all Goals and Steps due to cascade
   }
+
+  @BeforeEach
+  fun resetWireMock() {
+    with(wireMockService) {
+      resetAllStubsAndMappings()
+      stubHmppsAuthUserMe()
```
nathanrussell-moj-digital, line 39: I've added the stub for the HMPPS Auth 'me' endpoint in the before each, so that all integration tests automatically get it (they will need it - even the `GET` requests)

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/config/WiremockConfiguration.kt
```diff
@@ -0,0 +1,47 @@
...
+import mu.KotlinLogging
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.boot.test.util.TestPropertyValues
+import org.springframework.context.ConfigurableApplicationContext
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+
+private val logger = KotlinLogging.logger {}
+
+@Configuration
```
nathanrussell-moj-digital, line 16: Hmmm, looks strangely familiar! 😉

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/config/WiremockService.kt
```diff
@@ -0,0 +1,43 @@
...
+  ) {
+    wireMockServer.stubFor(
+      WireMock.get(WireMock.urlPathMatching("/auth/api/user/me"))
+        .willReturn(
+          ResponseDefinitionBuilder.responseDefinition()
+            .withStatus(200)
+            .withHeader("Content-Type", "application/json")
+            .withBody(
+              """
+                {
```
nathanrussell-moj-digital, line 28: This is the very simple object that HMPPS Auth `/api/user/me` returns - for the purpose of what we are doing in this PR we only care about the `activeCaseLoadId` field

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/config/WiremockService.kt
```diff
@@ -0,0 +1,43 @@
...
+class WiremockService(private val wireMockServer: WireMockServer) {
+
+  fun resetAllStubsAndMappings() {
+    wireMockServer.resetAll()
+  }
+
+  fun stubHmppsAuthUserMe(
+    username: String = "auser_gen",
+    displayName: String = "Albert User",
+    activeCaseLoadId: String = "BXI",
```
nathanrussell-moj-digital, line 18: `BXI` is Brixton

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/client/HmppsAuthClient.kt
```diff
@@ -0,0 +1,27 @@
...
+    val userProfile = hmppsAuthWebClient
+      .get()
+      .uri("/api/user/me")
+      .bearerToken(userToken)
+      .retrieve()
+      .bodyToMono(Map::class.java)
+      .block()
+
+    return userProfile?.get("activeCaseLoadId")?.toString()
+  }
```
nathanrussell-moj-digital, line 26: I've not added any error handling to this at all yet - so if HMPPS Auth returns a non-200, all manner of badness is likely to happen! Not really tested it yet, but I guess we need to think about what that might mean for our user journey / how we want to handle it

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/config/ResourceServerConfiguration.kt
```diff
@@ -35,7 +35,7 @@ class ResourceServerConfiguration {
         ).forEach { authorize(it, permitAll) }
         authorize(anyRequest, authenticated)
       }
-      oauth2ResourceServer { jwt { jwtAuthenticationConverter = AuthAwareTokenConverter() } }
+      oauth2ResourceServer { jwt { jwtAuthenticationConverter = tokenConverter } }
```
nathanrussell-moj-digital, line 38: I'm not sure why we were new'ing the `AuthAwareTokenConverter` up here rather than injecting it, but seeing as the `AuthAwareTokenConverter` now has a constructor arg (the HMPPS Auth Client) we need to inject it rather than new it up.

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/LocationAuditingEntityListener.kt
```diff
@@ -0,0 +1,64 @@
...
+
+import jakarta.persistence.PrePersist
+import jakarta.persistence.PreUpdate
+import java.lang.reflect.Field
+
+/**
+ * JPA Entity Listener that sets fields annotated with [CreatedLocation] or [LastModifiedLocation]] with the authenticated
+ * user's location, typically the prison in which they are working taken from their active caseload ID.
+ */
+class LocationAuditingEntityListener {
```
nathanrussell-moj-digital, line 11: This is exactly the same (copy and paste) of our existing `DisplayNameAuditingEntityListener`
Perhaps we can do something smarter by somehow combining the 2, but for the purpose of explaining the POC having it as a separate class probably makes it easier to understand 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/LocationAuditingEntityListener.kt
```diff
@@ -0,0 +1,64 @@
...
+
+  /**
+   * Entity lifecycle hook to update fields annotated with [CreatedLocation] or [LastModifiedLocation]
+   * on new entity creation (pre-persist)
+   */
+  @PrePersist
+  fun updateAuditDisplayNameFieldsOnEntityCreate(target: Any) {
```
mattwills-moj-digital, line 18: Typo 😱

PR #81, opened 2023-09-01
RR-275 Get Action Plan Summaries - controller and ITs

mattwills-moj-digital:
This PR adds the REST controller implementation and corresponding integration tests.


src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/GetActionPlanSummariesTest.kt
```diff
@@ -0,0 +1,155 @@
...
+      .expectStatus()
+      .isOk
+      .returnResult(ActionPlanSummaryListResponse::class.java)
+
+    // Then
+    val actual = response.responseBody.blockFirst()
+    assertThat(actual).hasEmptySummaries()
+  }
+
+  @Test
```
nathanrussell-moj-digital, line 82: Minor, and perhaps its not an issue; but we know that integration tests are slow - do we need 3 happy path tests?
I think we need the one that returns many summaries .... _maybe_ we need the empty collection one? (not sure); but I'm not sure we need the one that returns 1 summary .... What do you think? Does the service tests cover returning 0, 1 and many ?

mattwills-moj-digital: Happy to remove the single one 👍

nathanrussell-moj-digital (dismissed):
1 minor question about the integration tests; but happy to go with it if thats what we feel is right.
LGTM 👍

nathanrussell-moj-digital (approved):

PR #87, opened 2023-09-08
RR-308 - App Insights event for Updating a Goal

nathanrussell-moj-digital:
This PR adds the Update Goal event to the `TelemtryService` and wires it into the controller for the update goal request.

I've added awaitility, and added assertions that the events were sent in the relevant integration tests.
Strictly speaking there is no need for awaitility at this point as it is all synchronous, but my next PR will introduce an AOP aspect and an async `Future` or similar to send the events, rather than coupling it into the controller.

### DIFF
--- a/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/IntegrationTestBase.kt
+++ b/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/IntegrationTestBase.kt
@@ -1,17 +1,28 @@
 package uk.gov.justice.digital.hmpps.educationandworkplanapi.app
 
+import com.microsoft.applicationinsights.TelemetryClient
+import org.awaitility.Awaitility
 import org.junit.jupiter.api.BeforeEach
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.boot.test.context.SpringBootTest
 import org.springframework.boot.test.context.SpringBootTest.WebEnvironment.RANDOM_PORT
+import org.springframework.boot.test.mock.mockito.SpyBean
 import org.springframework.test.context.ActiveProfiles
 import org.springframework.test.web.reactive.server.WebTestClient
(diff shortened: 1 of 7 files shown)


src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/IntegrationTestBase.kt
```diff
@@ -1,25 +1,39 @@
...
+  }
+
   @Suppress("SpringJavaInjectionPointsAutowiringInspection")
   @Autowired
   lateinit var webTestClient: WebTestClient
 
   @Autowired
   lateinit var actionPlanRepository: ActionPlanRepository
 
+  @SpyBean
```
nathanrussell-moj-digital, line 34: I'm not a fan of spy beans, but in this case its the only way I could realistically do it - `TelemetryClient` is not our class, its a bean that is part of MS App Insights, so _what_ it does and _how_ it does it I dont really care. What we want to do is spy on it to make sure the relevant method was called (which you'll see in the integration tests below)

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/CreateGoalTest.kt
```diff
@@ -151,6 +153,15 @@ class CreateGoalTest : IntegrationTestBase() {
       .hasTargetDateRange(TargetDateRangeEntity.ZERO_TO_THREE_MONTHS)
       .hasStatus(StepStatus.NOT_STARTED)
       .wasCreatedBy(dpsUsername)
+
+    val expectedEventCustomDimensions = mapOf(
+      "status" to "ACTIVE",
+      "stepCount" to "1",
+      "reference" to goal.reference.toString(),
+    )
+    await.untilAsserted {
```
nathanrussell-moj-digital, line 162: `await` is not strictly necessary here as its currently sync; but this does allow me to change the implementation to an async approach and not have to worry about the integration tests

mattwills-moj-digital (approved):
AToAB 👍

PR #88, opened 2023-09-09
RR-308 - Decouple sending App Insights events from the controller

nathanrussell-moj-digital:
This PR decouples sending App Insights events from the controller, and send them asynchronously

It does this by refactoring the `GoalService` so that it injects a `GoalServiceDecorator` instance, where the default implementation is a no-op impl. The decorator interface defines a "before" and "after" method for each service method.

In this PR a custom `GoalServiceDecorator` is implemented that calls the telemtry service methods to track events. It does this asynchronously using kotlin coroutines.

### DIFF
--- a/domain/goal/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/goal/service/GoalService.kt
+++ b/domain/goal/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/goal/service/GoalService.kt
@@ -17,18 +17,25 @@ private val log = KotlinLogging.logger {}
  * [GoalPersistenceAdapter].
  *
  * This class is deliberately final so that it cannot be subclassed, ensuring that the business rules stay within the
- * domain.
+ * domain. Service method behaviour can however be customized and extended by creating an instance of this class with a
+ * [GoalServiceDecorator]. If not specified, the default implementation is a [NoopGoalServiceDecorator].
+ *
  */
 class GoalService(
   private val persistenceAdapter: GoalPersistenceAdapter,
+  private val goalServiceDecorator: GoalServiceDecorator = NoopGoalServiceDecorator(),
 ) {
 
   /**
    * Creates a new [Goal] for the prisoner identified by their prison number, with the data in the specified [CreateGoalDto]
    */
   fun createGoal(prisonNumber: String, createGoalDto: CreateGoalDto): Goal {
+    goalServiceDecorator.beforeCreateGoal(prisonNumber, createGoalDto)
     log.info { "Creating new Goal for prisoner [$prisonNumber]" }
     return persistenceAdapter.createGoal(prisonNumber, createGoalDto)
+      .also {
+        goalServiceDecorator.afterCreateGoal(prisonNumber, createGoalDto, it)
+      }
   }
 
   /**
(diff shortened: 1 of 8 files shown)


domain/goal/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/goal/service/GoalService.kt
```diff
@@ -17,27 +17,38 @@ private val log = KotlinLogging.logger {}
...
  *
  * This class is deliberately final so that it cannot be subclassed, ensuring that the business rules stay within the
- * domain.
+ * domain. Service method behaviour can however be customized and extended by creating an instance of this class with a
+ * [GoalServiceDecorator]. If not specified, the default implementation is a [NoopGoalServiceDecorator].
+ *
  */
 class GoalService(
   private val persistenceAdapter: GoalPersistenceAdapter,
+  private val goalServiceDecorator: GoalServiceDecorator = NoopGoalServiceDecorator(),
```
nathanrussell-moj-digital, line 26: It uses the no-op decorator by default, unless a different implementation is passed in on the constructor.

domain/goal/src/test/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/goal/service/GoalServiceTest.kt
```diff
@@ -17,13 +17,18 @@ import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.goal.aValidGo
...
-  @Mock
-  private lateinit var persistenceAdapter: GoalPersistenceAdapter
+  private var persistenceAdapter: GoalPersistenceAdapter = mock()
+
+  private var goalServiceDecorator: GoalServiceDecorator = mock()
+
+  @BeforeAll
+  fun setupService() {
+    service = GoalService(persistenceAdapter, goalServiceDecorator)
+  }
```
nathanrussell-moj-digital, line 31: Because the `GoalService` has an optional constructor arg (for the decorator), as far as the java byte code is concerned the class has 2 constructors (one for just the persistence adapter, and one for the persistence adapter and the decorator)
Because there are 2 constructors, InjectMocks doesn't work and mockito complains bitterly that it doesnt know which constructor to use! 
Therefore we have to construct the instance manually

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/AsyncGoalServiceDecorator.kt
```diff
@@ -0,0 +1,34 @@
...
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.goal.dto.UpdateGoalDto
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.goal.service.GoalServiceDecorator
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.goal.service.NoopGoalServiceDecorator
+
+/**
+ * Implementation of [GoalServiceDecorator] that decorates Goal Service methods with additional behaviours that are
+ * invoked asynchronously.
+ */
+@Component
+class AsyncGoalServiceDecorator(private val telemetryService: TelemetryService) : NoopGoalServiceDecorator() {
```
nathanrussell-moj-digital, line 17: I've not written a unit test for this class yet - I can do if we decide its the right design approach

mattwills-moj-digital: You didn't write the test first?! ;-)

Minor point, but I'm not sure about extending the Noop version. I would just implement the original interface :)

domain/goal/src/test/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/goal/service/GoalServiceTest.kt
mattwills-moj-digital: Minor, but I think this is the way we'd normally address this:

```
@ExtendWith(MockitoExtension::class)
class GoalServiceTest {
  private lateinit var service: GoalService

  @Mock
  private lateinit var persistenceAdapter: GoalPersistenceAdapter

  @Mock
  private lateinit var goalServiceDecorator: GoalServiceDecorator

  @BeforeEach
  fun setupService() {
    service = GoalService(persistenceAdapter, goalServiceDecorator)
  }
```

It's not necessarily better and in the absence of any internal state within the service class, you could argue that `Lifecycle.PER_CLASS` is more efficient, but it's a smaller change :)

nathanrussell-moj-digital:
Closing due to alternative (better) design approach in https://github.com/ministryofjustice/hmpps-education-and-work-plan-api/pull/90

PR #91, opened 2023-09-11
RR-309 - Initial Event related services

mattwills-moj-digital:
This PR makes the required changes to capture `Timeline` events, whenever an `ActionPlan` or `Goal` is created, or whenever an Goal, including its Steps, is updated.

The key changes are to the `TImelineEvent` class and the `AsyncGoalEventService` and `AsyncActionPlanEventService` classes, which in turn use `TimelineEventFactory` to create the required `TimelineEvent` classes.

The required entity and database changes will be added under RR-317.


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/event/AsyncGoalEventService.kt
```diff
@@ -0,0 +1,88 @@
...
+    runBlocking {
+      log.debug { "Goal updated event for prisoner [$prisonNumber]" }
+      // val events = resolveUpdateEvents(updatedGoal, existingGoal)
+      launch {
+        telemetryService.trackGoalUpdateEvent(updatedGoal)
+      }
+    }
+  }
+
+  private fun resolveUpdateEvents(updatedGoal: Goal, existingGoal: Goal): List<GoalEvent> {
```
mattwills-moj-digital, line 41: I'll probably refactor this out into a class called something like `GoalUpdatedEventResolver`.

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/event/GoalEvent.kt
```diff
@@ -0,0 +1,28 @@
+package uk.gov.justice.digital.hmpps.educationandworkplanapi.app.event
+
+import java.util.UUID
+
+/**
+ * Represents an event that has occurred to a Goal, or to an entity that's closely related to a Goal (such as one of its
+ * Steps, or its parent Action Plan). Because of this, the `reference` field could belong to a Goal, Step or
+ * Action Plan, however strictly speaking, `Event` may have been a better name, but it feels too vague.
+ */
+data class GoalEvent(
```
mattwills-moj-digital, line 10: I'm flipping between `Event` and `GoalEvent` 🤔

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/event/AsyncActionPlanEventService.kt
```diff
@@ -0,0 +1,19 @@
...
+private val log = KotlinLogging.logger {}
+
+/**
+ * Implementation of [ActionPlanEventService] for performing additional asynchronous actions related to [ActionPlan]
+ * events.
+ */
+@Component
+class AsyncActionPlanEventService() : ActionPlanEventService {
+  override fun actionPlanCreated(actionPlan: ActionPlan) {
+    log.info { "ActionPlan created event for prisoner [${actionPlan.prisonNumber}]" }
```
mattwills-moj-digital, line 17: WIP - the plan is to call a `TimelineEventService` with an `ACTION_PLAN_CREATED` `GoalEvent` (or `Event`?!)

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/event/AsyncGoalEventService.kt
```diff
@@ -0,0 +1,88 @@
...
+      }
+    }
+  }
+
+  override fun goalUpdated(prisonNumber: String, updatedGoal: Goal, existingGoal: Goal) {
+    runBlocking {
+      log.debug { "Goal updated event for prisoner [$prisonNumber]" }
+      // val events = resolveUpdateEvents(updatedGoal, existingGoal)
+      launch {
+        telemetryService.trackGoalUpdateEvent(updatedGoal)
```
mattwills-moj-digital, line 36: WIP - the plan is to call a `TimelineEventService` with a list of relevant `GoalEvent`s

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/event/AsyncGoalEventService.kt
```diff
@@ -0,0 +1,88 @@
...
+ * Implementation of [GoalEventService] for performing additional asynchronous actions related to [Goal] events.
+ */
+@Component
+class AsyncGoalEventService(private val telemetryService: TelemetryService) : GoalEventService {
+
+  override fun goalCreated(prisonNumber: String, createdGoal: Goal) {
+    runBlocking {
+      log.debug { "Goal created event for prisoner [$prisonNumber]" }
+      launch {
+        telemetryService.trackGoalCreateEvent(createdGoal)
```
mattwills-moj-digital, line 26: WIP - the plan is to call a `TimelineEventService` with a `GOAL_CREATED` `Event`.

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/event/GoalEvent.kt
mattwills-moj-digital: I've ended opting for `PlpEvent`.

domain/timeline/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/timeline/Timeline.kt
```diff
@@ -10,7 +10,7 @@ class Timeline(
 ) {
 
   val events: MutableList<TimelineEvent>
-    get() = field.also { events -> events.sortByDescending { it.eventDateTime } }
+    get() = field.also { events -> events.sortBy { it.timestamp } }
```
mattwills-moj-digital, line 13: The Timeline events should be ordered in ascending chronological order.

domain/timeline/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/timeline/TimelineEvent.kt
```diff
@@ -1,15 +1,82 @@
...
+   */
+  val sourceReference: String,
+  /**
+   * The type of the event - see [TimelineEventType].
+   */
+  val eventType: TimelineEventType,
+  /**
+   * Some useful contextual information about the event. For example a reason for the event.
+   */
+  val contextualInfo: String?,
```
mattwills-moj-digital, line 28: I'm not sure about adding this field (if at all). I might remove it as I don't think we'll use it for MVP.

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/JpaTimelinePersistenceAdapter.kt
```diff
@@ -0,0 +1,29 @@
...
+@Component
+class JpaTimelinePersistenceAdapter(
+  // private val timelineRepository: TimelineRepository,
+  // private val timelineMapper: TimelineEntityMapper,
+) : TimelinePersistenceAdapter {
+
+  @Transactional
+  override fun recordTimelineEvent(prisonNumber: String, event: TimelineEvent) {
+    log.info { "Recording TimelineEvent [$event] for prisoner [$prisonNumber]" }
+    // TODO RR-317
```
mattwills-moj-digital, line 20: I'll do all the JPA and database stuff in a separate PR (for which I've created a separate sub-task).

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/event/AsyncGoalEventService.kt
```diff
@@ -0,0 +1,55 @@
...
+  private val timelineEventResolver: TimelineEventResolver,
+  private val timelineService: TimelineService,
+) : GoalEventService {
+
+  override fun goalCreated(prisonNumber: String, createdGoal: Goal) {
+    runBlocking {
+      log.debug { "Goal created event for prisoner [$prisonNumber]" }
+      launch {
+        telemetryService.trackGoalCreateEvent(createdGoal)
+        timelineService.recordTimelineEvent(prisonNumber, goalCreatedTimelineEvent(createdGoal))
```
mattwills-moj-digital, line 31: I've decided against having a separate generic event, which in theory could have been used for either the Timeline or App Insights.

In reality, the events would have only existed in this class before being mapped to the `Timeline` domain (in order to pass to the `TimelineService`).

domain/goal/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/goal/service/GoalEventService.kt
```diff
@@ -10,10 +10,10 @@ interface GoalEventService {
...
    * Implementations providing custom code for when a [Goal] is created.
    */
-  fun goalCreated(createdGoal: Goal)
+  fun goalCreated(prisonNumber: String, createdGoal: Goal)
 
   /**
    * Implementations providing custom code for when a [Goal] is updated.
    */
-  fun goalUpdated(updatedGoal: Goal)
+  fun goalUpdated(prisonNumber: String, updatedGoal: Goal, existingGoal: Goal)
```
nathanrussell-moj-digital, line 18: Really really minor; and dont feel you have to do this, but I wonder what you think about changing the order of args and the name of the "existing" goal - something like this perhaps?
```
fun goalUpdated(prisonNumber: String, previousGoal: Goal, updatedGoal: Goal)
```
or
```
fun goalUpdated(prisonNumber: String, goalPreUpdate: Goal, goalPostUpdate: Goal)
```
or something like that ... you get the gist ...

mattwills-moj-digital: It's very important, not minor! 😉 

Funnily enough, I looked at the order of the arguments last night and thought the existing/previous goal should come first.

I'm not sure about `previous`, because it's the same goal, but I'll have a think about the name 👍

domain/timeline/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/timeline/Timeline.kt
nathanrussell-moj-digital: I agree with that from a domain / data point of view 👍 

I think the prototype screen however is wrong/misleading in this respect. I suspect once we develop the UI (post MVP) , if we have to do it as per the current prototype, they will soon change their mind 😁 because (in its current form of earliest event first) it means the CIAG will likely have to scroll to the bottom of the page to find the stuff they are really interested in (ie. I think they'll find the most recent goals/events more useful than the earlier stuff)

Clearly this is a UI concern though (how they want it rendered). So I agree with your change here that from a domain/data POV they should be ordered in ascending order; but I suspect either the UI or the controller code will re-sort the events based on how the UI wants to render them

domain/timeline/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/timeline/TimelineEvent.kt
```diff
@@ -1,15 +1,82 @@
...
+   */
+  val prisonId: String,
+  /**
+   * The ID of the logged-in user who created the event
+   */
+  val createdBy: String,
+  /**
+   * The full name (for display purposes) of the logged-in user who created the event
+   */
+  val createdByDisplayName: String,
```
nathanrussell-moj-digital, line 40: This might have to be optional, as I don't believe CIAG capture or hold the user's display name in their system

domain/goal/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/goal/service/GoalEventService.kt
mattwills-moj-digital: Actually, I've gone with previous, since it was the previous state of the goal at this point in time!

domain/timeline/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/timeline/service/TimelineService.kt
```diff
@@ -16,6 +17,23 @@ class TimelineService(
...
+  fun recordTimelineEvent(prisonNumber: String, event: TimelineEvent) {
+    persistenceAdapter.recordTimelineEvent(prisonNumber, event)
+  }
+
+  /**
+   * Records a collection of [TimelineEvent]s that have taken place for a prisoner.
+   */
+  fun recordTimelineEvents(prisonNumber: String, events: List<TimelineEvent>) {
+    // TODO RR-314 - write test
+    events.forEach {
```
nathanrussell-moj-digital, line 32: Just thinking out load here, and I could be wrong ....
But I wonder if the loop should in the persistence adapter method impl, rather than here in the service?
IE. have a method in the persistence adapter that takes the collection. The implementation would be annotated `@Transactional` and would do the looping (or perhaps a `saveAll` type method call?)

Anyway, I guess my worry is about TX boundaries. As you have it here the TX boundary would be around each save, so you might end up with some of the events saved, but not all of them.
If we do it as I've suggested the TX boundary would be around all of the timeline events, so if one fail, they all rollback 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/JpaGoalPersistenceAdapter.kt
```diff
@@ -41,7 +41,8 @@ class JpaGoalPersistenceAdapter(
       actionPlanRepository.saveAndFlush(this)
     }
 
-    return goalMapper.fromEntityToDomain(goalEntity)
+    val persisted = actionPlanEntity.goals!!.first { it.reference == goalEntity.reference }
```
nathanrussell-moj-digital, line 44: Just out of interest was there a problem/bug with the original impl? Or is this change just a code tidying / making it easier to read exercise ?

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/event/AsyncActionPlanEventService.kt
```diff
@@ -0,0 +1,39 @@
...
+  override fun actionPlanCreated(actionPlan: ActionPlan) {
+    runBlocking {
+      log.info { "ActionPlan created event for prisoner [${actionPlan.prisonNumber}]" }
+      launch {
+        timelineService.recordTimelineEvent(actionPlan.prisonNumber, actionPlanCreatedEvent(actionPlan))
+      }
+    }
+  }
+
+  private fun actionPlanCreatedEvent(actionPlan: ActionPlan) =
```
nathanrussell-moj-digital, line 31: This works for now, but I wonder once we start creating a number of different types of TimelineEvents I wonder if a mapstruct mapper or similar might be useful?

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/event/AsyncGoalEventService.kt
nathanrussell-moj-digital, line 31: I think you'll want these in separate `launch` blocks:
```
      launch {
        telemetryService.trackGoalCreateEvent(createdGoal)
      }
      launch {
        timelineService.recordTimelineEvent(prisonNumber, goalCreatedTimelineEvent(createdGoal))
      }
```

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/event/AsyncGoalEventService.kt
```diff
@@ -0,0 +1,55 @@
...
+    }
+  }
+
+  override fun goalUpdated(prisonNumber: String, updatedGoal: Goal, existingGoal: Goal) {
+    runBlocking {
+      log.debug { "Goal updated event for prisoner [$prisonNumber]" }
+      val timelineEvents = timelineEventResolver.resolveGoalUpdatedEvents(updatedGoal, existingGoal)
+      launch {
+        telemetryService.trackGoalUpdateEvent(updatedGoal)
+        timelineService.recordTimelineEvents(prisonNumber, timelineEvents)
```
nathanrussell-moj-digital, line 42: As above; separate `launch` blocks I think 🤔

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/event/TimelineEventResolver.kt
```diff
@@ -0,0 +1,109 @@
...
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.timeline.TimelineEvent
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.timeline.TimelineEventType
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.timeline.TimelineEventType.GOAL_UPDATED
+import java.util.UUID
+
+/**
+ * Responsible for identifying which [TimelineEvent]s have occurred, for example following an update to a [Goal].
+ */
+@Component
+class TimelineEventResolver {
```
nathanrussell-moj-digital, line 17: Oh, so this resolver is only responsible for working out and resolving **update** type events? Is that right?
I was wondering why the other methods that create TimelineEvents (that I've previously commented might suit a mapper?) were not in here .... 🤔

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/JpaGoalPersistenceAdapter.kt
mattwills-moj-digital: It's to ensure the returned goal has the "JPA" fields populated.

domain/timeline/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/timeline/TimelineEvent.kt
mattwills-moj-digital: mmm... that could be a problem. I think we'll still need to get it somehow.

domain/timeline/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/timeline/TimelineEvent.kt
mattwills-moj-digital: Agreed to keep it as I think we'll need it at some stage and it will save flyway scripts etc later.

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/TimelineEventFactory.kt
```diff
@@ -0,0 +1,131 @@
...
+      updatedGoal.notes != previousGoal.notes
+
+  private fun hasGoalStatusChanged(previousGoal: Goal, updatedGoal: Goal) =
+    updatedGoal.status != previousGoal.status
+
+  private fun hasStepBeenUpdated(previousStep: Step?, updatedStep: Step) =
+    previousStep != null && (
+      updatedStep.title != previousStep.title ||
+        updatedStep.targetDateRange != previousStep.targetDateRange ||
+        updatedStep.sequenceNumber != previousStep.sequenceNumber
```
mattwills-moj-digital, line 95: I'm in two minds about including `sequenceNumber` here. If say two steps are re-ordered then this will result in two 'STEP_UPDATED' events, which may not seem right.

Having said that, the UI can always decide to filter this and perhaps just report that 'a' step was updated.

domain/timeline/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/timeline/TimelineEvent.kt
```diff
@@ -1,15 +1,83 @@
...
+      createdByDisplayName = createdByDisplayName,
+      timestamp = Instant.now(),
+    )
+  }
+}
+
+/**
+ * The events that the business are interested in (for example to display a history of these events on screen).
+ *
+ * These are currently limited to CIAG induction and PLP related events, but could be expanded in future.
```
nathanrussell-moj-digital, line 70: 👍 
(and another good reason why createdByDisplayName is optional - not only do we know that CIAG dont have it currently, but as and when/if this is expanded in future for events from other sources it's likely those other sources wont have a displayName either)

nathanrussell-moj-digital (approved):
ATOaB 👍 
GS!

PR #97, opened 2023-09-18
RR-305 - Track a Goal's notes character length in App Insights

nathanrussell-moj-digital:
This PR tracks the number of characters used in the `notes` field of  `Goal` in App Insights events for both Create and Update.

We already have code that tracks the `Goal` Create and Update events, so this PR is simply a case of adding the field to the event's custom dimensions 👍


mattwills-moj-digital (commented):
Seems an odd requirement, but that's not your fault! Otherwise LGTM.

nathanrussell-moj-digital:
> Seems an odd requirement, but that's not your fault! Otherwise LGTM.

I agree!, seems a very odd requirement / solution. What they want to do is to be able to monitor and measure how useful the notes field is; and the way they've decided to do that is by tracking the number of characters that people use in the Goal !
Meh! 🤷‍♂️

PR #108, opened 2023-09-28
RR-307 - TelemetryEventType now responsible for the events custom dimensions

nathanrussell-moj-digital:
This PR makes a minor refactoring to the enum `TelemetryEventType` so that it is responsible for generating the custom dimensions map, rather than doing it in `TelemetryService`

Previously the enum `TelemetryEventType` knew about and was responsible for the App Insights customEvent name (the `value` field), but the `TelemetryService` knew about how to generate the custom dimensions for each event type. This felt wrong.

This PR adds a field to the enum `TelemetryEventType` which is a function that returns the custom dimensions specific for that event type. It feels right, in that the enum now knows about everything it needs in order to send the App Insights telemetry event. It knows about the customEvent name, and also the map of custom dimensions 👍

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/TelemetryEventType.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/TelemetryEventType.kt
@@ -1,18 +1,58 @@
 package uk.gov.justice.digital.hmpps.educationandworkplanapi.app.service
 
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.goal.Goal
+
 /**
- * An enumeration of the types of Goal events that can be sent to the Telemetry Service.
+ * An enumeration of the types of Goal events that can be sent to the Telemetry Service, where each item has:
+ *   * a text field `value` that is the name of the customEvent recorded in App Insights.
+ *   * a function `customDimensions` that returns the customDimensions that are recorded with the App Insights event.
  */
-enum class TelemetryEventType(val value: String) {
-  GOAL_CREATED("goal-created"),
-  GOAL_UPDATED("goal-updated"),
-  GOAL_STARTED("goal-started"),
-  GOAL_COMPLETED("goal-completed"),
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/TelemetryEventType.kt
```diff
@@ -1,18 +1,58 @@
...
+    "goal-updated",
+    {
+      mapOf(
+        "reference" to it.reference.toString(),
+        "notesCharacterCount" to (it.notes?.length ?: 0).toString(),
+      )
+    },
+  ),
+
+  GOAL_STARTED("goal-started", { emptyMap() }),
```
nathanrussell-moj-digital, line 33: All these event types with an emptyMap of custom dimensions are event types that we are not currently sending (but are likely to be asked to; at which point we can implement the required custom dimensions for that particular event type 👍 )

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/TelemetryService.kt
```diff
@@ -45,38 +45,6 @@ class TelemetryService(
...
-   * Returns a map of data representing the custom dimensions for the `step-remove` event.
-   */
-  private fun stepRemoveEventCustomDimensions(goal: Goal): Map<String, String> =
-    with(goal) {
-      mapOf(
-        "reference" to reference.toString(),
-        "stepCount" to steps.size.toString(),
-      )
-    }
+  private fun sendTelemetryEventForGoal(goal: Goal, telemetryEventType: TelemetryEventType) =
```
nathanrussell-moj-digital, line 48: This is the one and only place where we actually send the tracking event via the `telemetryClient`. Using the passed in `TelemetryEventType` it has everything it needs in order to send a telemetry tracking event to App Insights 👍

mattwills-moj-digital (approved):

PR #110, opened 2023-10-02
RR-307 - Send correlationId with telemetry events

nathanrussell-moj-digital:


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/TelemetryService.kt
```diff
@@ -17,34 +18,47 @@ class TelemetryService(
   private val telemetryEventTypeResolver: TelemetryEventTypeResolver,
 ) {
 
-  fun trackGoalCreatedEvent(goal: Goal) {
-    sendTelemetryEventForGoal(goal, GOAL_CREATED)
+  fun trackGoalCreatedEvent(createdGoal: Goal, correlationId: UUID = UUID.randomUUID()) {
```
mattwills-moj-digital, line 21: I'm just wondering if it would be a good idea to use the same `correlationId` for both timelineEvents and AppInsights one? If so, then we might not want to default it here.

mattwills-moj-digital (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/TelemetryService.kt
nathanrussell-moj-digital: Hi @mattwills-moj-digital - did you want me to look at changing this / addressing this comment (I'm not sure how hard it might be just yet) - or are you happy to merge it as is and we can change it if necessary later?

mattwills-moj-digital: As discussed, we're leaving this for now, but could be something to consider in the future.

PR #114, opened 2023-10-05
RR-385 - Add correlationId to TimelineEvent

mattwills-moj-digital:


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/TimelineEventFactory.kt
```diff
@@ -35,6 +35,8 @@ class TimelineEventFactory {
    */
```
mattwills-moj-digital, line 35: Additionally, I could create events for each of the Goals added when a new ActionPlan is created (within the `actionPlanCreatedEvent()` method which isn't shown here). They could all have the same correlationId and therefore be grouped accordingly.

mattwills-moj-digital: Done.

src/main/resources/db/migration/V2023.10.03.0001__add_parent_ref_to_timeline_event.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE timeline_event ADD COLUMN correlation_id UUID;
```
mattwills-moj-digital, line 1: I'm considering updating each existing event with a random UUID, so that I can then make the column mandatory (NOT NULL). Updating them with a "proper" correlationId, which would group them appropriately would be pretty complex and overkill for test data. Any thoughts on this?

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/CreateActionPlanTest.kt
```diff
@@ -235,20 +233,6 @@ class CreateActionPlanTest : IntegrationTestBase() {
       .hasTitle(createStepRequest.title)
       .hasStatus(StepStatus.NOT_STARTED)
       .wasCreatedBy(dpsUsername)
-
-    // assert timeline event is created successfully
```
mattwills-moj-digital, line 239: I'm relying on the dedicated `GetTimelineTest` to test this functionality. Otherwise these `// Then` sections were becoming massive!

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/TimelineEventFactory.kt
```diff
@@ -9,22 +9,39 @@ import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.goal.StepStat
...
+      buildTimelineEvent(
+        actionPlan.goals[0],
+        actionPlan.reference,
+        TimelineEventType.ACTION_PLAN_CREATED,
+        contextualInfo = null,
+        correlationId = correlationId,
+      ),
     )
+    actionPlan.goals.forEach {
+      events.add(goalCreatedTimelineEvent(it, correlationId))
```
mattwills-moj-digital, line 32: This is where the necessary `GOAL_CREATED` events are created when saving a new `ActionPlan`.

domain/timeline/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/timeline/TimelineEvent.kt
```diff
@@ -42,25 +42,33 @@ data class TimelineEvent(
    * The date and time when the event occurred.
    */
   val timestamp: Instant,
+  /**
+   * An optional correlationId for this and any other TimelineEvents that occurred at the same time (i.e. within same
+   * atomic action). For example, this could be an update to a Goal and one or more of its child Steps.
+   */
+  val correlationId: UUID?,
```
nathanrussell-moj-digital, line 49: Interesting that its optional .... I'm sure I'll discover why as I see the rest of the change in this PR, but I'm wondering when we might ever create a timeline event that does not have a correlation id .... 🤔

domain/timeline/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/timeline/TimelineEvent.kt
```diff
@@ -42,25 +42,33 @@ data class TimelineEvent(
...
 ) {
   companion object {
     fun newTimelineEvent(
+      reference: UUID = UUID.randomUUID(),
```
nathanrussell-moj-digital, line 53: Was this a deliberate change, or was it because you felt it was missing from the method args?
By allowing it to be passed as a method arg it means that the (production) code that is new'ing up a new timeline event could in theory specify it's own reference? 🤔 
Is that what we meant?

domain/timeline/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/timeline/TimelineEvent.kt
```diff
@@ -42,25 +42,33 @@ data class TimelineEvent(
...
       sourceReference: String,
       eventType: TimelineEventType,
       contextualInfo: String? = null,
       prisonId: String,
       actionedBy: String,
       actionedByDisplayName: String? = null,
       timestamp: Instant = Instant.now(),
+      correlationId: UUID = UUID.randomUUID(),
```
nathanrussell-moj-digital, line 61: Re: my comment above, I can 100% see the use case where correlationId needs to be passed in / specified by the caller, but not sure about reference

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/GetTimelineTest.kt
```diff
@@ -148,25 +176,53 @@ class GetTimelineTest : IntegrationTestBase() {
...
+          .hasSourceReference(goal1Reference.toString())
+          .hasContextualInfo("Learn German")
+          .hasCorrelationId(actionPlanCreatedCorrelationId)
+      }
+      .event(3) {
         it.hasEventType(TimelineEventType.GOAL_CREATED)
           .hasPrisonId("BXI")
           .hasSourceReference(goal2Reference.toString())
           .hasContextualInfo("Learn French")
+          .doesNotHaveCorrelationId(actionPlanCreatedCorrelationId)
```
nathanrussell-moj-digital, line 203: I cant think of a better name right now, but `doesNotHaveCorrelationId` made me think initially that you were expecting it to not have a correlation ID at all, rather than "it has a correlationID, but just not _this_ one" 🤔

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/TimelineEventFactory.kt
```diff
@@ -132,6 +155,7 @@ class TimelineEventFactory {
     sourceReference: UUID,
     eventType: TimelineEventType,
     contextualInfo: String?,
+    correlationId: UUID = UUID.randomUUID(),
```
nathanrussell-moj-digital, line 158: Re: my question on correlationID cardinality above .... here in the factory method correlationID is not optional, so assuming this is the only way/place to create a timeline event I dont think it can ever be null can it?

src/main/resources/db/migration/V2023.10.03.0001__add_parent_ref_to_timeline_event.sql
nathanrussell-moj-digital: Ah ha .... here's the cardinality rationale 😁 (I should have waited till I'd read this comment! 🤣 )

Yes, agreed, generate and assign a random UUID for all existing timeline events (we are only in dev anyway), then make the db column non-nullable; then you can make it non-null in the DB entity and domain classes 👍

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -415,6 +415,11 @@ components:
           format: date-time
           description: An ISO-8601 timestamp representing when the event occurred.
           example: '2023-06-19T09:39:44Z'
+        correlationId:
+          type: string
+          format: uuid
+          description: An optional correlationId for this and any other TimelineEvents that occurred at the same time (i.e. within same atomic action). For example, this could be an update to a Goal, plus one or more of its child Steps. If required, this allows a client to easily group the events together.
```
nathanrussell-moj-digital, line 421: I'd be tempted to make correlationID non-nullable

domain/timeline/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/timeline/TimelineEvent.kt
mattwills-moj-digital: I'm trying to recall why I did this! I think it was purely for consistency/symmetry, so that both parameters had `: UUID = UUID.randomUUID()` (i.e. my OCD got the better of me!). Happy to change it back, but I trust the clients in this case not to provide a value unless they need to.

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/GetTimelineTest.kt
mattwills-moj-digital: `.correlationIdIsNotEqualTo()`?

domain/timeline/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/timeline/TimelineEvent.kt
```diff
@@ -42,6 +42,11 @@ data class TimelineEvent(
    * The date and time when the event occurred.
    */
   val timestamp: Instant,
+  /**
+   * An optional correlationId for this and any other TimelineEvents that occurred at the same time (i.e. within same
```
nathanrussell-moj-digital, line 46: Your kdoc is now wrong! 😢

nathanrussell-moj-digital (approved):
LGTM 👍

PR #117, opened 2023-10-12
RR-421 - Domain model for a CIAG Induction

mattwills-moj-digital:
A first iteration of a domain model for a CIAG induction. See some of the comments and kdoc to explain some of the decisions.

### DIFF
--- a/domain/induction/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/induction/InPrisonInterests.kt
+++ b/domain/induction/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/induction/InPrisonInterests.kt
@@ -1,11 +1,19 @@
 package uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.induction
 
+import java.time.Instant
+
 /**
  * Represents any in-prison work or training interests a Prisoner might have during their time in prison.
  */
 data class InPrisonInterests(
   val inPrisonWorkInterests: List<InPrisonWorkInterest>,
   val inPrisonTrainingInterests: List<InPrisonTrainingInterest>,
+  val createdBy: String?,
+  val createdByDisplayName: String?,
+  val createdAt: Instant?,
+  val lastUpdatedBy: String?,
+  val lastUpdatedByDisplayName: String?,
+  val lastUpdatedAt: Instant?,
 )
 
 data class InPrisonWorkInterest(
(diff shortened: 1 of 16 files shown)


domain/induction/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/induction/Induction.kt
```diff
@@ -8,8 +8,41 @@ import java.util.UUID
...
+   */
   val prisonNumber: String,
+  /**
+   * Details of the Prisoner's work aspirations.
+   */
+  val workOnRelease: WorkOnRelease,
+  /**
+   * Qualifications that the Prisoner may have achieved previously.
+   */
+  val previousQualifications: PreviousQualifications,
```
nathanrussell-moj-digital, line 27: I dont know the answer to this (and we might not know/understand the answer) until we've done some more digging or written more code perhaps), but I wonder about the cardinality of these objects as direct children of `Induction`

For example,`InPrisonInterests` here is mandatory - you have to have one of these instances.
Within `InPrisonInterests` are 2 lists which are also mandatory, but could be empty.
So, the question is .... if I have an `Induction` with an `InPrisonInterests` object, where the collection of `InPrisonWorkInterest` is empty .... does that mean I have no interest to work in prison? Or does it mean I wasnt asked it on the induction because it ended up going the short route?  🤔

mattwills-moj-digital: As discussed, I've made the domain fairly lenient in terms of what's mandatory at the top level, as I didn't want to assume which questions have to be asked (except for whether they're hoping to work in the future, which is a key part of an induction). I'm conscious that the screen flow could change and just because we always ask something at the moment, doesn't mean we always will in the future.

However, if a question is asked (for example their education history), then there are some mandatory values (for example their highest level of education). I'm not entirely happy with the domain in this respect, as some of the internal lists can be empty and others can't (for example some of the enums have `NONE` as an option, which has to be provided if the other options are not applicable). I've tried to make this clear in the kdoc, but this is one area that might need improving. Personally I would prefer it if `NONE` or `NOT_SURE` was an option in all of the enums for consistency.

domain/induction/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/induction/InPrisonInterests.kt
```diff
@@ -0,0 +1,62 @@
...
+  ENGLISH_LANGUAGE_SKILLS,
+  FORKLIFT_DRIVING,
+  INTERVIEW_SKILLS,
+  MACHINERY_TICKETS,
+  NUMERACY_SKILLS,
+  RUNNING_A_BUSINESS,
+  SOCIAL_AND_LIFE_SKILLS,
+  WELDING_AND_METALWORK,
+  WOODWORK_AND_JOINERY,
+  OTHER,
```
mattwills-moj-digital, line 61: In some cases they cannot select `NONE`, but have to select `OTHER` and give a reason. I need to understand this better, but I'm not suggesting we change it now.

nathanrussell-moj-digital (approved):
LGTM

PR #119, opened 2023-10-17
RR-422 - Induction related tables and entities

mattwills-moj-digital:
This PR introduces quite a few DB tables and JPA entities to represent a CIAG Induction.

Note that I've created an `InductionRepository` and a corresponding integration test locally to make sure I can save and retrieve an `InductionEntity`. I'll add test data builders in a subsequent PR, but I'll save the integration test until it's hooked up with the REST API.

### DIFF
--- a/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/TempRepositoryTest.kt
+++ b/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/TempRepositoryTest.kt
@@ -1,0 +1,265 @@
+package uk.gov.justice.digital.hmpps.educationandworkplanapi.app
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import org.springframework.transaction.annotation.Transactional
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.aValidPrisonNumber
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.induction.AffectAbilityToWork
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.induction.FutureWorkInterestsEntity
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.induction.HighestEducationLevel
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.induction.HopingToWork
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.induction.InPrisonInterestsEntity
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.induction.InPrisonTrainingInterest
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.induction.InPrisonTrainingType
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.induction.InPrisonWorkInterest
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.induction.InPrisonWorkType
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.induction.InductionEntity
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.induction.InterestType
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.induction.NotHopingToWorkReason
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.induction.PersonalInterestEntity
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.induction.PersonalSkillEntity
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.induction.PersonalSkillsAndInterestsEntity
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.induction.PreviousQualificationsEntity
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.induction.PreviousTrainingEntity
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.induction.PreviousWorkExperiencesEntity
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.induction.QualificationEntity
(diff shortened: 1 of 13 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/entity/induction/FutureWorkInterestsEntity.kt
```diff
@@ -0,0 +1,118 @@
...
+  @CreatedBy
+  var createdBy: String? = null,
+
+  @Column
+  @UpdateTimestamp
+  var updatedAt: Instant? = null,
+
+  @Column
+  @LastModifiedBy
+  var updatedBy: String? = null,
```
mattwills-moj-digital, line 99: Every entity also has these audit type fields, but not the `created/updatedAtPrison` or `displayName` fields.

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/entity/induction/PreviousTrainingEntity.kt
```diff
@@ -0,0 +1,79 @@
...
+  @UuidGenerator
+  var id: UUID? = null,
+
+  @Column(updatable = false)
+  @field:NotNull
+  var reference: UUID? = null,
+
+  @ElementCollection(targetClass = TrainingType::class)
+  @Enumerated(value = EnumType.STRING)
+  @CollectionTable(name = "training_type")
```
mattwills-moj-digital, line 43: This is how I'm representing a list of enums. I've distinguished these "collection tables" in the comments in the Flyway script.

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/entity/induction/InductionNotFoundException.kt
```diff
@@ -0,0 +1,7 @@
+package uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.induction
+
+/**
+ * Thrown when a Induction does not exist for a given Prisoner.
+ */
+class InductionNotFoundException(val prisonNumber: String) :
```
nathanrussell-moj-digital, line 6: I think this exception class might be in the wrong package? I think it should probably live in the domain source tree, in the same package as the `Induction` domain class ?
I think thats what we've done with our other domain classes and their exceptions?

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/entity/induction/InductionEntity.kt
```diff
@@ -0,0 +1,106 @@
...
+import java.time.Instant
+import java.util.UUID
+
+/**
+ * Represents a Prisoner's Induction, which is typically carried out by a CIAG officer shortly after the Prisoner
+ * has entered a Prison (either when starting a new sentence, or after being transferred from another Prison).
+ */
+@Table(name = "induction")
+@Entity
+@EntityListeners(AuditingEntityListener::class)
```
nathanrussell-moj-digital, line 31: To correctly audit the display names you'll need our entity listener as well as the standard one.
This line should be
```
@EntityListeners(value = [AuditingEntityListener::class, DisplayNameAuditingEntityListener::class])

```

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/entity/induction/FutureWorkInterestsEntity.kt
nathanrussell-moj-digital: Why do the other entities (and presumably domain classes) not have the prison and display name fields?
I have a funny feeling that we'll need them on some of the entities, because I think the screens show who last updated a given section of the CIAG Induction and in what prison.

I could be wrong, and I'm happy to go with it as is - we can always add them in later once we get into the detail of this stuff 👍  (Rome wasn't built in a day 😁 )

nathanrussell-moj-digital (commented):
I think all of your entity classes are fine as regular classes (not data classes) .... except you've not implemented equals/hashcode (and to a lesser importance, toString)

You are going to need an equals/hashcode contract that operates on the ID field alone. This is what we have in `GoalEntity` - I think you'll want something similar in each of your entity classes:
```
  override fun equals(other: Any?): Boolean {
    if (this === other) return true
    if (other == null || Hibernate.getClass(this) != Hibernate.getClass(other)) return false
    other as GoalEntity

    return id != null && id == other.id
  }

  override fun hashCode(): Int = javaClass.hashCode()

  override fun toString(): String {
    return this::class.simpleName + "(id = $id, reference = $reference, title = $title)"
  }
```

nathanrussell-moj-digital:
> Note that I've created an InductionRepository and a corresponding integration test locally to make sure I can save and retrieve an InductionEntity. I'll add test data builders in a subsequent PR, but I'll save the integration test until it's hooked up with the REST API.

Sounds like a sensible and perfectly reasonable approach 👍

mattwills-moj-digital:
> I think all of your entity classes are fine as regular classes (not data classes) .... except you've not implemented equals/hashcode (and to a lesser importance, toString)

Oops, I meant to add those. Will do shortly (and I'll remove the exception I accidentally copied as you suspected!).

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/entity/induction/FutureWorkInterestsEntity.kt
mattwills-moj-digital: The problem is that I have no idea what's required! Not now or in the future. I originally added all those fields to each entity, but it seemed overkill.

nathanrussell-moj-digital (dismissed):
LGTM 👍

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/TempRepositoryTest.kt
```diff
@@ -0,0 +1,265 @@
...
+@Deprecated("A temporary IT until we have the REST endpoint in place")
+internal class TempRepositoryTest : IntegrationTestBase() {
+
+  @Test
+  @Transactional
+  fun `should process CIAG induction created event`() {
+    // Given
+    val prisonNumber = aValidPrisonNumber()
+    val prisonId = "BXI"
+    // TODO - add test data builder
```
mattwills-moj-digital, line 45: Note the `@Deprecated` annotation and the TODOs in here. It's only temporary :)

nathanrussell-moj-digital (approved):
LGTM 👍

PR #122, opened 2023-10-19
RR-424 Create Induction DTOs, plus builders

mattwills-moj-digital:
DTOs for creating an Induction. Mappers etc to follow in separate PRs.

### DIFF
--- a/domain/induction/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/induction/dto/CreateInPrisonInterestsDto.kt
+++ b/domain/induction/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/induction/dto/CreateInPrisonInterestsDto.kt
@@ -1,0 +1,10 @@
+package uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.induction.dto
+
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.induction.InPrisonTrainingInterest
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.induction.InPrisonWorkInterest
+
+data class CreateInPrisonInterestsDto(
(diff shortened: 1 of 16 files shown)


domain/induction/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/induction/dto/CreateInPrisonInterestsDto.kt
```diff
@@ -0,0 +1,10 @@
+package uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.induction.dto
+
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.induction.InPrisonTrainingInterest
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.induction.InPrisonWorkInterest
+
+data class CreateInPrisonInterestsDto(
+  val inPrisonWorkInterests: List<InPrisonWorkInterest>,
```
nathanrussell-moj-digital, line 7: I was about to question about a DTO that has child members that are the domain objects (and slightly ridicule / compare you with someone else we know! 🤣 ) .... but these are enums? Yes? `InPrisonWorkInterest` is a domain enum, not a domain class?

nathanrussell-moj-digital (approved):
LGTM 👍

domain/induction/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/induction/dto/CreateInPrisonInterestsDto.kt
mattwills-moj-digital: A few of the classes are domain classes actually, but they are simple value objects with no identity, so I think it's ok.

PR #137, opened 2023-10-25
RR-429 - GET Induction Swagger spec

mattwills-moj-digital:
This PR adds the GET Induction endpoint to our swagger spec. To avoid breaking the CIAG UI, it's as close as possible to the existing version in the CIAG API: https://ciag-induction-api-dev.hmpps.service.justice.gov.uk/swagger-ui/index.html#/ciag-resource-controller/getCIAGProfileForOffenderId

The structure (e.g. names of fields) are identical, but I've pointed out any diversions away from the current spec in the comments.

Here's how the endpoint looks in our Swagger, alongside the recently added create Induction endpoint:
image

### DIFF
--- a/src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
+++ b/src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   title: Education and Work Plan API
-  version: '1.4.2'
+  version: '1.4.3'
   description: Education and Work Plan API
   contact:
     name: Matt Wills
@@ -169,7 +169,10 @@ paths:
         required: true
     post:
       summary: Creates an Induction for a Prisoner.
-      description: Creates an Induction for a Prisoner (if it does not already exist). This endpoint has been migrated from hmpps-ciag-careers-induction-api and the data structure has been kept the same in order to prevent breaking the UI when it is ports to this endpoint.
+      description: |
+        Creates an Induction for a Prisoner (if it does not already exist). This endpoint has been migrated from the
+        `hmpps-ciag-careers-induction-api` and the data structure has been kept the same in order to prevent breaking
+        the UI when it is ports to this endpoint.
       tags:
         - Induction
       responses:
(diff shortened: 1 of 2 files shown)


src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -454,6 +470,250 @@ components:
         - actionedBy
         - timestamp
         - correlationId
+    CiagInductionResponse:
+      title: CiagInductionResponse
+      type: object
+      description: A Prisoner's CIAG Induction containing information, such as their education history and future work interests.
+      properties:
+        reference:
```
mattwills-moj-digital, line 478: A new field which I'm assuming won't break the UI.

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -454,6 +470,250 @@ components:
...
+        - modifiedBy
+        - modifiedDateTime
+    PreviousWorkResponse:
+      title: PreviousWorkResponse
+      description: A Prisoner's previous work experience. Migrated from hmpps-ciag-careers-induction-api.
+      type: object
+      properties:
+        id:
+          type: string
+          format: uuid
```
mattwills-moj-digital, line 557: I've changed the format of the id from an integer to a UUID. I'm expecting it to break the CIAG UI (though perhaps not at runtime?), but this is something we can easily fix when we port the UI to call our API.

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -454,6 +470,250 @@ components:
         - actionedBy
         - timestamp
         - correlationId
+    CiagInductionResponse:
```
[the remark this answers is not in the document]
nathanrussell-moj-digital: > I've included Ciag in the name of both so that we can easily add our own versions in the future (though I appreciate I haven't done this throughout with the other related classes, so this idea is somewhat flawed!).

LOL - I was just going to ask about that, but you've answered it 😁 No problem with this approach, and I like your thinking about futures etc 👍

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
nathanrussell-moj-digital: Presumably this is in place of the current `id` field?; and presumably we'll need to get the CIAG UI to read this field (`reference`) rather than `id` and to use it in the URI path when making the subsequent `PUT` request ?

nathanrussell-moj-digital: So we'll need to remember that as part of our migration plans, as and when we switch the UI over to using this new API ?

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
nathanrussell-moj-digital: 👍 , agreed, as discussed yesterday 👍

nathanrussell-moj-digital (approved):
LGTM 👍

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
mattwills-moj-digital: It's a new field. The `offenderId` is the PK for the induction!

PR #140, opened 2023-10-27
RR-430 - Induction domain to CIAG API mappers

mattwills-moj-digital:
This PR adds the mappers to convert from our Induction domain model to the CIAG API model. Some of this was a bit fiddly since their models do not closely match in some places, so I have resorted to standard Kotlin classes, rather than MapStruct interfaces.


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/mapper/induction/CreateCiagInductionRequestMapper.kt
```diff
@@ -20,11 +19,11 @@ class CreateCiagInductionRequestMapper(
     return CreateInductionDto(
       prisonNumber = prisonNumber,
       workOnRelease = workOnReleaseMapper.toCreateWorkOnReleaseDto(request),
-      previousQualifications = qualificationsMapper.toCreatePreviousQualificationsDto(
+      previousQualifications = qualificationsAndTrainingMapper.toCreatePreviousQualificationsDto(
         request = request.qualificationsAndTraining,
         prisonId = prisonId,
       ),
-      previousTraining = previousTrainingMapper.toCreatePreviousTrainingDto(
+      previousTraining = qualificationsAndTrainingMapper.toCreatePreviousTrainingDto(
```
mattwills-moj-digital, line 26: It occurred to me that I was never going to be able to convert from two separate objects (`PreviousQualifications` and `PreviousTraining`) to one (`EducationAndQualificationResponse`) using two separate mappers, so I've combined `qualificationsMapper` and `previousTrainingMapper` into one `qualificationsAndTrainingMapper`.

nathanrussell-moj-digital (approved):
LGTM 👍

PR #143, opened 2023-10-27
RR-432 - Update Induction swagger spec

mattwills-moj-digital:
PR to add the swagger spec for the update Induction endpoint.

As part of this change, I've prefixed the existing create Induction related request objects with `Request...`, hence the large diff, but really only the swagger spec needs reviewing.

### DIFF
--- a/src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
+++ b/src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   title: Education and Work Plan API
-  version: '1.4.3'
+  version: '1.4.4'
   description: Education and Work Plan API
   contact:
     name: Matt Wills
@@ -196,6 +196,30 @@ paths:
         '404':
           $ref: '#/components/responses/404ErrorResponse'
       operationId: get-ciag-prisoner-induction
+    put:
+      summary: Updates an Induction for a Prisoner.
+      description: |
+        Provides the ability to update a Prisoner's Induction. Completely overwrites the existing Prisoner's Induction with the
+        information provided in the update request. If a field is not included in the update request, it will be removed from the 
+        database, effectively making the update request the authoritative source of information. This endpoint has been migrated
+        from the `hmpps-ciag-careers-induction-api` and the data structure has been kept the same in order to prevent breaking
+        the CIAG UI when it is ports to this endpoint.
+      tags:
+        - Induction
+      responses:
+        '200':
+          description: CIAG Induction updated successfully.
+          content:
+            application/json:
+              schema:
+                "$ref": "#/components/schemas/CiagInductionResponse"
+        '400':
+          $ref: '#/components/responses/400ErrorResponse'
+        '404':
+          $ref: '#/components/responses/404ErrorResponse'
+      operationId: update-ciag-prisoner-induction
+      requestBody:
+        $ref: '#/components/requestBodies/UpdateCiagInduction'
 
 #
 # --------------------------------------------------------------------------------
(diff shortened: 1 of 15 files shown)


src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   title: Education and Work Plan API
-  version: '1.4.3'
+  version: '1.4.4'
```
mattwills-moj-digital, line 4: 💥

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -1044,6 +1068,179 @@ components:
...
 
+    UpdateCiagInductionRequest:
+      title: UpdateCiagInductionRequest
+      description: A request to update a CIAG Induction. Migrated from hmpps-ciag-careers-induction-api.
+      type: object
+      properties:
+        reference:
+          type: string
+          format: uuid
+          description: The Induction's unique reference. Currently optional (to avoid breaking the existing CIAG API), but will be mandatory in future (and will most likely become a path variable).
```
mattwills-moj-digital, line 1079: As per the description here, I've made this optional for now. My plan is to overwrite every field in the DB, except this one if it's not provided in the request. That might not sound ideal, but it's hopefully only a temporary measure.

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -1044,6 +1068,179 @@ components:
...
+          $ref: '#/components/schemas/UpdatePrisonWorkAndEducationRequest'
+      required:
+        - hopingToGetWork
+        - prisonId
+
+    UpdatePreviousWorkRequest:
+      description: A request to persist the previous work experience of the Prisoner. Migrated from hmpps-ciag-careers-induction-api.
+      type: object
+      properties:
+        id:
```
mattwills-moj-digital, line 1120: This is the only difference from the existing "Create" request objects.

Thinking about it, I could consider using `allOf` here to reduce some of the duplication 🤔

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -196,6 +196,30 @@ paths:
...
+      description: |
+        Provides the ability to update a Prisoner's Induction. Completely overwrites the existing Prisoner's Induction with the
+        information provided in the update request. If a field is not included in the update request, it will be removed from the 
+        database, effectively making the update request the authoritative source of information. This endpoint has been migrated
+        from the `hmpps-ciag-careers-induction-api` and the data structure has been kept the same in order to prevent breaking
+        the CIAG UI when it is ports to this endpoint.
+      tags:
+        - Induction
+      responses:
+        '200':
```
nathanrussell-moj-digital, line 210: Do we know whether the CIAG UI reads the response back from the PUT operation?
If not we could do what we did with the create and just return a status code? In this case a 204?
(Unless you are foreseeing a circumstance where we might want to get the current updated resource back as part of the request? I'm just thinking if we could drop it it might make the implementation easier?)

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
nathanrussell-moj-digital: Yeah, sounds like a good idea 👍

nathanrussell-moj-digital (dismissed):

nathanrussell-moj-digital (approved):
LGTM 👍

PR #145, opened 2023-10-31
RR-431 - Corrections to swagger spec and update request builders

mattwills-moj-digital:
As the title suggests, this PR makes some corrections to the swagger spec and also adds the required test data builders for the new Update Induction API model objects. It also adds the required Update DTOs.

This is a precursor before I create a separate PR to add the mappers for converting the update model objects to the new Update DTOs.

### DIFF
--- a/src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
+++ b/src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   title: Education and Work Plan API
-  version: '1.4.4'
+  version: '1.4.5'
   description: Education and Work Plan API
   contact:
     name: Matt Wills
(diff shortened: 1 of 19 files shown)


src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -987,10 +987,33 @@ components:
           items:
             $ref: '#/components/schemas/WorkExperience'
         workInterests:
-          $ref: '#/components/schemas/WorkInterests'
+          $ref: '#/components/schemas/CreateWorkInterestsRequest'
       required:
         - hasWorkedBefore
 
+    CreateWorkInterestsRequest:
```
mattwills-moj-digital, line 994: We need a separate WorkInterests object for the request, due to the lack of an `id` (i.e. external reference).

Note that unlike the other 1st class citizen objects in the API, `WorkInterests` does not have `modifiedBy` or `modifiedDateTime` fields, which means that both the PUT (update) request object and the GET response object are identical. As a result, `WorkInterests` is used by both the PUT and GET endpoints, but the POST (create) endpoint needs a separate object due to the lack of an `id`.

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -1182,26 +1227,16 @@ components:
 
     WorkInterests:
```
mattwills-moj-digital, line 1228: This is the object used by both the PUT and GET endpoints.

nathanrussell-moj-digital (approved):
LGTM 👍

PR #149, opened 2023-11-02
RR-431 - Mappers for updating SkillsAndInterests and InPrisonInterests

mattwills-moj-digital:
This PR introduces the required `MapStruct` methods to update `PersonalSkillsAndInterests` and `InPrisonInterests`. Since both of the respective mappers use two internal "child" mappers, it made sense to do these at the same time. 

Unfortunately it's resulted in a lot of repetitive code and long test methods, which I'll see if I can do something about once I've finished the rest of the mapping.


nathanrussell-moj-digital (approved):
LGTM 👍

PR #152, opened 2023-11-03
RR-431 - Remaining mappers for updating an Induction

mattwills-moj-digital:


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/induction/InductionEntityMapper.kt
```diff
@@ -19,16 +30,79 @@ import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.induction.dto
...
+  @Mapping(target = "personalSkillsAndInterests", expression = "java( updatePersonalSkillsAndInterests(entity, dto) )")
+  @Mapping(target = "previousQualifications", expression = "java( updatePreviousQualifications(entity, dto) )")
+  @Mapping(target = "previousTraining", expression = "java( updatePreviousTraining(entity, dto) )")
+  @Mapping(target = "previousWorkExperiences", expression = "java( updatePreviousWorkExperiences(entity, dto) )")
+  @Mapping(target = "workOnRelease", expression = "java( updateWorkOnRelease(entity, dto) )")
+  abstract fun updateEntityFromDto(@MappingTarget entity: InductionEntity, dto: UpdateInductionDto)
+
+  fun updateFutureWorkInterests(entity: InductionEntity, dto: UpdateInductionDto): FutureWorkInterestsEntity? =
+    futureWorkInterestsEntityMapper.updateEntityFromDto(entity.futureWorkInterests, dto.futureWorkInterests)
+      .let { entity.futureWorkInterests }
```
mattwills-moj-digital, line 83: Due to the fact that the `updateEntityFromDto()` methods in the associated classes are void, I can't call them directly from the java expression statements within the `@Mapping` annotations above. For example, this doesn't compile:

```
  @Mapping(target = "workOnRelease", expression = "java( workOnReleaseEntityMapper.updateEntityFromDto(entity.workOnRelease, dto.workOnRelease) )")
```

I'll read up more on MapStruct as it's possible I've missed something...

src/test/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/induction/InductionEntityMapperTest.kt
```diff
@@ -157,4 +160,39 @@ class InductionEntityMapperTest {
...
+    val updateInductionDto = aValidUpdateInductionDto(
+      reference = inductionReference,
+      prisonNumber = prisonNumber,
+      prisonId = "MDI",
+    )
+    val expectedInductionEntity = existingInductionEntity.deepCopy().apply {
+      id
+      reference = inductionReference
+      createdAtPrison = "BXI"
+      updatedAtPrison = "MDI"
```
mattwills-moj-digital, line 182: There's actually very little to update in the Induction entity itself - everything else is contained within associated classes, which I've added separate mappers and unit tests for. The main intention of this test is to make sure the relevant mappers are called correctly (i.e. in the `verify()` methods below).

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/induction/InductionEntityMapper.kt
nathanrussell-moj-digital: It kinda makes sense though ... If you are trying to use a mapping annotation with a target attribute as you are, then the source or expression attribute needs to return a value doesnt it? Otherwise it cant map into the `target` field?

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/induction/PersonalSkillsAndInterestsEntityMapper.kt
```diff
@@ -50,8 +50,8 @@ abstract class PersonalSkillsAndInterestsEntityMapper {
   @Mapping(target = "skills", expression = "java( updateSkills(entity, dto) )")
   @Mapping(target = "interests", expression = "java( updateInterests(entity, dto) )")
   abstract fun updateEntityFromDto(
-    @MappingTarget entity: PersonalSkillsAndInterestsEntity,
-    dto: UpdatePersonalSkillsAndInterestsDto,
+    @MappingTarget entity: PersonalSkillsAndInterestsEntity?,
```
nathanrussell-moj-digital, line 53: Why have all the args for these various `updateEntity....` methods changed to nullable? 🤔

src/test/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/induction/InductionEntityMapperTest.kt
nathanrussell-moj-digital: 👍 
Given the intent (which may become lost/diluted after this PR?), any thoughts to naming the test something like
`should update induction by calling the correct child entity mappers` or similar?

src/test/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/induction/UpdatePreviousWorkExperiencesEntityMapperTest.kt
```diff
@@ -0,0 +1,187 @@
...
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.induction.aValidWorkExperienceEntity
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.induction.assertThat
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.deepCopy
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.induction.aValidWorkExperience
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.induction.dto.aValidUpdatePreviousWorkExperiencesDto
+import java.util.UUID
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.induction.WorkExperienceType as WorkExperienceTypeEntity
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.induction.WorkExperienceType as WorkExperienceTypeDomain
+
+class UpdatePreviousWorkExperiencesEntityMapperTest {
```
nathanrussell-moj-digital, line 14: 👍 
Agreed re: doing these as separate mapper tests, rather than trying to express all of this in the main induction mapper test 👍

src/test/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/induction/InductionEntityMapperTest.kt
mattwills-moj-digital: I'll change it to `should update induction and call related mappers`()

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/induction/PersonalSkillsAndInterestsEntityMapper.kt
mattwills-moj-digital: In some cases they're not mandatory - for example if a question is never asked, so the entity won't exist.

I think I mentioned on Slack that I need to go through the cardinality of each field in more detail and do more thorough testing of the permutations. I don't like merging code that I know is going to be wrong in places, but this mapping code is quite involved as it is and I'd rather get it deployed and then test the various permutations. For example, what happens if an induction has qualifications, but they are then omitted in a subsequent update request. Are the original qualifications removed? There's lots to test!

nathanrussell-moj-digital: 💯 👍 
OK, makes sense and happy with this approach, though I don't think the DTO arg would ever be null would it? Doesn't matter here and now though, if you are stepping back and looking at the cardinality of stuff separately then I'm sure it'll all get mopped up then 😁 👍

nathanrussell-moj-digital (approved):
LGTM 👍

PR #161, opened 2023-11-15
RR-493 - Fix mapping when initial object is null

mattwills-moj-digital:
This PR fixes an issue where the domain objects could not be updated if they were null originally. It also ensures that a reference is not mandatory in the incoming update request objects!

Some of the logic is quite complex, so I've added comments to try and explain it.

### DIFF
--- a/domain/induction/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/induction/Induction.kt
+++ b/domain/induction/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/induction/Induction.kt
@@ -28,10 +28,10 @@ data class Induction(
    */
   val previousQualifications: PreviousQualifications?,
   /**
-   * Any additional training that the Prisoner may have done previously. Null if the Prisoner has not been asked about
-   * their additional training.
+   * Any additional training that the Prisoner may have done previously. Mandatory as the Prisoner is always asked about
+   * it.
    */
-  val previousTraining: PreviousTraining?,
+  val previousTraining: PreviousTraining,
   /**
    * Details of any previous work experience that the Prisoner may have had. Null if the Prisoner has not been asked
    * about their work history.
--- a/domain/induction/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/induction/dto/UpdateFutureWorkInterestsDto.kt
+++ b/domain/induction/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/induction/dto/UpdateFutureWorkInterestsDto.kt
@@ -4,7 +4,7 @@ import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.induction.Wor
 import java.util.UUID
 
 data class UpdateFutureWorkInterestsDto(
-  val reference: UUID,
+  val reference: UUID?,
   val interests: List<WorkInterest>,
   val prisonId: String,
 )
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/induction/FutureWorkInterestsEntityMapper.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/induction/FutureWorkInterestsEntityMapper.kt
@@ -62,6 +62,24 @@ abstract class FutureWorkInterestsEntityMapper {
     dto: UpdateFutureWorkInterestsDto?,
   )
 
+  @ExcludeJpaManagedFieldsIncludingDisplayNameFields
+  @GenerateNewReference
+  @Mapping(target = "createdAtPrison", source = "prisonId")
+  @Mapping(target = "updatedAtPrison", source = "prisonId")
+  @Mapping(target = "interests", ignore = true)
+  abstract fun fromUpdateDtoToEntity(dto: UpdateFutureWorkInterestsDto?): FutureWorkInterestsEntity
+
+  @AfterMapping
+  fun addInterests(dto: UpdateFutureWorkInterestsDto, @MappingTarget entity: FutureWorkInterestsEntity) {
+    val existingTypes = entity.interests?.map { it.key() }
+    dto.interests.forEach {
+      // ensure we only add new interests
+      if (existingTypes == null || !existingTypes.contains(it.key())) {
+        entity.addChild(workInterestEntityMapper.fromDomainToEntity(it), entity.interests())
+      }
+    }
+  }
+
   fun updateInterests(entity: FutureWorkInterestsEntity, dto: UpdateFutureWorkInterestsDto): List<WorkInterestEntity> {
     val existingInterests = entity.interests!!
     val updatedInterests = dto.interests
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/induction/InductionEntityMapper.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/induction/InductionEntityMapper.kt
@@ -39,7 +39,7 @@ abstract class InductionEntityMapper {
   private lateinit var inPrisonInterestsEntityMapper: InPrisonInterestsEntityMapper
 
   @Autowired
-  private lateinit var personalSkillsAndInterestsEntityMapper: PersonalSkillsAndInterestsEntityMapper
+  private lateinit var skillsAndInterestsEntityMapper: PersonalSkillsAndInterestsEntityMapper
 
   @Autowired
   private lateinit var previousQualificationsEntityMapper: PreviousQualificationsEntityMapper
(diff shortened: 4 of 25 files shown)


domain/induction/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/induction/Induction.kt
```diff
@@ -28,10 +28,10 @@ data class Induction(
    */
   val previousQualifications: PreviousQualifications?,
   /**
-   * Any additional training that the Prisoner may have done previously. Null if the Prisoner has not been asked about
-   * their additional training.
+   * Any additional training that the Prisoner may have done previously. Mandatory as the Prisoner is always asked about
+   * it.
    */
-  val previousTraining: PreviousTraining?,
+  val previousTraining: PreviousTraining,
```
mattwills-moj-digital, line 34: On reflection, I originally made the domain too flexible. For the purposes of mapping, particularly when mapping two separate entities (e.g. eduction and training) into one response object, it's considerably easier if one of them is guaranteed to be populated!

domain/induction/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/induction/dto/UpdateFutureWorkInterestsDto.kt
```diff
@@ -4,7 +4,7 @@ import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.induction.Wor
 import java.util.UUID
 
 data class UpdateFutureWorkInterestsDto(
-  val reference: UUID,
+  val reference: UUID?,
```
mattwills-moj-digital, line 7: Oops, I should have foreseen this! If an object within the domain is null initially, then it will be impossible to provide a reference in the update request 🤦‍♂️

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/induction/FutureWorkInterestsEntityMapper.kt
```diff
@@ -62,6 +62,24 @@ abstract class FutureWorkInterestsEntityMapper {
...
+  @Mapping(target = "updatedAtPrison", source = "prisonId")
+  @Mapping(target = "interests", ignore = true)
+  abstract fun fromUpdateDtoToEntity(dto: UpdateFutureWorkInterestsDto?): FutureWorkInterestsEntity
+
+  @AfterMapping
+  fun addInterests(dto: UpdateFutureWorkInterestsDto, @MappingTarget entity: FutureWorkInterestsEntity) {
+    val existingTypes = entity.interests?.map { it.key() }
+    dto.interests.forEach {
+      // ensure we only add new interests
+      if (existingTypes == null || !existingTypes.contains(it.key())) {
```
mattwills-moj-digital, line 77: I was caught out here initially because this method is called when mapping from `UpdateFutureWorkInterestsDto` to `FutureWorkInterestsEntity`, regardless of whether it's after the new method `fromUpdateDtoToEntity()` immediately above here, or from the existing after `updateEntityFromDto()` one.

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/induction/InductionEntityMapper.kt
```diff
@@ -78,31 +78,61 @@ abstract class InductionEntityMapper {
...
-
-  fun updatePreviousTraining(entity: InductionEntity, dto: UpdateInductionDto): PreviousTrainingEntity? =
-    previousTrainingEntityMapper.updateEntityFromDto(entity.previousTraining, dto.previousTraining)
-      .let { entity.previousTraining }
-
-  fun updatePreviousWorkExperiences(entity: InductionEntity, dto: UpdateInductionDto): PreviousWorkExperiencesEntity? =
-    previousWorkExperiencesEntityMapper.updateEntityFromDto(entity.previousWorkExperiences, dto.previousWorkExperiences)
-      .let { entity.previousWorkExperiences }
+  fun updateFutureWorkInterests(entity: InductionEntity, dto: UpdateInductionDto): FutureWorkInterestsEntity? {
+    return if (entity.futureWorkInterests == null) {
```
mattwills-moj-digital, line 82: If the original entity is null, then we can't call the existing `updateEntityFromDto` methods, since they use the `@MappingTarget annotation`, which relies on a non null target object.

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/mapper/induction/QualificationsAndTrainingResourceMapper.kt
```diff
@@ -42,24 +52,30 @@ abstract class QualificationsAndTrainingResourceMapper {
...
-        additionalTraining = toTrainingTypesApi(training.trainingTypes),
-        additionalTrainingOther = if (training.trainingTypes.contains(OTHER)) training.trainingTypeOther else null,
-        modifiedBy = training.lastUpdatedBy!!,
-        modifiedDateTime = toOffsetDateTime(training.lastUpdatedAt)!!,
-      )
+    training: PreviousTraining,
+  ): EducationAndQualificationResponse {
+    // We always expect to have a value within training (even if it only contains TrainingType.NONE), so we use that
+    // as the "primary" object from which to populate the JPA generated fields from. However, if qualifications have
+    // been updated more recently, then we populate modifiedDateTime and modifiedBy from there instead.
```
mattwills-moj-digital, line 59: I think this is right because if I send in a request to update the list of qualifications, but not the training, then I would expect the modifiedDate of the `qualificationsAndTraining` object in the response to have an updated timestamp.

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/induction/InductionEntityMapper.kt
```diff
@@ -78,31 +78,61 @@ abstract class InductionEntityMapper {
...
+      workExperiencesEntityMapper.fromUpdateDtoToEntity(dto.previousWorkExperiences)
+    } else {
+      workExperiencesEntityMapper.updateEntityFromDto(entity.previousWorkExperiences, dto.previousWorkExperiences)
+        .let { entity.previousWorkExperiences }
+    }
+  }
 
   fun updateWorkOnRelease(entity: InductionEntity, dto: UpdateInductionDto): WorkOnReleaseEntity? =
-    workOnReleaseEntityMapper.updateEntityFromDto(entity.workOnRelease, dto.workOnRelease)
+    workOnReleaseEntityMapper.updateEntityFromDto(entity.workOnRelease!!, dto.workOnRelease)
```
mattwills-moj-digital, line 136: `workOnRelease` can never be null.

domain/induction/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/domain/induction/dto/UpdateFutureWorkInterestsDto.kt
nathanrussell-moj-digital: 👍 , yes, and thats also consistent with what we do in classes like `UpdateStepDto` for the same reason. Nice 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/induction/FutureWorkInterestsEntityMapper.kt
nathanrussell-moj-digital: 100% agreed. I wondered the other day when we were talking about it whether MapStruct was the right tool for the job in this case due to the complexity of mapping. But you are right - at least with it working and a great set of tests (specifically the integration tests) we can consider rewriting this if we feel it is the right thing to do 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/induction/InductionEntityMapper.kt
nathanrussell-moj-digital: Would changing the method signature of `workOnReleaseEntityMapper.updateEntityFromDto` be a better approach here? Or can it be null in _some_ cases, just not here?

nathanrussell-moj-digital (approved):
LGTM 👍 
Just one minor comment, but happy to approve, or re-approve if you choose to change it

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/induction/InductionEntityMapper.kt
mattwills-moj-digital: The entity fields are always nullable, which is something I'd like to investigate at some point.

PR #164, opened 2023-11-16
RR-493 - Missing "modified" fields and optional reference field

mattwills-moj-digital:
This PR fixes two issues that would have resulted in breaking changes when migrating from the CIAG API to our API:

- Adds "modified" fields to the `WorkInterestsResponse` to make it match the CIAG API's response.
- Makes an Induction's `reference` field optional in the update request.

Once this is merged, I will scrutinise the JSON between the two APIs and make sure they match :)

### DIFF
--- a/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/UpdateInductionTest.kt
+++ b/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/UpdateInductionTest.kt
@@ -43,8 +43,9 @@ import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.induc
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.induction.aValidUpdateEducationAndQualificationsRequest
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.induction.aValidUpdatePreviousWorkRequest
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.induction.aValidUpdateSkillsAndInterestsRequest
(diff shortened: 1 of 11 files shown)


src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/UpdateInductionTest.kt
```diff
@@ -854,4 +863,8 @@ class UpdateInductionTest : IntegrationTestBase() {
       qualificationsAndTraining = qualificationsAndTraining,
       inPrisonInterests = inPrisonInterests,
     )
+
+  private fun shortDelayForTimestampChecking() {
+    Thread.sleep(200)
```
nathanrussell-moj-digital, line 868: Did you have to bump it from 100 to 200? Due to intermittent failures?
If that is the case I would guess we might need to bump it back to the original 500 so that it always rounds up ?

nathanrussell-moj-digital (approved):
LGTM 👍

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/UpdateInductionTest.kt
mattwills-moj-digital: I played around with different figures and settled on 200. I don't recall encountering any errors, however there is a "fundamental flaw" concerning the `isAfter()` tests above, as the timestamp of the child entities will always be milliseconds after the main parent. I need to change this:

```
assertThat(updatedInduction.workExperience!!.modifiedDateTime).isAfter(initialModifiedAt)
```
To this:
```
assertThat(updatedInduction.workExperience!!.modifiedDateTime).isAfter(persistedInduction.workExperience!!.modifiedDateTime)
```
Or perhaps write an extension function 🤔 

However, I'd like to do this in a separate PR so that I can test that the APIs match ASAP.

PR #170, opened 2023-11-20
RR-503 migration entities and repository

mattwills-moj-digital:


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/ciagmigration/jpa/entity/induction/ParentMigrationEntity.kt
```diff
@@ -0,0 +1,21 @@
+package uk.gov.justice.digital.hmpps.educationandworkplanapi.app.ciagmigration.jpa.entity.induction
+
+/**
+ * Abstract class for "parent" entities, which contain one or more collections of [KeyAwareChildMigrationEntity] entities.
+ */
+abstract class ParentMigrationEntity {
+
```
mattwills-moj-digital, line 7: This is the same as the existing `ParentEntity`, except it doesn't have the method:

```
abstract fun childEntityUpdated()
```

This is because we are only going to use these entities for inserts, not updates, and we will use the timestamps that we import from the CIAG API.

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/ciagmigration/jpa/entity/induction/InductionMigrationEntity.kt
```diff
@@ -0,0 +1,110 @@
...
+import jakarta.persistence.Entity
+import jakarta.persistence.FetchType
+import jakarta.persistence.GeneratedValue
+import jakarta.persistence.Id
+import jakarta.persistence.JoinColumn
+import jakarta.persistence.OneToOne
+import jakarta.persistence.Table
+import jakarta.validation.constraints.NotNull
+import org.hibernate.Hibernate
+import org.hibernate.annotations.CreationTimestamp
```
nathanrussell-moj-digital, line 14: You appear to still be using `CreationTimestamp` and `UpdateTimestamp` - these need to go don't they?

mattwills-moj-digital: Oops, I was too quick 👍

nathanrussell-moj-digital (approved):
LGTM 👍

PR #171, opened 2023-11-20
RR-503 - Determine which inductions need to be imported

mattwills-moj-digital:


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/ciagmigration/CiagInductionMigrationService.kt
```diff
@@ -12,22 +12,34 @@ private val log = KotlinLogging.logger {}
...
-        .bodyToMono(CiagInductionResponse::class.java)
-        .block()
-
-      log.info { "Deserialised CIAG Induction: $ciagInduction" }
+      val prisonNumbers = timelineLookupService.getPrisonNumbersToMigrate()
+      log.info { "Identified ${prisonNumbers.size} Inductions to migrate" }
+
+      prisonNumbers.forEach {
+        val ciagInduction = getCiagInduction(it)
+        log.info { "Deserialized Induction for prisoner ${ciagInduction!!.offenderId}" }
```
mattwills-moj-digital, line 29: It will be interesting to see if this works. I suspect there's all kind of dodgy data in dev, so I'm half expecting this to fail. If it does then we need to find out why and what we next (e.g. delete the Induction(s) in question, or fix the ccde).

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/ciagmigration/CiagInductionMigrationService.kt
```diff
@@ -12,22 +12,34 @@ private val log = KotlinLogging.logger {}
...
-        .uri("/ciag/induction/A5077DY")
-        .retrieve()
-        .bodyToMono(CiagInductionResponse::class.java)
-        .block()
-
-      log.info { "Deserialised CIAG Induction: $ciagInduction" }
+      val prisonNumbers = timelineLookupService.getPrisonNumbersToMigrate()
+      log.info { "Identified ${prisonNumbers.size} Inductions to migrate" }
+
+      prisonNumbers.forEach {
```
mattwills-moj-digital, line 27: We're expecting to have a small number of Inductions to migrate, so this should be fine. Otherwise I'd probably have to call Sacintha's `/list` endpoint, but I've never tested it. Also, given it pulls back all the data for each Induction, I'm not convinced it's _that_ much more efficient anyway! 😄

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/ciagmigration/PrisonerInductionTimelineLookupService.kt
```diff
@@ -0,0 +1,23 @@
+package uk.gov.justice.digital.hmpps.educationandworkplanapi.app.ciagmigration
+
+import org.springframework.stereotype.Service
+import org.springframework.transaction.annotation.Transactional
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.timeline.TimelineEntity
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.timeline.TimelineEventType
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.repository.TimelineRepository
+
+@Service
+class PrisonerInductionTimelineLookupService(private val timelineRepository: TimelineRepository) {
```
mattwills-moj-digital, line 10: I struggled to name this, so suggestions welcomed, but hopefully this will suffice?

nathanrussell-moj-digital (approved):
Love it. As you say it will be interesting to see if this runs OK in `dev` and deserializes all of the inductions correctly (bearing in mind that the inductions processed will only be those that we have a timeline event for, so will be a smaller set than "all" inductions from day 1. I guess the likelihood of these latter inductions is that they'll all likely have "correct" data structures.)
Lets get it merged and see what gets logged in `dev` 🤞

PR #172, opened 2023-11-20
RR-503 - Builders for CIAG response and migration entities

mattwills-moj-digital:


nathanrussell-moj-digital (approved):
Looks good, but very minor nit-picking (😁 ), I think the packages should be
* `uk.gov.justice.digital.hmpps.educationandworkplanapi.app.ciagmigration.jpa.entity.induction` for the entities
* `uk.gov.justice.digital.hmpps.educationandworkplanapi.app.ciagmigration.resource.model` for the models

Up to you whether you want to change these, as I know its all temporary code 👍

mattwills-moj-digital:
Indeed I was aware the package structured differed slightly, but I wasn't overly concerned since it's just temporary code. The problem I had was that I didn't like empty packages, with nothing but a child package in them (e.g. `resource.model`), so I shortened the package path. Also, when it comes to the builders, we only ever build "data" objects, so the separation of things like `model` didn't seem so necessary here.

That said, I've removed `induction` from the entity package, as it was superfluous, and mirrored the packages in the builders. As a result, the diff is now a lot bigger!

nathanrussell-moj-digital (approved):
LGTM 👍

PR #176, opened 2023-11-21
RR-503 - Migration mappers

mattwills-moj-digital:
This PR adds the necessary mappers to convert the response objects we get from calling the CIAG API to our entities (or more accurately, a copy of our entities without the JPA annotations, for reasons I've explained previously).


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/ciagmigration/mapper/FutureWorkInterestsMigrationMapper.kt
```diff
@@ -0,0 +1,78 @@
...
+
+@Mapper(
+  uses = [
+    LocalDateTimeMapper::class,
+  ],
+  nullValueIterableMappingStrategy = NullValueMappingStrategy.RETURN_DEFAULT,
+)
+abstract class FutureWorkInterestsMigrationMapper {
+
+  fun toFutureWorkInterestsMigrationEntity(
```
mattwills-moj-digital, line 22: I've added this type of method to each mapper class because the generated one (which is now based on `createFutureWorkInterestsMigrationEntity()` below) had this null check at the beginning of each one:

```
if ( prisonId == null && workInterests == null ) {
   return null;
}
```
An or (`||`) would have been better, but on the basis `prisonId` is mandatory in the API, but workInterests isn't, I've added an if statement here to check only the latter.

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/ciagmigration/mapper/FutureWorkInterestsMigrationMapper.kt
```diff
@@ -0,0 +1,78 @@
...
+  ): FutureWorkInterestsMigrationEntity?
+
+  @AfterMapping
+  fun addExperiences(
+    workInterests: WorkInterestsResponse?,
+    @MappingTarget entity: FutureWorkInterestsMigrationEntity,
+  ) {
+    workInterests?.particularJobInterests?.forEach {
+      entity.addChild(
+        toWorkInterestMigrationEntity(it, workInterests.workInterestsOther, entity.updatedAt!!, entity.updatedBy!!),
```
mattwills-moj-digital, line 56: We don't have `modifiedDate` or `modifiedBy` within CIAG API response at this level, so I'm using the value from the parent entity (which won't be null because it would have been set in `createFutureWorkInterestsMigrationEntity()` above).

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/ciagmigration/mapper/InductionMigrationMapper.kt
```diff
@@ -0,0 +1,62 @@
...
+  private val workExperiencesMapper: PreviousWorkExperiencesMigrationMapper,
+  private val inPrisonInterestsMapper: InPrisonInterestsMigrationMapper,
+  private val skillsAndInterestsMapper: SkillsAndInterestsMigrationMapper,
+  private val workInterestsMapper: FutureWorkInterestsMigrationMapper,
+) {
+  fun toInductionMigrationEntity(ciagInduction: CiagInductionResponse): InductionMigrationEntity =
+    with(ciagInduction) {
+      // prisonId shouldn't be null, but we've seen bugs here on the CIAG side recently.
+      // We can default it for now, so long as the service is still only rolled out to Lancaster Farms!
+      val prisonId = ciagInduction.prisonId ?: "LFI"
```
mattwills-moj-digital, line 23: I'm defaulting this for now, but I might need to change this. As I've said in the comments, it shouldn't be null anyway, but it's not mandatory in their API spec.

src/test/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/ciagmigration/mapper/InductionMigrationMapperTest.kt
```diff
@@ -0,0 +1,375 @@
...
+    trainingMapper = trainingMapper,
+    workExperiencesMapper = workExperiencesMapper,
+    inPrisonInterestsMapper = inPrisonInterestsMapper,
+    skillsAndInterestsMapper = skillsAndInterestsMapper,
+    workInterestsMapper = workInterestsMapper,
+  )
+
+  @Test
+  fun `should map to InductionMigrationEntity`() {
+    // Given
```
mattwills-moj-digital, line 116: This is the largest `Given` section I've ever written and I can't say I'm proud of it! 😂 

I could make the builders use the same values, which would massively reduce the amount of code, but it wouldn't be as explicit 🤔

src/test/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/ciagmigration/mapper/InductionMigrationMapperTest.kt
```diff
@@ -0,0 +1,375 @@
...
+      updatedBy = "bjones_gen",
+      updatedByDisplayName = "bjones_gen",
+    )
+
+    // When
+    val actual = inductionMapper.toInductionMigrationEntity(ciagInduction)
+
+    // Then
+    assertThat(actual).usingRecursiveComparison()
+      .ignoringFieldsMatchingRegexes(".*id", ".*reference", ".*parent")
```
mattwills-moj-digital, line 372: Crucially this ensures that all the dates are as we expect.

src/test/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/ciagmigration/mapper/InductionMigrationMapperTest.kt
```diff
@@ -0,0 +1,375 @@
...
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.ciagmigration.jpa.entity.HighestEducationLevel as HighestEducationLevelEntity
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.ciagmigration.jpa.entity.HopingToWork as HopingToWorkEntity
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.ciagmigration.jpa.entity.InPrisonTrainingType as InPrisonTrainingTypeEntity
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.ciagmigration.jpa.entity.InPrisonWorkType as InPrisonWorkTypeEntity
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.ciagmigration.jpa.entity.TrainingType as TrainingTypeEntity
+
+class InductionMigrationMapperTest {
+
+  private val localDateTimeMapper = LocalDateTimeMapper()
+  private val workOnReleaseMapper = WorkOnReleaseMigrationMapperImpl().also {
```
mattwills-moj-digital, line 61: Rather than creating a separate unit test for each mapper and then mocking them in this one, I think it's better to test them fully from the top level one. The downside is the injection of the `localDateTimeMapper` in all the mappers here, plus the massive test method below!

However, using mocks means you get to dictate what each mapper returns, which means it's not tested "end to end" (except at the integration test level, which doesn't exist yet).

src/testFixtures/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/ciagmigration/jpa/entity/FutureWorkInterestsMigrationEntityBuilder.kt
```diff
@@ -35,10 +35,18 @@ fun aValidWorkInterestMigrationEntity(
   workType: WorkInterestType = WorkInterestType.OTHER,
   workTypeOther: String? = "Any job I can get",
   role: String? = "Any role",
+  createdAt: Instant? = Instant.now(),
+  createdBy: String? = "asmith_gen",
+  updatedAt: Instant? = Instant.now(),
+  updatedBy: String? = "bjones_gen",
```
mattwills-moj-digital, line 41: Missing fields within the builders 🙄

nathanrussell-moj-digital (approved):
LGTM

PR #177, opened 2023-11-22
RR-508 - Change paths to match that of current implementation in CIAG API

nathanrussell-moj-digital:
We know these paths are not ideal, but having the paths the same as the current CIAG API will make migrating the clients much much easier.
We will look to change the paths (and the request and response objects) when we re-design the API once we have migrated the UI clients


mattwills-moj-digital (approved):
🤮

PR #184, opened 2023-11-24
RR-511 - Allow empty collections in create requests

mattwills-moj-digital:
Technically the API was accepting an empty/null collection of qualifications when creating an induction, but this was then being mapped to a null collection in the domain object afterwards.

I had tests for sending null/empty lists of qualifications in update requests, but not create ones (though I had a test for sending no qualifications data, including no `educationLevel`).


nathanrussell-moj-digital (approved):
LGTM 👍

PR #186, opened 2023-11-27
RR-512 - Save multiple GOAL_CREATED events with the same correlation ID

nathanrussell-moj-digital:
This PR fixes a problem where when saving multiple `GOAL_CREATED` events in the same request , they need to be saved with the same correlation ID
Previously they were being saved with a new correlation ID for each `Goal`. This has been changed so that irrespective of how many `Goal`s were created as part of the originating request, they all use the same correlation ID.


mattwills-moj-digital (approved):
LGTM 👍

PR #189, opened 2023-11-28
RR-515 - Add hasWorkedBefore field/column

mattwills-moj-digital:
When updating a Prisoner's previous work history, the CIAG UI only asks if they have worked before - it does not capture any details of any relevant work experience if they selected 'Yes'. This is different to the create journey, which captures a mandatory list of one or more work experiences (i.e. if they selected 'Yes').

This PR ensures that `hasWorkedBefore` is always set to the value the UI sends, rather than being derived from the presence of data in the `experiences` collection.

This PR also ensures that we save the value that we migrate from the CIAG API to this API, so that we will always have an initial value (unless the whole of the `previousWorkExperiences` object is null because they were not asked the question).


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/ciagmigration/jpa/entity/PreviousWorkExperiencesMigrationEntity.kt
```diff
@@ -36,6 +36,9 @@ class PreviousWorkExperiencesMigrationEntity(
   @field:NotNull
   var reference: UUID? = null,
 
+  @field:NotNull
```
mattwills-moj-digital, line 39: This is not null in the "Migration" entity, but not in "our" entity - at least not yet. We may choose to make it mandatory after the data migration from the CIAG API.

src/test/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/mapper/induction/FutureWorkInterestsResourceMapperTest.kt
```diff
@@ -128,7 +128,7 @@ class FutureWorkInterestsResourceMapperTest {
   }
 
   @Test
-  fun `should map to UpdateFutureWorkInterestsDto when particular job interests is null`() {
+  fun `should map to UpdateFutureWorkInterestsDto when particular job interests is empty`() {
```
mattwills-moj-digital, line 131: This is a hangover from my previous PR.

nathanrussell-moj-digital (approved):
LGTM 👍

PR #190, opened 2023-11-28
Enable migration and change replicaCount to 1 in all envs

mattwills-moj-digital:
PR to enable the CIAG data migration in all envs and also temporarily set the replica count to 1.


nathanrussell-moj-digital (approved):
LGTM 👍

PR #196, opened 2023-12-01
RR-502 Remove CIAG Induction migration functionality

mattwills-moj-digital:
Now that we've migrated all the Inductions from the CIAG API, we can remove all of the migration related functionality. In doing so, this will speed up the build and reduce the API's resources (fewer entities, repositories and Spring beans in general).


nathanrussell-moj-digital (approved):
LGTM 👍

PR #200, opened 2023-12-05
RR-476 - Correctly set actionedBy fields when saving GOAL_UPDATED timeline event

nathanrussell-moj-digital:
This PR fixes a problem with GOAL_UPDATED timeline events where the actionedBy fields were not being correctly set to the user that performed the goal update.

The problem was because the the goal was not saved & flushed within the Goal JPA Adapter , which meant the Goal that was returned from the method was the updated goal in respect of the fields mapped from the DTO, but crucially did not have the JPA managed fields updated on it.

---
I'm in 2 minds about whether to write a flyway migration to correct any current timeline event records that have the wrong actioned_by fields. Currently, using this query on `prod`, there are no affected records, which is good:
```
select te.* from timeline_event te 
inner join goal g on g.reference::text = te.source_reference
where te.event_type = 'GOAL_UPDATED'
and te.actioned_by != g.updated_by;
```
But I wonder if we should have the flyway migration as part of this PR anyway, just for completeness?
Other envs such as `dev` will have incorrect records (which is how we found this bug to start with), but I wonder if we care too much about `dev`? 
I've just checked with the above query in `dev`, and there are 15 affected records

### DIFF
--- a/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/IntegrationTestBase.kt
+++ b/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/IntegrationTestBase.kt
@@ -27,8 +27,10 @@ import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.Actio
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.CiagInductionResponse
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.CreateActionPlanRequest
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.CreateCiagInductionRequest
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.CreateGoalRequest
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.TimelineResponse
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.actionplan.aValidCreateActionPlanRequest
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.actionplan.aValidCreateGoalsRequest
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.withBody
 import uk.gov.justice.hmpps.sqs.HmppsQueueService
 import uk.gov.justice.hmpps.sqs.MissingQueueException
@@ -153,4 +155,27 @@ abstract class IntegrationTestBase {
       .isOk
       .returnResult(CiagInductionResponse::class.java)
       .responseBody.blockFirst()!!
+
+  fun createGoal(
+    prisonNumber: String,
+    createGoalRequest: CreateGoalRequest,
+    username: String = "auser_gen",
+    displayName: String = "Albert User",
+  ) {
+    val createGoalsRequest = aValidCreateGoalsRequest(goals = listOf(createGoalRequest))
+    webTestClient.post()
+      .uri("/action-plans/{prisonNumber}/goals", prisonNumber)
+      .withBody(createGoalsRequest)
+      .bearerToken(
+        aValidTokenWithEditAuthority(
+          username = username,
+          displayName = displayName,
+          privateKey = keyPair.private,
+        ),
+      )
+      .contentType(MediaType.APPLICATION_JSON)
+      .exchange()
+      .expectStatus()
+      .isCreated()
+  }
 }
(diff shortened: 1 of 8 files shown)


src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/IntegrationTestBase.kt
```diff
@@ -153,4 +155,27 @@ abstract class IntegrationTestBase {
       .isOk
       .returnResult(CiagInductionResponse::class.java)
       .responseBody.blockFirst()!!
+
+  fun createGoal(
```
nathanrussell-moj-digital, line 159: I moved this method here into the base class `IntegrationTestBase` from `UpdateGoalTest`, and also added method args for the username that the JWT token should be created with (and therefore that is the user that is recorded as the createdBy etc fields on the created Goal)

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/UpdateGoalTest.kt
```diff
@@ -183,7 +183,12 @@ class UpdateGoalTest : IntegrationTestBase() {
         ),
       ),
     )
-    createGoal(prisonNumber, createGoalRequest)
+    createGoal(
```
nathanrussell-moj-digital, line 186: The initial goal (that we are about to edit) is created by `auser_gen`

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/UpdateGoalTest.kt
```diff
@@ -212,7 +217,13 @@ class UpdateGoalTest : IntegrationTestBase() {
     webTestClient.put()
       .uri(URI_TEMPLATE, prisonNumber, goalReference)
       .withBody(updateGoalRequest)
-      .bearerToken(aValidTokenWithEditAuthority(privateKey = keyPair.private))
+      .bearerToken(
```
nathanrussell-moj-digital, line 220: Here where we edit the goal we use a JWT token for the user `buser_gen`

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/UpdateGoalTest.kt
```diff
@@ -238,6 +249,16 @@ class UpdateGoalTest : IntegrationTestBase() {
           }
           .wasCreatedAtPrison("BXI")
           .wasUpdatedAtPrison("MDI")
+          .wasCreatedBy("auser_gen")
```
nathanrussell-moj-digital, line 252: Assert the GoalEntity has the correct createdBy and updatedBy fields (which they would have done before, this was not the bug)

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/UpdateGoalTest.kt
```diff
@@ -266,19 +287,4 @@ class UpdateGoalTest : IntegrationTestBase() {
         .isEqualTo(stepRemovedEventProperties["correlationId"])
     }
   }
-
-  private fun createGoal(
```
nathanrussell-moj-digital, line 270: This method moved to the superclass

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/JpaGoalPersistenceAdapter.kt
```diff
@@ -53,7 +55,8 @@ class JpaGoalPersistenceAdapter(
     val goalEntity = getGoalEntityByPrisonNumberAndGoalReference(prisonNumber, updatedGoalDto.reference)
     return if (goalEntity != null) {
       goalMapper.updateEntityFromDto(goalEntity, updatedGoalDto)
-      goalMapper.fromEntityToDomain(goalEntity)
+      val persistedEntity = goalRepository.saveAndFlush(goalEntity)
```
nathanrussell-moj-digital, line 58: Previously, without the saveAndFlush, the goal entity was updated and the JPA fields were correctly being set and updated .... but the flush (and therefore the JPA managed fields set) was being done at the transaction boundary. This meant that the domain goal that was returned (via the mapper) did not have the correct updatedBy fields because they had not been set by JPA at that point

nathanrussell-moj-digital: OK, will do 👍

nathanrussell-moj-digital: OK, done

mattwills-moj-digital (dismissed):
LGTM. Thanks for making the additional change 👍

mattwills-moj-digital (approved):

PR #201, opened 2023-12-05
RR-519 Swagger spec - new create Induction endpoint

mattwills-moj-digital:
This PR introduces the swagger spec for a new endpoint to create Inductions. To keep it consistent with our other endpoints, this new endpoint will be available under `/inductions` (plural).

As part of this change, I've also deprecated the endpoints we migrated from the CIAG API and have tagged them as "Legacy":

image

### DIFF
--- a/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/UpdateInductionTest.kt
+++ b/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/UpdateInductionTest.kt
@@ -25,8 +25,8 @@ import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.Highe
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.HopingToWork
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.InPrisonTrainingType
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.InPrisonWorkType
-import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.PersonalInterest
-import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.PersonalSkill
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.PersonalInterestType
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.PersonalSkillType
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.ReasonNotToWork
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.TimelineEventType
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.TrainingType
@@ -183,9 +183,9 @@ class UpdateInductionTest : IntegrationTestBase() {
       ),
       skillsAndInterests = aValidUpdateSkillsAndInterestsRequest(
         id = persistedInduction.skillsAndInterests!!.id!!,
-        skills = setOf(PersonalSkill.COMMUNICATION),
+        skills = setOf(PersonalSkillType.COMMUNICATION),
         skillsOther = null,
-        personalInterests = setOf(PersonalInterest.CRAFTS),
+        personalInterests = setOf(PersonalInterestType.CRAFTS),
         personalInterestsOther = null,
       ),
       qualificationsAndTraining = aValidUpdateEducationAndQualificationsRequest(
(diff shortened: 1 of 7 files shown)


src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -26,7 +26,7 @@ paths:
       summary: Retrieves a list of Action Plans for a given list of Prisoners.
       description: Retrieves a list of Action Plans for a given list of Prisoners. This is a POST, rather than a GET, to allow a long list of prison numbers.
       tags:
-        - Action Plan
+        - Action Plans
```
mattwills-moj-digital, line 29: There was an inconsistency with singular and plural tag names. I've made them all plural to match the paths.

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -767,6 +802,59 @@ components:
         - id
         - modifiedBy
         - modifiedDateTime
+    CiagInductionSummaryListResponse:
```
mattwills-moj-digital, line 805: I've moved this from below to put it next to the other response objects.

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/UpdateInductionTest.kt
```diff
@@ -176,9 +176,9 @@ class UpdateInductionTest : IntegrationTestBase() {
       ),
       skillsAndInterests = aValidUpdateSkillsAndInterestsRequest(
         id = persistedInduction.skillsAndInterests!!.id!!,
-        skills = setOf(PersonalSkill.COMMUNICATION),
+        skills = setOf(PersonalSkillType.COMMUNICATION),
```
mattwills-moj-digital, line 179: I've renamed the enum `PersonalSkill` to `PersonalSkillType` to allow for a new `PersonalSkill` object.

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -1463,60 +1614,141 @@ components:
...
+        personalSkillsAndInterests:
+          $ref: '#/components/schemas/CreatePersonalSkillsAndInterestsRequest'
+        futureWorkInterests:
+          $ref: '#/components/schemas/CreateFutureWorkInterestsRequest'
       required:
-        - ciagProfileList
-
-    CiagInductionSummaryResponse:
-      title: CiagInductionSummaryResponse
+        - workOnRelease
```
mattwills-moj-digital, line 1637: I'm making this the only mandatory field. I've considered making `previousTraining` mandatory too, as I think that's always asked (and `NONE` is a viable option), but I don't want to make decisions based on the current screens. 

Stepping back from the screen flow, it makes sense that a prisoner is always asked if they hope to work  - it feels like a key part of an induction - but gathering other apsects, such as their previous training, does not seem vital.

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -1463,60 +1614,141 @@ components:
...
-        - createdDateTime
-        - modifiedBy
-        - modifiedDateTime
+        - trainingTypes
+    CreatePreviousWorkExperiencesRequest:
+      title: CreatePreviousWorkExperiencesRequest
+      type: object
+      description: A request to create a Prisoner's previous work experience.
+      properties:
+        hasWorkedBefore:
```
mattwills-moj-digital, line 1695: It could be argued this is redundant based on the `experiences` array, but I think it's useful for convenience. Also, we know that the current screens do not send a list of work experiences when updating an Induction (because Gary didn't have clear requirements 😉), which is another (albeit lesser) reason to keep it.

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -168,13 +168,14 @@ paths:
         in: path
         required: true
     post:
+      deprecated: true
```
mattwills-moj-digital, line 171: This makes it pretty clear in swagger editor that the endpoint is deprecated!

nathanrussell-moj-digital:
The swagger spec is difficult to review (because of the size of it and therefore the diff is hard to read), so its really difficult to see/follow the proposed spec.
I've taken the raw file and loaded it into swagged editor, which makes it easier 👍 , but I have questions 😁 
Perhaps easiest to talk this through on a call?

src/testFixtures/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/resource/model/induction/EducationAndQualificationsRequestBuilder.kt
```diff
@@ -40,19 +40,19 @@ fun aValidUpdateEducationAndQualificationsRequest(
 )
 
 fun aValidAchievedQualification(
-  subject: String? = "English",
-  level: AchievedQualification.Level? = AchievedQualification.Level.LEVEL_3,
-  grade: String? = "A",
+  subject: String = "English",
```
mattwills-moj-digital, line 43: The cardinality in the API model now matches the domain/entity models.

nathanrussell-moj-digital (approved):
I still cant read this diff properly to understand the scope of the change (and specifically what types, if any, are shared with the current CIAG Induction endpoints; and what the cardinality is)
All I can do is trust what you've done here, and I'm sure that over subsequent PRs for both the spec and the implementation it will become a little clearer.
And of course, we can always change it if we find we need to 👍

PR #202, opened 2023-12-05
RR-520 Swagger spec for retrieving Inductions

mattwills-moj-digital:
This PR specifies a new endpoint to retrieve Inductions.

image

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/mapper/induction/QualificationsAndTrainingResourceMapper.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/mapper/induction/QualificationsAndTrainingResourceMapper.kt
@@ -65,7 +65,7 @@ abstract class QualificationsAndTrainingResourceMapper {
     }
     return EducationAndQualificationResponse(
       id = training.reference,
-      educationLevel = toHighestEducationLevelApi(qualifications?.educationLevel),
+      educationLevel = toHighestEducationLevelApi(qualifications?.educationLevel) ?: HighestEducationLevelApi.NOT_SURE,
       qualifications = toAchievedQualifications(qualifications?.qualifications),
       additionalTraining = toTrainingTypesApi(training.trainingTypes),
       additionalTrainingOther = if (training.trainingTypes.contains(OTHER)) training.trainingTypeOther else null,
--- a/src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
+++ b/src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   title: Education and Work Plan API
-  version: '1.7.0'
+  version: '1.7.1'
   description: Education and Work Plan API
   contact:
     name: Matt Wills
@@ -1090,6 +1090,7 @@ components:
           example: 'BXI'
       required:
         - reference
+        - educationLevel
         - qualifications
         - createdBy
         - createdByDisplayName
@@ -2089,7 +2090,7 @@ components:
     HighestEducationLevel:
       title: HighestEducationLevel
       type: string
-      description: The highest education level of a Prisoner.
+      description: The highest education level of a Prisoner. Defaults to 'NOT_SURE' if no value has been provided.
       enum:
         - PRIMARY_SCHOOL
         - SECONDARY_SCHOOL_LEFT_BEFORE_TAKING_EXAMS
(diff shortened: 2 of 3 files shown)


src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -856,6 +886,564 @@ components:
...
+          description: An ISO-8601 timestamp representing when the Induction was last updated. This will be the same as the created date if it has not yet been updated.
+          example: '2023-06-19T09:39:44Z'
+        updatedAtPrison:
+          type: string
+          description: The identifier of the prison that the prisoner was resident at when the Induction was updated.
+          example: 'BXI'
+      required:
+        - reference
+        - prisonNumber
+        - workOnRelease
```
mattwills-moj-digital, line 955: Like with the create request, this is the only mandatory "child" object at this level.

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -856,6 +886,564 @@ components:
...
+          description: An ISO-8601 timestamp representing when this resource was last updated. This will be the same as the created date if it has not yet been updated.
+          example: '2023-06-19T09:39:44Z'
+        updatedAtPrison:
+          type: string
+          description: The identifier of the prison that the prisoner was resident at when this resource was updated.
+          example: 'BXI'
+      required:
+        - reference
+        - hopingToWork
+        - notHopingToWorkReasons
```
mattwills-moj-digital, line 1030: You'll note in this PR that all the collections are mandatory, but I've explained in the description of each one that they "can be empty but not null".

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/mapper/induction/QualificationsAndTrainingResourceMapper.kt
```diff
@@ -65,7 +65,7 @@ abstract class QualificationsAndTrainingResourceMapper {
     }
     return EducationAndQualificationResponse(
       id = training.reference,
-      educationLevel = toHighestEducationLevelApi(qualifications?.educationLevel),
+      educationLevel = toHighestEducationLevelApi(qualifications?.educationLevel) ?: HighestEducationLevelApi.NOT_SURE,
```
mattwills-moj-digital, line 68: In addition to defaulting this in the request (for future create/update requests), I'm now defaulting this in the response for existing entries. I didn't rush this through last night, but I'm confident the UI handles it (for both journeys) so I'm including it now.

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -856,6 +886,565 @@ components:
...
+        - createdBy
+        - createdByDisplayName
+        - createdAt
+        - createdAtPrison
+        - updatedBy
+        - updatedByDisplayName
+        - updatedAt
+        - updatedAtPrison
+
+    InductionSummaryListResponse:
```
mattwills-moj-digital, line 1422: These two objects (`InductionSummaryListResponse` and `InductionSummaryResponse`) are identical to the existing CIAG objects (which are prefixed with `Ciag`). I don't see any reason to change them yet, but having separate objects allows us to change them independently.

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -267,6 +267,36 @@ paths:
...
+      tags:
+        - Inductions
+      responses:
+        '200':
+          $ref: '#/components/responses/Induction'
+        '404':
+          $ref: '#/components/responses/404ErrorResponse'
+      operationId: get-prisoner-induction
+
+  '/inductions':
```
nathanrussell-moj-digital, line 283: I'm not sure about this (the POST endpoint to get the induction summaries)
I wonder if we should whip it out of this PR and think about it separately. 

FWIW, my current thought process is that just because its in the current CIAG API, it doesn't mean we need to migrate it, and certainly not exactly as is. The only reason it's there is so that the Prisoner List screen - which is now _our_ Prisoner List screen in _our_ UI - knows if there is an induction. And it also calls the action plan summary endpoint for the same reason (to know if there are any Goals).
Those 2 summary endpoints were implemented specifically for the Prisoner List page so that it could get data from 2 sources to workout the status and routing.
It's no longer 2 sources .... so why do we need 2 API calls? 🤔 

I wonder if we'd be better off not implementing a copy of the summary API, and instead stepping back and thinking about what the UI needs and why, and designing a single API that addresses that ?
(in a separate story of course 👍 )

What do you think?

mattwills-moj-digital: That makes complete sense. I'll whip it out (as you put it!).

PR #203, opened 2023-12-05
RR-521 - Update Induction swagger spec

mattwills-moj-digital:
This PR adds the swagger spec for a new update Induction endpoint.

The following sceenshot from swagger-editor makes it look like the `UpdateInductionRequest` object contains the "create" objects by mistake, but that's because the "update" objects are based on the respective create ones, using the `allOf` notation. They are the same as the create objects, except for the `reference` field.

image

### DIFF
--- a/src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
+++ b/src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
@@ -2348,7 +2348,6 @@ components:
           $ref: '#/components/schemas/UpdateFutureWorkInterestsRequest'
       required:
         - reference
-        - workOnRelease
     UpdateWorkOnReleaseRequest:
       description: A request to update a Prisoner's views on obtaining work after release.
       allOf:


src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -2353,6 +2371,106 @@ components:
...
+          $ref: '#/components/schemas/UpdatePersonalSkillsAndInterestsRequest'
+        futureWorkInterests:
+          $ref: '#/components/schemas/UpdateFutureWorkInterestsRequest'
+      required:
+        - reference
+        - workOnRelease
+    UpdateWorkOnReleaseRequest:
+      description: A request to update a Prisoner's views on obtaining work after release.
+      allOf:
+        - $ref: '#/components/schemas/CreateWorkOnReleaseRequest'
```
mattwills-moj-digital, line 2404: All of these "update" request objects are the same as the create ones, except for the addition of a `reference` field. This field is optional in all cases, except for this object, since there must always be a `WorkOnRelease` resource.

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -2308,6 +2326,106 @@ components:
       required:
         - interests
 
+    UpdateInductionRequest:
+      title: UpdateInductionRequest
+      description: A request to update a Prisoner's Induction.
+      type: object
+      properties:
+        reference:
```
nathanrussell-moj-digital, line 2334: Interesting that you added `reference` to this top level request object ... did you have any thoughts about passing it as a path parameter to perhaps make the API a little more RESTful in this respect? 🤔

mattwills-moj-digital: As discussed, I've followed the same approach as we have for `/action-plans` and `/timelines`.

Perhaps the question is not so much about whether we should have `reference` in the path, but more of whether these objects should have a reference at all? Something to think about but I'm not going to change that now :)

nathanrussell-moj-digital (approved):
Lovely 😁 👍

PR #206, opened 2023-12-06
RR-446 - Temporarily disable cronjob

nathanrussell-moj-digital:
Temporarily disable the "export data to analytics platform" cronjob (with `suspend: true`) because it is likely that we will need to promote `main` to `preprod` and `prod` soon, but the analytical platform setup work is not yet complete in `preprod` and `prod` which means we can't allow the cronjob to run in those environments yet.

### DIFF
--- a/helm_deploy/hmpps-education-and-work-plan-api/templates/export-database-to-analytical-platform-cronjob.yaml
+++ b/helm_deploy/hmpps-education-and-work-plan-api/templates/export-database-to-analytical-platform-cronjob.yaml
@@ -8,7 +8,8 @@ kind: CronJob
 metadata:
   name: export-database-to-analytical-platform-cronjob
 spec:
-  schedule: "0 9-17 * * 1-5" # On the hour between 09:00 and 17:00 Mon-Fri. For testing only; once smoke tested and working change to every day at 1am -> "0 1 * * *"
+  suspend: true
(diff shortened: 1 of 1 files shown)


helm_deploy/hmpps-education-and-work-plan-api/templates/export-database-to-analytical-platform-cronjob.yaml
```diff
@@ -8,7 +8,8 @@ kind: CronJob
 metadata:
   name: export-database-to-analytical-platform-cronjob
 spec:
-  schedule: "0 9-17 * * 1-5" # On the hour between 09:00 and 17:00 Mon-Fri. For testing only; once smoke tested and working change to every day at 1am -> "0 1 * * *"
+  suspend: true
+  schedule: "0 1 * * *" # 1am every day
```
nathanrussell-moj-digital, line 12: Now that we have tested this in `dev` and seen the data copy across every hour, we can change the cron to once a day.
The real change in this PR however is the `suspend: true` property, which means the cronjob wont run at all. This means that if we have to promote `main` to `preprod` and `prod` soon (which I think we will to deliver other features and bugfixes that we've been working on recently) then there is no risk of this running in `preprod` or `prod`

mattwills-moj-digital (approved):

PR #207, opened 2023-12-06
RR-533 - Fix mapping of qualifications when highest level of education is not provided

mattwills-moj-digital:
This fixes a bug we've discovered where qualifications are not saved during the short route.


nathanrussell-moj-digital (approved):
👍

PR #208, opened 2023-12-11
RR-476 - Update Goal audit fields when updating Steps

nathanrussell-moj-digital:
This PR fixes a bug when an `UpdateGoalRequest` was processed that contained new or edited Steps, but not removed Steps, and no changes to the parent Goal. IE. a request to update a Goal where the request was to either add new Steps only, or update existing Steps only.

In these cases the parent Goal was not updated at all, so the JPA managed audit fields were not updated. This had the side effect of the Timeline event for the updated Goal having the wrong "updated by/at" fields (as these are taken from the Goal, and these had not been updated)


mattwills-moj-digital (approved):
Looks really good to me 😉

PR #209, opened 2023-12-12
RR-446 - Use AP helm chart instead of manual template config

nathanrussell-moj-digital:
This PR replaces the hand written helm template that sets up the cronjob to export to the Analytical Platform with a published helm chart from the Data Platform team.
From talking to them about setting up preprod and prod, they said I should make this change because they'd rather people were using their helm chart rather than a custom handwritten template.


helm_deploy/hmpps-education-and-work-plan-api/values.yaml
```diff
@@ -53,3 +53,10 @@ generic-service:
 
 generic-prometheus-alerts:
   targetApplication: hmpps-education-and-work-plan-api
+
+generic-data-analytics-extractor:
+  cronJobSchedule: "0 1 * * *" # 1am every day
+  databaseSecretName: rds-postgresql-instance-output
+  destinationS3SecretName: analytical-platform-reporting-s3-bucket
+  serviceAccountName: hmpps-education-and-work-plan-prod-to-ap-s3
+  enabled: false
```
nathanrussell-moj-digital, line 62: Leaving it disabled for now whilst we get the corresponding stuff working at the other end (ie. the stuff in the Analytical Platform that will receive this data)

mattwills-moj-digital (approved):

PR #217, opened 2023-12-14
RR-548 - Ensure timestamp and author fields are updated

mattwills-moj-digital:
We've discovered an issue where the timestamp and author fields are wrong when an `INDUCTION_UPDATED` timeline event occurs. This was due to the way the main Induction's `updatedAt`/`updatedBy`/`updatedByDisplayName` fields were not being correctly updated in most circumstances when the "Update Induction" endpoint is called. This PR ensures that these fields are now updated whenever the update endpoint is called.

### DIFF
--- a/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/UpdateInductionTest.kt
+++ b/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/UpdateInductionTest.kt
@@ -544,7 +544,7 @@ class UpdateInductionTest : IntegrationTestBase() {
       .hasPrisonId("BXI")
       .wasModifiedBy("auser_gen")
       .wasCreatedAt(persistedInduction.createdDateTime)
-      .wasLastModifiedAt(persistedInduction.modifiedDateTime) // should be unchanged
+      .wasLastModifiedAfter(persistedInduction.modifiedDateTime)
 
     assertThat(updatedInduction.workExperience).isEquivalentTo(expectedWorkExperience)
     assertThat(updatedInduction.workExperience!!.modifiedDateTime).isAfter(persistedInduction.workExperience!!.modifiedDateTime)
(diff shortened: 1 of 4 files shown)


src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/UpdateInductionTest.kt
```diff
@@ -544,7 +544,7 @@ class UpdateInductionTest : IntegrationTestBase() {
       .hasPrisonId("BXI")
       .wasModifiedBy("auser_gen")
       .wasCreatedAt(persistedInduction.createdDateTime)
-      .wasLastModifiedAt(persistedInduction.modifiedDateTime) // should be unchanged
+      .wasLastModifiedAfter(persistedInduction.modifiedDateTime)
```
mattwills-moj-digital, line 547: Due to some logic within `CiagInductionResponseMapper` (see below), if we weren't updating the `workOnRelease` object (e.g. by switching from long route to short route), then I expected the `lastModified` date to stay the same.

Coming back to it now, this seems "fundamentally wrong" and I think it makes much more sense to ensure the main Induction's timestamp etc is updated each time an induction is changed.

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/mapper/induction/CiagInductionResponseMapper.kt
```diff
@@ -42,12 +42,8 @@ class CiagInductionResponseMapper(
         inPrisonInterests = inPrisonInterestsMapper.toPrisonWorkAndEducationResponse(inPrisonInterests),
         createdBy = createdBy!!,
         createdDateTime = instantMapper.toOffsetDateTime(createdAt)!!,
-        // The main Induction domain object is immutable (the data that can be changed belongs in associated "child"
-        // objects). However, the CIAG API stores details regarding a prisoner's work aspirations at the root level,
-        // which means that any changes to these values needs to be reflected in the "modified" fields of the root
-        // Induction in the response.
```
mattwills-moj-digital, line 48: I was never entirely comfortable with this, as it felt like a fudge to the mapping between our domain model and and the externally facing CIAG API model. I think simply returning the modifiedBy/modifiedDateTime of the main Induction (which is now always updated) makes much more send.

nathanrussell-moj-digital (approved):
Amazing; thanks for sorting this so quickly 👍 🎉

PR #227, opened 2024-01-05
RR-524 - Update induction

mattwills-moj-digital:
This PR implements the update Induction endpoint (i.e. PUT `/inductions/{prisonNumber}`).

As per the swagger spec, an Induction's property (e.g. `previousWorkExperience`) is only updated if it is provided in the request. This means that when switching from say the "long route" to the "short" one, a Prisoner's previous work experience will remain in-tact, even though it is not asked/provided by the UI during the "short" question set.

Due to the fact that these fields may or may not be provided during an update request, the integration tests have been written from a "business" point of view - i.e. based on what constitutes a "short" or "long" question set, and attempt to mirror what the UI is likely to send.

### DIFF
--- a/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/UpdateInductionTest.kt
+++ b/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/UpdateInductionTest.kt
@@ -1,0 +1,449 @@
+package uk.gov.justice.digital.hmpps.educationandworkplanapi.app.resource
+
+import org.assertj.core.api.Assertions.assertThat
+import org.awaitility.kotlin.await
+import org.junit.jupiter.api.Test
+import org.mockito.ArgumentCaptor
+import org.mockito.ArgumentMatchers.eq
+import org.mockito.Mockito.times
+import org.mockito.Mockito.verify
+import org.mockito.kotlin.capture
+import org.mockito.kotlin.firstValue
+import org.springframework.http.HttpStatus
+import org.springframework.http.HttpStatus.NOT_FOUND
+import org.springframework.http.MediaType.APPLICATION_JSON
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.aValidPrisonNumber
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.aValidTokenWithEditAuthority
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.aValidTokenWithViewAuthority
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.IntegrationTestBase
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.bearerToken
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.ErrorResponse
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.HighestEducationLevel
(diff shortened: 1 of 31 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/induction/WorkOnReleaseEntityMapper.kt
```diff
@@ -31,5 +31,5 @@ interface WorkOnReleaseEntityMapper {
   @ExcludeReferenceField
   @Mapping(target = "createdAtPrison", ignore = true)
   @Mapping(target = "updatedAtPrison", source = "prisonId")
-  fun updateExistingEntityFromDto(@MappingTarget entity: WorkOnReleaseEntity, dto: UpdateWorkOnReleaseDto)
+  fun updateExistingEntityFromDto(@MappingTarget entity: WorkOnReleaseEntity, dto: UpdateWorkOnReleaseDto?)
```
mattwills-moj-digital, line 34: At present, the UI always populates the "workOnRelease" values, but it doesn't have to in future.

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -2367,6 +2367,10 @@ components:
           format: uuid
           description: The Induction's unique reference.
           example: c88a6c48-97e2-4c04-93b5-98619966447b
+        prisonId:
```
mattwills-moj-digital, line 2370: A previous omission on my part 🙄

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/mapper/induction/InPrisonInterestsResourceMapper.kt
```diff
@@ -23,4 +25,6 @@ interface InPrisonInterestsResourceMapper {
   @Mapping(target = "updatedAt", source = "lastUpdatedAt")
   @Mapping(target = "updatedAtPrison", source = "lastUpdatedAtPrison")
   fun toInPrisonInterestsResponse(inPrisonInterests: InPrisonInterests?): InPrisonInterestsResponse?
+
+  fun toUpdateInPrisonInterestsDto(request: UpdateInPrisonInterestsRequest, prisonId: String): UpdateInPrisonInterestsDto
```
mattwills-moj-digital, line 29: The mapper methods (and tests) are really simple at this level, since we're simply mapping to DTOs that have the same fields as the update request objects.

The more complicated update logic is within the entity mappers, which hasn't been changed in this PR.

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/UpdateInductionTest.kt
```diff
@@ -0,0 +1,449 @@
...
+
+    // Then
+    val actual = response.responseBody.blockFirst()
+    assertThat(actual)
+      .hasStatus(NOT_FOUND.value())
+      .hasUserMessage("Induction not found for prisoner [$prisonNumber]")
+  }
+
+  @Test
+  fun `should update induction for prisoner who is no longer looking to work`() {
```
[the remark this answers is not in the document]
nathanrussell-moj-digital: Wow! These are long tests to read, but I get it, and I think you're right about writing them from a business point of view 👍

nathanrussell-moj-digital (approved):
LGTM 👍

PR #229, opened 2024-01-08
RR-534 - Reduced detail of logged objects to prevent risk of PII exposure

nathanrussell-moj-digital:
This PR reduces the detail of logged objects to prevent risk of PII exposure


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/JpaTimelinePersistenceAdapter.kt
```diff
@@ -23,7 +23,7 @@ class JpaTimelinePersistenceAdapter(
 
   @Transactional
   override fun recordTimelineEvent(prisonNumber: String, event: TimelineEvent): TimelineEvent {
-    log.info { "Recording TimelineEvent [$event] for prisoner [$prisonNumber]" }
```
nathanrussell-moj-digital, line 26: Basically we dont want (or need) the full object logged. In the case of timeline events and sqs messages (below) they _could_ contain PII
(and yes, I know the SQS listener isn't technically in use at the moment - perhaps we consider removing that in another PR ?)

mattwills-moj-digital (approved):

PR #230, opened 2024-01-08
RR-564 - PrisonAPI swagger spec and codegen

mattwills-moj-digital:
This PR imports the [Prison API swagger spec](https://prison-api-dev.prison.service.justice.gov.uk/swagger-ui/index.html) and generates the associated model classes we need to retrieve a Prisoner's current and previous prisons.

There's no need to review the swagger spec itself, which is over 33,000 lines long! 😱


build.gradle.kts
```diff
@@ -197,22 +197,49 @@ tasks.register<GenerateTask>("buildEducationAndWorkPlanModel") {
   )
 }
 
+tasks.register<GenerateTask>("buildPrisonApiModel") {
+  validateSpec.set(true)
+  generatorName.set("kotlin-spring")
+  templateDir.set("$projectDir/src/main/resources/static/openapi/templates")
+  inputSpec.set("$projectDir/src/main/resources/static/openapi/PrisonApi.json")
```
nathanrussell-moj-digital, line 204: One for another day, and not looking to hold up this PR, but I wonder if there's some syntax or config where we could tell the codegen to get the swagger spec from a URL rather than us hosting a copy of it in the project.
Something like `inputSpec.set("https://prison-api-dev...../v3/api-docs")`

I guess the pros could be that we would always be using the latest spec and that there would be no risk or question as to whether we've modified it. And we'd not have to be hosting a massive spec that we have no idea whether its the latest or not (especially as the hmpps generated specs are not versioned!)

The cons might be that it might (probably would) slow down the builds, especially locally. Also, if a breaking change happened to the prison API we wouldnt know until our build breaks, and we'd be forced fix it

nathanrussell-moj-digital (approved):
LGTM 👍

build.gradle.kts
mattwills-moj-digital: I had the same thoughts. The main thing that put me off doing anything on the fly was the sheer size of their spec and the potential impact on our build (which will already have been slowed down by this PR alone, though hopefully not by much). It takes several seconds to download their spec as it is and I've experienced a timeout in swagger editor before, so we might want to do something about that!

Also, in reality, once our API is stable (and we've moved on to pastures new?!), I think a change to the Prison API will break our service in production long before anyone runs our build again?

PR #235, opened 2024-01-10
RR-526 - Added `prisonNumber` to `InductionResponse`

nathanrussell-moj-digital:


mattwills-moj-digital (approved):
LGTM (though I could argue the client already has the prisonNumber?!).

PR #247, opened 2024-02-06
RR-593: Adds DPR DB user

leejacobson-moj:


helm_deploy/hmpps-education-and-work-plan-api/values.yaml
```diff
@@ -42,6 +42,9 @@ generic-service:
       DB_NAME: "database_name"
       DB_USER: "database_username"
       DB_PASS: "database_password"
+    digital-prison-reporting:
```
nathanrussell-moj-digital, line 45: Have you created this secret yet in `kubectl` for the 3 namespaces?
If not we'll have to hold off merging this as it wont deploy (k8s will moan bitterly that we are referencing a secret to pass into the service pod that doesn't exist)

src/main/resources/db/migration/common/V2023.06.19.0001__add_goal_related_tables.sql
nathanrussell-moj-digital, line 1: Do we know for sure that this is going to work? (moving the flyway files)

As I understand it, flyway knows that its run a migration by calculating a hash of the migration file content, and storing that in its table.
What I don't know is whether the filename (and path) is part of the hash; and / or whether it stores the filename (and path) in the table somewhere.

I guess what we need to be 100% sure on is that flyway is not going to see these moved files as "new" files that it needs to run. That would be very (very! 😬 ) bad!

helm_deploy/hmpps-education-and-work-plan-api/values.yaml
leejacobson-moj: Yep – added to all three namespaces.

nathanrussell-moj-digital (commented):
I'm not sure about creating a `dev` profile 🤔 

It's not so much that I don't want to, it's more about understanding why we need to do it / what it is giving us (bearing in mind we've not needed one to date)

I think my biggest worry though is the move of the previously run migration files. We have got to be 100% sure it's not going to try to re-run them!

src/main/resources/db/migration/common/V2023.06.19.0001__add_goal_related_tables.sql
leejacobson-moj: I tested locally and it seems to work. Looking a the flyway schema it appears to just look at the script name and version rather than to full path.

I'll test again now though to be 100% sure.

leejacobson-moj: Yep – seems to validate the prior migrations, and just run the additional:

```
2024-02-06 12:20:39.230  INFO 17826 --- [           main] o.f.core.internal.command.DbValidate     : Successfully validated 23 migrations (execution time 00:00.021s) |  
2024-02-06 12:20:39.261  INFO 17826 --- [           main] o.f.core.internal.command.DbMigrate      : Current version of schema "public": 2023.12.05.0002 |  
2024-02-06 12:20:39.270  INFO 17826 --- [           main] o.f.core.internal.command.DbMigrate      : Migrating schema "public" to version "2024.01.06.0001 - digital prison reporting user" |  
2024-02-06 12:20:39.298  INFO 17826 --- [           main] o.f.core.internal.command.DbMigrate      : Successfully applied 1 migration to schema "public", now at version v2024.01.06.0001 (execution time 00:00.024s) |  
```

nathanrussell-moj-digital: Awesome; thanks for checking 👍

nathanrussell-moj-digital (approved):
Given that we've answered the question about whether it will re-run the migrations or not (and thankfully it doesn't 😁 ); there's no technical reason why we shouldn't do this (ie. introduce profiles).
I have no problem with it, so happy to approve; but as discussed run it past @mattwills-moj-digital as well just to make sure he is happy/aware of this change, and in case he has spotted something I've not 😁 

(We dont usually need 2 approvals on a PR, but in this case I think it would be prudent 👍 )

mattwills-moj-digital (approved):
LGTM

PR #261, opened 2024-03-15
RR-728: Adds SAR endpoint to retrieve PLP goal and induction data

leejacobson-moj:
## Overview

Adds support for SAR endpoint, enabling retrieval of goal and induction data.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -56,6 +56,7 @@ dependencies {
   implementation(project("domain:timeline"))
   implementation(project("domain:induction"))
 
+  implementation("uk.gov.justice.service.hmpps:hmpps-kotlin-spring-boot-starter:0.2.2")
   implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.7.3")
   implementation("org.springframework.boot:spring-boot-starter-security")
   implementation("org.springframework.boot:spring-boot-starter-oauth2-client")
--- a/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/SubjectAccessRequest.kt
+++ b/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/SubjectAccessRequest.kt
@@ -1,0 +1,235 @@
+package uk.gov.justice.digital.hmpps.educationandworkplanapi.app.resource
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import org.springframework.http.HttpStatus.FORBIDDEN
+import org.springframework.http.MediaType
+import org.springframework.test.web.reactive.server.WebTestClient
+import org.springframework.web.util.UriBuilder
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.aValidPrisonNumber
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.aValidTokenWithNoAuthorities
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.anotherValidPrisonNumber
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.IntegrationTestBase
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.service.SubjectAccessRequestService
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.bearerToken
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.buildAccessToken
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.ErrorResponse
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.actionplan.aValidCreateActionPlanRequest
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.actionplan.aValidCreateGoalRequest
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.assertThat
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.induction.aValidCreateInductionRequestForPrisonerLookingToWork
+import java.time.LocalDate
+
+class SubjectAccessRequest : IntegrationTestBase() {
+  companion object {
+    private const val URI_TEMPLATE = "/subject-access-request"
+  }
+
+  @Test
+  fun `should return unauthorized given no bearer token`() {
+    webTestClient.get()
+      .uri { uriBuilder: UriBuilder ->
+        uriBuilder
+          .path(URI_TEMPLATE)
+          .queryParam("prn", aValidPrisonNumber())
+          .build()
+      }
+      .exchange()
+      .expectStatus()
+      .isUnauthorized
+  }
+
+  @Test
+  fun `should return forbidden given bearer token without required role`() {
+    // When
+    val response = webTestClient.get()
+      .uri { uriBuilder: UriBuilder ->
+        uriBuilder
+          .path(URI_TEMPLATE)
+          .queryParam("prn", aValidPrisonNumber())
+          .build()
+      }
+      .bearerToken(aValidTokenWithNoAuthorities(privateKey = keyPair.private))
+      .exchange()
+      .expectStatus()
+      .isForbidden
+      .returnResult(ErrorResponse::class.java)
+
+    // Then
+    val actual = response.responseBody.blockFirst()
+    assertThat(actual)
+      .hasStatus(FORBIDDEN.value())
+      .hasUserMessage("Access Denied")
+      .hasDeveloperMessage("Access denied on uri=/subject-access-request")
+  }
+
+  @Test
+  fun `should get induction and goals for specific prisoner without date filtering`() {
+    // Given
+    val prisonerNumbers = listOf(aValidPrisonNumber(), anotherValidPrisonNumber())
+    prisonerNumbers.map {
+      createInduction(
+        prisonNumber = it,
+        createInductionRequest = aValidCreateInductionRequestForPrisonerLookingToWork(),
+      )
+      createActionPlan(
+        prisonNumber = it,
+        createActionPlanRequest = aValidCreateActionPlanRequest(
+          goals = listOf(aValidCreateGoalRequest(title = "Goal 1")),
+        ),
+      )
+    }
+
+    // When
+    val response = webTestClient.sarRequest(aValidPrisonNumber(), null, null)
+
+    // Then
+    val responseBody = response
+      .expectStatus().isOk
+      .expectBody(PLPSubjectAccessRequestContent::class.java)
+      .returnResult().responseBody
+
+    val content = responseBody!!.content
+    with(content) {
+      assertThat(induction?.prisonNumber).isEqualTo(aValidPrisonNumber())
+      assertThat(goals).hasSize(1)
+      assertThat(goals.first().title).isEqualTo("Goal 1")
+    }
+  }
+
+  @Test
+  fun `should get induction and goals for specific prisoner with from date filtering`() {
+    // Given
+    val prisonerNumbers = listOf(aValidPrisonNumber(), anotherValidPrisonNumber())
+    prisonerNumbers.map {
+      createInduction(
+        prisonNumber = it,
+        createInductionRequest = aValidCreateInductionRequestForPrisonerLookingToWork(),
+      )
+      createActionPlan(
+        prisonNumber = it,
+        createActionPlanRequest = aValidCreateActionPlanRequest(
+          goals = listOf(aValidCreateGoalRequest(title = "Goal 1")),
+        ),
+      )
+    }
+
+    // When
+    val responseFromLastWeek = webTestClient.sarRequest(aValidPrisonNumber(), LocalDate.now().minusWeeks(1), null)
+    val responseFromTomorrow = webTestClient.sarRequest(aValidPrisonNumber(), LocalDate.now().plusDays(1), null)
+
+    // Then
+    val responseBody = responseFromLastWeek
+      .expectStatus().isOk
+      .expectBody(PLPSubjectAccessRequestContent::class.java)
+      .returnResult().responseBody
+
+    val fromLastWeekContent = responseBody!!.content
+
+    with(fromLastWeekContent) {
+      assertThat(induction?.prisonNumber).isEqualTo(aValidPrisonNumber())
+      assertThat(goals).hasSize(1)
+      assertThat(goals.first().title).isEqualTo("Goal 1")
+    }
+
+    responseFromTomorrow.expectStatus().isNoContent
+  }
+
+  @Test
+  fun `should get induction and goals for specific prisoner with to date filtering`() {
+    // Given
+    val prisonerNumbers = listOf(aValidPrisonNumber(), anotherValidPrisonNumber())
+    prisonerNumbers.map {
+      createInduction(
+        prisonNumber = it,
+        createInductionRequest = aValidCreateInductionRequestForPrisonerLookingToWork(),
+      )
+      createActionPlan(
+        prisonNumber = it,
+        createActionPlanRequest = aValidCreateActionPlanRequest(
+          goals = listOf(aValidCreateGoalRequest(title = "Goal 1")),
+        ),
+      )
+    }
+
+    // When
+    val responseToLastWeek = webTestClient.sarRequest(aValidPrisonNumber(), null, LocalDate.now().minusWeeks(1))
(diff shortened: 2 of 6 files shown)


build.gradle.kts
```diff
@@ -56,6 +56,7 @@ dependencies {
   implementation(project("domain:timeline"))
   implementation(project("domain:induction"))
 
+  implementation("uk.gov.justice.service.hmpps:hmpps-kotlin-spring-boot-starter:0.2.2")
```
nathanrussell-moj-digital, line 59: I'm a bit confused with this - why do we need to bring this in? Is it not pulled in via the plugin at the top
```
id("uk.gov.justice.hmpps.gradle-spring-boot") version "5.15.3"
```
?

Are they not one and the same thing? Confused! 🤔

leejacobson-moj: My knowledge of Gradle isn't the best so I could be missing something here – but no, they're not the same and I don't believe it's included by the spring boot plugin.

The spring boot plugin is this, https://github.com/ministryofjustice/dps-gradle-spring-boot
The Kotlin spring boot starter library is this, https://github.com/ministryofjustice/hmpps-kotlin-lib

I was mostly just following what other teams have done though, and they have included both,
https://github.com/search?q=org%3Aministryofjustice+implementation%28%22uk.gov.justice.service.hmpps%3Ahmpps-kotlin-spring-boot-starter%3A0.2.1%22%29&type=code

nathanrussell-moj-digital: Cool, so I follow that they are not the same thing 👍 , but I would have thought the dps-grafle-spring-boot plugin should bring this dependency in for us 🤔 

Perhaps it's related to this comment right at the top of the [readme](https://github.com/ministryofjustice/hmpps-kotlin-lib/blob/main/README.md) of the spring boot starter library?:

> Warning
> This is a work in progress and is not ready for use yet.

Given that comment, and the fact that the library is not included in the main plugin, I wonder if we are are bit premature in doing this? Is it a bit too early for this perhaps?

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/SubjectAccessRequestService.kt
```diff
@@ -0,0 +1,46 @@
...
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.induction.Induction
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.induction.InductionNotFoundException
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.induction.service.InductionService
+import uk.gov.justice.hmpps.kotlin.sar.HmppsPrisonSubjectAccessRequestService
+import uk.gov.justice.hmpps.kotlin.sar.HmppsSubjectAccessRequestContent
+import java.time.LocalDate
+import java.time.ZoneOffset
+
+@Service
+class SubjectAccessRequestService(
```
nathanrussell-moj-digital, line 15: I _think_ the classname needs to be suffixed with `Test` in order for it to be picked up by the test runner via the build

I could be wrong about that, but for consistency with all our other tests, it should have that class name suffix. Can we rename it `SubjectAccessRequestServiceTest` please?

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -2468,6 +2493,32 @@ components:
...
+          type: array
+          items:
+            $ref: '#/components/schemas/GoalResponse'
+
+    SarSuccessResponse:
+      properties:
+        content:
+          $ref: '#/components/schemas/SubjectAccessRequestContent'
+
+    SarErrorResponse:
```
nathanrussell-moj-digital, line 2511: I think we already have the equivalent of this type, in that we already have `ErrorResponse` in this spec and is already a generated class.
I dont think we need this ... 🤔

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/SubjectAccessRequest.kt
```diff
@@ -0,0 +1,235 @@
...
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.bearerToken
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.buildAccessToken
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.ErrorResponse
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.actionplan.aValidCreateActionPlanRequest
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.actionplan.aValidCreateGoalRequest
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.assertThat
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.induction.aValidCreateInductionRequestForPrisonerLookingToWork
+import java.time.LocalDate
+
+class SubjectAccessRequest : IntegrationTestBase() {
```
nathanrussell-moj-digital, line 23: I think this class name needs to be suffixed with `Test`

leejacobson-moj: Yeah, I think you're right. Thank you – good spot. I'll update now.

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/SubjectAccessRequestService.kt
```diff
@@ -0,0 +1,46 @@
...
+  private val inductionService: InductionService,
+  private val actionPlanService: ActionPlanService,
+) : HmppsPrisonSubjectAccessRequestService {
+  data class SubjectAccessRequestContent(val induction: Induction?, val goals: List<Goal>)
```
nathanrussell-moj-digital, line 19: I think this data class is wrong ...
The `SubjectAccessRequestContent` is a REST API response type, and you have it correctly defined in the swagger spec, complete with the API model versions of the Induction and Goals.
Because it's in the swagger spec you'll find it gets generated for you, so you dont need to write the class again here.

But the bigger problem is that they actually contain different types.

The generated class is an API model class, and contains properties of induction and goals which are also API model classes. This is correct.

But your definition here is not an API model type (its in a different package), and crucially the induction and goal properties are NOT API model types. The types here are the domain classes, which we never surface through the API. The domain classes may contain fields that we would not want to surface through the REST API.

There are mappers that map between the domain classes and the corresponding REST API classes

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -298,6 +298,31 @@ paths:
...
+        - Subject Access Request
+
+      parameters:
+        - $ref: "#/components/parameters/prisonReferenceNumber"
+        - $ref: "#/components/parameters/fromDate"
+        - $ref: "#/components/parameters/toDate"
+      responses:
+        '200':
+          $ref: '#/components/responses/SarContentFound'
+        '204':
```
nathanrussell-moj-digital, line 317: I know these are status codes from the controller in the spring boot starter, but they feel weird ...
I wonder what the difference between 204 (no content found) and 209 (subject not recognised) is, and crucially what we have to implement (what exceptions do we throw??) to result in those status codes?
It feels like we need these 2 cases as specific tests in the integration test class?

FWIW, to me both cases feel like a 404!
If the answer is "return a 204 when its a prisoner but we hold no data for them" and "return a 209 when its not a recognised prisoner"  I don't know how we will determine and differentiate those 2 cases 🤔

nathanrussell-moj-digital: Looking at the controller implementation, it looks like it will be a "noContent" response (204) if the service implementation returns null. The service implementation as written does this, so I think we could (and should) write an integration test for that 👍 

Re: it returning a 209, I can't see how that could ever be the case! I wonder if we should remove the 209 from our swagger spec because I honestly dont think we could ever return that kind of response!

```
  fun getSarContentByReference(
    @RequestParam(name = "prn")
    @Parameter(description = "NOMIS Prison Reference Number")
    prn: String?,
    @RequestParam(name = "crn")
    @Parameter(description = "nDelius Case Reference Number")
    crn: String?,
    @RequestParam(value = "fromDate")
    @Parameter(description = "Optional parameter denoting minimum date of event occurrence which should be returned in the response")
    fromDate: LocalDate?,
    @RequestParam(value = "toDate")
    @Parameter(description = "Optional parameter denoting maximum date of event occurrence which should be returned in the response")
    toDate: LocalDate?,
  ): ResponseEntity<Any> {
    if (prn.isNullOrBlank() && crn.isNullOrBlank()) {
      return ResponseEntity.badRequest().body(
        ErrorResponse(
          status = HttpStatus.BAD_REQUEST,
          userMessage = "One of prn or crn must be supplied.",
          developerMessage = "One of prn or crn must be supplied.",
        ),
      )
    }

    val content = if (service is HmppsPrisonSubjectAccessRequestService && !prn.isNullOrBlank()) {
      service.getPrisonContentFor(prn, fromDate, toDate)
    } else if (service is HmppsProbationSubjectAccessRequestService && !crn.isNullOrBlank()) {
      service.getProbationContentFor(crn, fromDate, toDate)
    } else if (service is HmppsPrisonProbationSubjectAccessRequestService) {
      service.getContentFor(prn, crn, fromDate, toDate)
    } else {
      return ResponseEntity.status(209).build()
    }

    return content?.let { ResponseEntity.ok(it) } ?: ResponseEntity.noContent().build()
  }
```

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -2643,6 +2694,18 @@ components:
         application/json:
           schema:
             $ref: '#/components/schemas/InductionResponse'
+    SarContentFound:
+      description: Request successfully processed - content found
+      content:
+        application/json:
+          schema:
+            $ref: '#/components/schemas/SarSuccessResponse'
+    SarIncorrectRequest:
```
nathanrussell-moj-digital, line 2703: Rather than define a specific type for this (400 BAD REQUEST), we already have a type defined that you can reference instead 👍 
```
400ErrorResponse
```

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/SubjectAccessRequestService.kt
nathanrussell-moj-digital: I think this more like what you want the service to do, making use of the REST API model classes, and the mappers to map from domain to model :

```
package uk.gov.justice.digital.hmpps.educationandworkplanapi.app.service

import org.springframework.stereotype.Service
import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.resource.mapper.actionplan.GoalResourceMapper
import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.resource.mapper.induction.InductionResourceMapper
import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.goal.service.ActionPlanService
import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.induction.InductionNotFoundException
import uk.gov.justice.digital.hmpps.educationandworkplanapi.domain.induction.service.InductionService
import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.SubjectAccessRequestContent
import uk.gov.justice.hmpps.kotlin.sar.HmppsPrisonSubjectAccessRequestService
import uk.gov.justice.hmpps.kotlin.sar.HmppsSubjectAccessRequestContent
import java.time.LocalDate
import java.time.ZoneOffset

@Service
class SubjectAccessRequestService(
  private val inductionService: InductionService,
  private val actionPlanService: ActionPlanService,
  private val inductionResourceMapper: InductionResourceMapper,
  private val goalResourceMapper: GoalResourceMapper,
) : HmppsPrisonSubjectAccessRequestService {

  override fun getPrisonContentFor(
    prn: String,
    fromDate: LocalDate?,
    toDate: LocalDate?,
  ): HmppsSubjectAccessRequestContent? {
    val fromDateInstance = fromDate?.atStartOfDay()?.toInstant(ZoneOffset.UTC)
    val toDateInstance = toDate?.atStartOfDay()?.toInstant(ZoneOffset.UTC)

    val goals = actionPlanService.getActionPlan(prn).goals
      .filter { fromDateInstance == null || it.createdAt?.isAfter(fromDateInstance) ?: true }
      .filter { toDateInstance == null || it.createdAt?.isBefore(toDateInstance) ?: true }
      .map {
        goalResourceMapper.fromDomainToModel(it)
      }
      .toSet()

    val induction = try {
      inductionService.getInductionForPrisoner(prn)
        .takeIf { fromDateInstance == null || it.createdAt?.isAfter(fromDateInstance) ?: true }
        .takeIf { toDateInstance == null || it?.createdAt?.isBefore(toDateInstance) ?: true }
        ?.let {
          inductionResourceMapper.toInductionResponse(it)
        }
    } catch (e: InductionNotFoundException) {
      null
    }

    if (goals.isEmpty() && induction == null) return null
    return HmppsSubjectAccessRequestContent(
      content = SubjectAccessRequestContent(induction, goals),
    )
  }
}
```

nathanrussell-moj-digital:
One other thing that kind of applies to many places in this PR, and thats of terminology.

Throughout the PLP codebases (both the API and UI) we've used the term `prisonNumber` as the reference or ID of the prisoner.
(Whilst we didn't like it, it came from my previous work here with Mike Halma on Send Legal Mail who assured me the correct and official terminology was `prisonNumber`; so we have adopted that in this service as well)

With that in mind, could you replace `prn` with `prisonNumber` please?

(`prn` is a term we've not used anywhere, so I'd be cautious about us introducing and using it now, just because the downstream library uses it)


Sorry if it feels like I'm pulling this to pieces! I'm not doing that; honestly! 😬

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
nathanrussell-moj-digital: I take that back - I **can** see how a 209 would be returned (but it took me many many reads of the controller code in the library! 🤣 )

Our service impl is an implementation of `HmppsPrisonSubjectAccessRequestService` and requires a `prn`
But the check re: returning a 400 bad request only returns 400 if **both** the `prn` and `crn` query string params are blank.

Therefore, it would be possible to call the API `/subject-access-request?crn=xxxxx` - that would pass the 400 bad request check, but then return a 209
I guess its kinda correct, in that our implementation requires a `prn` but you've not specified one on the query string. So saying the `Subject Identifier is not recognised by this service` is correct 🤷 

Interestingly a HTTP 209 appears to be a non-standard response code! 🤷‍♂️ 
https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/SubjectAccessRequestTest.kt
```diff
@@ -0,0 +1,298 @@
...
+    // When
+    val response = webTestClient.sarRequest(aValidPrisonNumber(), null, null)
+
+    // Then
+    val responseBody = response
+      .expectStatus().isOk
+      .expectBody(HmppsSubjectAccessRequestContent::class.java)
+      .returnResult().responseBody
+
+    val content = objectMapper.convertValue(responseBody!!.content, SubjectAccessRequestContent::class.java)
```
nathanrussell-moj-digital, line 160: Nice 👍 
I like it that we are using the real target types, rather than reproducing them in the tests. Nice 👍

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/SubjectAccessRequestTest.kt
```diff
@@ -0,0 +1,298 @@
...
+
+    // Then
+    val actual = response.responseBody.blockFirst()
+    assertThat(actual)
+      .hasStatus(FORBIDDEN.value())
+      .hasUserMessage("Access Denied")
+      .hasDeveloperMessage("Access denied on uri=/subject-access-request")
+  }
+
+  @Test
```
nathanrussell-moj-digital, line 68: 👍 
Really thorough suite of tests here, testing all the different response status codes etc.
Tests like this are ace because they really help document the functionality etc (even though it's not "our" controller) 👍 
💯

PR #264, opened 2024-03-29
RR-741 - Use real postgres database in integration tests

nathanrussell-moj-digital:


.circleci/config.yml
```diff
@@ -15,12 +15,18 @@ executors:
   integration-test-environment:
     docker:
       - image: cimg/openjdk:21.0
-# For future use - change ITs to use a real postgres instance rather than H2
```
nathanrussell-moj-digital, line 18: We always thought we'd end up making this change sooner or later! 😁 
The approach we've used has been copied from Send Legal Mail

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/CreateInductionTest.kt
```diff
@@ -85,15 +82,11 @@ class CreateInductionTest : IntegrationTestBase() {
   }
 
   @Test
-  @Transactional
```
nathanrussell-moj-digital, line 88: A couple of existing integration tests needed to change, which actually is indicative of the core problem that this PR solves (h2 is not the same as postgres !)

We have a handful of integration tests that rely on data being in the database, typically those that update an induction or goal.
In order to setup the relevant records we do it in the `// Given` block of the test, and in some cases we really struggled with this!
Our thought process was to use the injected repository and simply save records. But we ran into transaction issues, where the transaction boundary was not clear or obvious. It resulted in us "saving" a record, but when the actual test method was run in the `// When` block, it could not find the record, ultimately because the transaction was not committed. We ended up managing the transactions ourselves as seen here, but its messy and we were never happy with it.
We decided that a better approach was to use the REST API itself to create any records etc that the test needed, and that over time we would migrate our tests to use that approach.

Fast forward to this PR ....
Swapping out `h2` with `postgres` actually caused this test to fail, which actually highlights another (but related) shortcoming of `h2`.
The test fails because the data in `aValidInductionEntity` appears to be incorrect in some way (!!) and the test fails with a foreign key violation on one of the child properties (I forget which one). The point is `h2` did not fail in this way, which points to `h2` not implementing constraints and/or foreign keys in the same way as `postgres`

This fix for this problem is to swap it out with a call to the REST API (via the helper method `createInduction`), which we were planning on doing at some point anyway!

src/integrationTest/resources/db/h2/V2023.08.15.0001__add_action_plan_fields.sql
nathanrussell-moj-digital, line 1: No need for `h2` specific DDL anymore 👍

leejacobson-moj (approved):
Looks good. Approved.

PR #265, opened 2024-04-02
RR-741 - Handle DB exceptions to prevent them being leaked in the error response

nathanrussell-moj-digital:
This PR handles DB exceptions in the GlobalExceptionHandler to prevent them being leaked in the error response
There's also some minor work around where the transaction boundary is because otherwise the induction would have been created even though it was the timeline record that failed to write (it feels correct that if the timeline failed to update, then the main request should also fail)

Without this fix, this is the error response that leaks details:
```
{
  "status": 500,
  "errorCode": null,
  "userMessage": "Unexpected error: could not execute statement [ERROR: value too long for type character varying(3)] [insert into timeline_event (timeline_id,actioned_by,actioned_by_display_name,contextual_info,correlation_id,created_at,event_type,prison_id,reference,source_reference,timestamp,id) values (?,?,?,?,?,?,?,?,?,?,?,?)]; SQL [insert into timeline_event (timeline_id,actioned_by,actioned_by_display_name,contextual_info,correlation_id,created_at,event_type,prison_id,reference,source_reference,timestamp,id) values (?,?,?,?,?,?,?,?,?,?,?,?)]",
  "developerMessage": "could not execute statement [ERROR: value too long for type character varying(3)] [insert into timeline_event (timeline_id,actioned_by,actioned_by_display_name,contextual_info,correlation_id,created_at,event_type,prison_id,reference,source_reference,timestamp,id) values (?,?,?,?,?,?,?,?,?,?,?,?)]; SQL [insert into timeline_event (timeline_id,actioned_by,actioned_by_display_name,contextual_info,correlation_id,created_at,event_type,prison_id,reference,source_reference,timestamp,id) values (?,?,?,?,?,?,?,?,?,?,?,?)]",
  "moreInfo": null
}
```

With this fix the error response is this:
```
{
  "status": 503,
  "errorCode": "Service Unavailable",
  "userMessage": "Service unavailable",
  "developerMessage": "null",
  "moreInfo": null
}
```

### DIFF
--- a/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/CreateInductionTest.kt
+++ b/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/CreateInductionTest.kt
@@ -11,6 +11,7 @@ import org.mockito.kotlin.times
 import org.mockito.kotlin.verify
 import org.springframework.http.HttpStatus.BAD_REQUEST
 import org.springframework.http.HttpStatus.FORBIDDEN
+import org.springframework.http.HttpStatus.SERVICE_UNAVAILABLE
 import org.springframework.http.MediaType.APPLICATION_JSON
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.aValidPrisonNumber
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.aValidTokenWithEditAuthority
@@ -105,6 +106,34 @@ class CreateInductionTest : IntegrationTestBase() {
     assertThat(actual)
       .hasStatus(FORBIDDEN.value())
       .hasUserMessage("An Induction already exists for prisoner $prisonNumber")
+  }
+
+  @Test
+  fun `should fail to create induction given database exception`() {
+    // Given
+    val prisonNumber = aValidPrisonNumber()
+
+    val invalidPrisonId = "longer than 3 chars" // created_at_prison column on timeline_event table is set to 3 characters, so this value results in a database exception
+    val createRequest = aValidCreateInductionRequestForPrisonerNotLookingToWork(prisonId = invalidPrisonId)
+
+    // When
+    val response = webTestClient.post()
+      .uri(URI_TEMPLATE, prisonNumber)
+      .withBody(createRequest)
+      .bearerToken(aValidTokenWithEditAuthority(privateKey = keyPair.private))
+      .contentType(APPLICATION_JSON)
+      .exchange()
+      .expectStatus()
+      .is5xxServerError
+      .returnResult(ErrorResponse::class.java)
+
+    // Then
+    val actual = response.responseBody.blockFirst()
+    assertThat(actual)
+      .hasStatus(SERVICE_UNAVAILABLE.value())
+      .hasUserMessage("Service unavailable")
+
+    inductionDoesNotExistForPrisoner(prisonNumber)
   }
 
   @Test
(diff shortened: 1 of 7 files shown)


src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/CreateInductionTest.kt
```diff
@@ -107,6 +108,34 @@ class CreateInductionTest : IntegrationTestBase() {
       .hasUserMessage("An Induction already exists for prisoner $prisonNumber")
   }
 
+  @Test
+  fun `should fail to create induction given database exception`() {
```
nathanrussell-moj-digital, line 112: Really we should be validating our inputs a little better than we are, but as it stands at the moment we can cause this database error at will because we are not validating this particular field (and it was this field in the request that the app sec field used which caused the problem)

I'll look at improving our request body validation in a subsequent PR, but doing so will likely mean this test no longer works as written, and it may be more difficult to create a database exception like this. Will have to give it some thought .... 🤔

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/GlobalExceptionHandler.kt
```diff
@@ -150,6 +152,19 @@ class GlobalExceptionHandler(
     return populateErrorResponseAndHandleExceptionInternal(e, BAD_REQUEST, request)
   }
 
+  /**
+   * Exception handler to return a 503 Service Unavailable ErrorResponse, specifically for a DataAccessException.
+   */
+  @ExceptionHandler(DataAccessException::class)
```
nathanrussell-moj-digital, line 158: New exception handler method to handle any database exception (all `spring-data` exceptions extend `DataAccessException`)

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/GlobalExceptionHandler.kt
```diff
@@ -150,6 +152,19 @@ class GlobalExceptionHandler(
...
+  fun handleDataAccessException(
+    e: DataAccessException,
+    request: WebRequest,
+  ): ResponseEntity<Any>? {
+    log.error("Unexpected database exception", e)
+    request.setAttribute(ERROR_MESSAGE, "Service unavailable", SCOPE_REQUEST)
```
nathanrussell-moj-digital, line 164: A nice vague error message that will get rendered out in the API `ErrorResponse`, but the exception itself is logged so we can still find and investigate it 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/GlobalExceptionHandler.kt
```diff
@@ -175,17 +190,10 @@ class GlobalExceptionHandler(
   }
 
   @ExceptionHandler(Exception::class)
-  fun handleException(e: Exception): ResponseEntity<ErrorResponse?>? {
+  fun unexpectedExceptionHandler(e: Exception, request: WebRequest): ResponseEntity<Any>? {
```
nathanrussell-moj-digital, line 193: This was the previous exception handler that was being used for database errors. I've refactored it so that it's response body content is equally vague, but we still log the exception.

I also had to rename the method because I needed to add `request: WebRequest` to the method args, but there is already a method `handleException(e: Exception, request: WebRequest)` in the superclass that this class extends.

leejacobson-moj: Nice – makes sense. Side thought, but it might be worth setting up Sentry to log / track these unexpected errors at some point.

leejacobson-moj (approved):
Great work! Approved.

PR #268, opened 2024-05-16
RR-719 - refactor domains to better reflect their business use case

nathanrussell-moj-digital:
This PR refactors the `domain` modules and packages to better reflect their purpose and business use cases.
The reason for this refactoring is to allow us to start to model "notes" / "conversation histories" in the `learningandworkprogress` domain as a sibling to `induction` (as both the induction and the conversation histories "belong" to Learning And Work Progress)

Note: whilst this is a massive PR, it's just about changing package names and therefore the imports. It touches pretty much every class in the repo, but its just package and import changes.

### Old domain structure
Screenshot 2024-05-16 at 09 45 35

### New domain structure
Screenshot 2024-05-16 at 13 00 36


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/SubjectAccessRequestService.kt
```diff
@@ -35,7 +35,7 @@ class SubjectAccessRequestService(
       inductionService.getInductionForPrisoner(prisonNumber)
         .takeIf { fromDateInstance == null || it.createdAt?.isAfter(fromDateInstance) ?: true }
         .takeIf { toDateInstance == null || it?.createdAt?.isBefore(toDateInstance) ?: true }
-    } catch (e: InductionNotFoundException) {
+    } catch (e: uk.gov.justice.digital.hmpps.domain.learningandworkprogress.induction.InductionNotFoundException) {
```
leejacobson-moj, line 38: Is the full class path needed here?

leejacobson-moj (commented):
Looks good. Just one minor query.

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/SubjectAccessRequestService.kt
nathanrussell-moj-digital: Blimey, good spot! 😁 
I thought I'd caught and reverted all of those! Its basically where the IDE did the refactoring and for some reason wanted to do it in steps (I think) - for some reason it fully qualified all of the exceptions like this, and then the linter moaned at me because there was an unused import!

I'll correct it 👍

leejacobson-moj (approved):
Looks good. Approved 👍

PR #273, opened 2024-05-17
RR-719 - Introduced "conversation" domain classes and services

nathanrussell-moj-digital:
This PR introduces the new domain classes, service class and persistence layer adapter interface for "conversation" (which is how we are modelling the concept of "notes")

Screenshot 2024-05-17 at 15 00 14


leejacobson-moj (approved):
Looks good. Approved 👍

PR #277, opened 2024-05-21
RR-799 - Async services using Spring's ThreadPoolTaskExecutor

nathanrussell-moj-digital:
This PR makes our async services properly async by using Spring's `ThreadPoolTaskExecutor` and the `Async` annotation.
([based on this medium post](https://medium.com/@ahmedmohamedmar3y2017/springboot-asynchronous-methods-58524b444fb4))

The original code was always meant to be async, but was flawed. It attempted to use kotlin coroutines, but the use of the `runBlocking` annotation meant that it wasn't async at all and actually was blocking the main request thread.

(There is an [alternative PR](https://github.com/ministryofjustice/hmpps-education-and-work-plan-api/pull/278) that implements these async services using kotlin coroutines)

### DIFF
--- a/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/CreateInductionTest.kt
+++ b/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/CreateInductionTest.kt
@@ -11,7 +11,6 @@ import org.mockito.kotlin.times
 import org.mockito.kotlin.verify
 import org.springframework.http.HttpStatus.BAD_REQUEST
 import org.springframework.http.HttpStatus.FORBIDDEN
-import org.springframework.http.HttpStatus.SERVICE_UNAVAILABLE
 import org.springframework.http.MediaType.APPLICATION_JSON
 import uk.gov.justice.digital.hmpps.domain.aValidPrisonNumber
 import uk.gov.justice.digital.hmpps.domain.anotherValidPrisonNumber
@@ -109,7 +108,7 @@ class CreateInductionTest : IntegrationTestBase() {
   }
 
   @Test
-  fun `should fail to create induction given database exception`() {
+  fun `should create induction but fail to create timeline event given timeline database exception`() {
     // Given
     val prisonNumber = aValidPrisonNumber()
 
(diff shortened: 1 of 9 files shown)


src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/CreateInductionTest.kt
```diff
@@ -109,31 +108,32 @@ class CreateInductionTest : IntegrationTestBase() {
   }
 
   @Test
-  fun `should fail to create induction given database exception`() {
+  fun `should create induction but fail to create timeline event given timeline database exception`() {
```
nathanrussell-moj-digital, line 111: The fact I've had to change this test proves that the previous implementation wasn't async at all!

The basis of this test is that a database exception is thrown whilst writing the timeline event, where the database exception is caused by using an invalid prison ID which is longer than the column width (yes, I know we should be doing request body validation ..... )

Anyway, the timeline event is recorded by the "async" timeline event service

Previously the entire request was failed because the timeline event service failed (with the database exception), and that exception was propagated all the way back to the controller and the global exception handler.
This proves that the timeline event service was never actually async!

Now that the timeline service is async the test has to change because we now expect the main request to succeed, and the timeline event (which is a "nice to have", but not a core part of the request) should silently fail.

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/GetTimelineTest.kt
```diff
@@ -123,57 +124,60 @@ class GetTimelineTest : IntegrationTestBase() {
...
-          .hasEventType(TimelineEventType.GOAL_CREATED)
-          .hasPrisonId("BXI")
-          .wasActionedBy("auser_gen")
-          .hasActionedByDisplayName("Albert User")
-          .hasContextualInfo(goal.title!!)
-          .hasCorrelationId(actionPlanCreatedCorrelationId)
-      }
+
+    // When
+    await.untilAsserted {
```
nathanrussell-moj-digital, line 131: Because the timeline event is now properly async we need to use awaitility to wait until these assertions pass (awaitility is setup with a 3 second timeout in the integration test superclass)

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/config/ThreadPoolConfiguration.kt
```diff
@@ -0,0 +1,18 @@
...
+import org.springframework.context.annotation.Configuration
+import org.springframework.core.task.TaskExecutor
+import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor
+
+@Configuration
+class ThreadPoolConfiguration {
+  @Bean("threadPoolTaskExecutor")
+  fun taskExecutor(): TaskExecutor =
+    ThreadPoolTaskExecutor().apply {
+      corePoolSize = 200
```
nathanrussell-moj-digital, line 13: I took these values from the medium post mentioned in the PR description - we can tweak these as necessary

leejacobson-moj (approved):
Nice work. Approved 👍

PR #278, opened 2024-05-21
RR-799 - Async services using kotlin coroutines

nathanrussell-moj-digital:
This PR makes our async services use coroutines "properly"
([kotlin coroutines](https://kotlinlang.org/docs/coroutines-basics.html))

The original code was always meant to be async, but was flawed. It attempted to use kotlin coroutines, but the use of the `runBlocking` annotation meant that it wasn't async at all and actually was blocking the main request thread.

This PR changes that so that all methods from the controller request method down through the stack are marked as `suspend` functions, and the methods in the async classes use the coroutine scope from the suspend function, rather than creating a new (blocking) scope with `runBlocking`

Whilst this seems to work and is arguably a better, more "correct" way of using coroutines that previously implemented, I'm still not sure it's properly async (or certainly not "fire and forget" which is what we want for the use cases here)

(There is an [alternative PR](https://github.com/ministryofjustice/hmpps-education-and-work-plan-api/pull/277) that implements these async services using spring's async annotation and a thread pool task executor)

### DIFF
--- a/domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/induction/service/InductionEventService.kt
+++ b/domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/induction/service/InductionEventService.kt
@@ -10,10 +10,10 @@ interface InductionEventService {
   /**
    * Implementations providing custom code for when a [Induction] is created.
    */
-  fun inductionCreated(createdInduction: Induction)
+  suspend fun inductionCreated(createdInduction: Induction)
 
   /**
    * Implementations providing custom code for when a [Induction] is updated.
    */
-  fun inductionUpdated(updatedInduction: Induction)
+  suspend fun inductionUpdated(updatedInduction: Induction)
 }
--- a/domain/learningandworkprogress/src/test/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/induction/service/InductionServiceTest.kt
+++ b/domain/learningandworkprogress/src/test/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/induction/service/InductionServiceTest.kt
@@ -1,5 +1,6 @@
 package uk.gov.justice.digital.hmpps.domain.learningandworkprogress.induction.service
 
+import kotlinx.coroutines.test.runTest
 import org.assertj.core.api.Assertions.assertThat
 import org.assertj.core.api.Assertions.catchThrowableOfType
 import org.junit.jupiter.api.Nested
@@ -41,7 +42,7 @@ class InductionServiceTest {
   @Nested
   inner class CreateInduction {
     @Test
-    fun `should create induction`() {
+    fun `should create induction`() = runTest {
       // Given
       val createInductionDto = aValidCreateInductionDto()
       val induction = aFullyPopulatedInduction()
@@ -66,7 +67,7 @@ class InductionServiceTest {
 
       // When
       val exception = catchThrowableOfType(
-        { service.createInduction(createInductionDto) },
+        { runTest { service.createInduction(createInductionDto) } },
         InductionAlreadyExistsException::class.java,
       )
 
@@ -114,7 +115,7 @@ class InductionServiceTest {
   @Nested
   inner class UpdateInduction {
     @Test
-    fun `should update induction`() {
+    fun `should update induction`() = runTest {
       // Given
       val updateInductionDto = aValidUpdateInductionDto(prisonNumber = PRISON_NUMBER)
       val expected = aFullyPopulatedInduction(prisonNumber = PRISON_NUMBER)
@@ -137,7 +138,7 @@ class InductionServiceTest {
 
       // When
       val exception = catchThrowableOfType(
-        { service.updateInduction(updateInductionDto) },
+        { runTest { service.updateInduction(updateInductionDto) } },
         InductionNotFoundException::class.java,
       )
 
--- a/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/CreateInductionTest.kt
+++ b/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/CreateInductionTest.kt
@@ -2,6 +2,7 @@ package uk.gov.justice.digital.hmpps.educationandworkplanapi.app.resource
 
 import org.assertj.core.api.Assertions.assertThat
 import org.awaitility.kotlin.await
+import org.junit.jupiter.api.Disabled
 import org.junit.jupiter.api.Test
 import org.mockito.ArgumentCaptor
 import org.mockito.kotlin.capture
@@ -11,7 +12,6 @@ import org.mockito.kotlin.times
 import org.mockito.kotlin.verify
 import org.springframework.http.HttpStatus.BAD_REQUEST
 import org.springframework.http.HttpStatus.FORBIDDEN
-import org.springframework.http.HttpStatus.SERVICE_UNAVAILABLE
 import org.springframework.http.MediaType.APPLICATION_JSON
 import uk.gov.justice.digital.hmpps.domain.aValidPrisonNumber
 import uk.gov.justice.digital.hmpps.domain.anotherValidPrisonNumber
@@ -109,7 +109,8 @@ class CreateInductionTest : IntegrationTestBase() {
   }
 
   @Test
-  fun `should fail to create induction given database exception`() {
+  @Disabled("test is indeterminate due to the async nature of the coroutine - sometimes it will succeed meaning the controller has returned before the timeline database error; other times it will fail because the timeline database error happens before the controller has returned and the exception is propagated back")
+  fun `should create induction but fail to create timeline event given timeline database exception`() {
     // Given
     val prisonNumber = aValidPrisonNumber()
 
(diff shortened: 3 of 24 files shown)


domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/induction/service/InductionEventService.kt
```diff
@@ -10,10 +10,10 @@ interface InductionEventService {
   /**
    * Implementations providing custom code for when a [Induction] is created.
    */
-  fun inductionCreated(createdInduction: Induction)
+  suspend fun inductionCreated(createdInduction: Induction)
```
nathanrussell-moj-digital, line 13: Having to declare the interface method sigs as `suspend` here in the domain feels slightly dirty to me. It feels like an implementation detail is bleeding into the domain, and our domain is meant to be "pure" / "clean" making no framework etc demands on the implementation.

(I accept its a kotlin project and environment, so perhaps not a major issue, but you can see the effect this method sig change has on the java bytecode in `UpdateGoalRequestReferenceConstraintValidatorTest` below 👇 )

domain/learningandworkprogress/src/test/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/induction/service/InductionServiceTest.kt
```diff
@@ -41,7 +42,7 @@ class InductionServiceTest {
   @Nested
   inner class CreateInduction {
     @Test
-    fun `should create induction`() {
+    fun `should create induction`() = runTest {
```
nathanrussell-moj-digital, line 45: All the tests that invoke a `suspend` function now need to be wrapped in `runTest` which provides the coroutine scope to the function

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/CreateInductionTest.kt
```diff
@@ -109,31 +109,33 @@ class CreateInductionTest : IntegrationTestBase() {
   }
 
   @Test
-  fun `should fail to create induction given database exception`() {
+  @Disabled("test is indeterminate due to the async nature of the coroutine - sometimes it will succeed meaning the controller has returned before the timeline database error; other times it will fail because the timeline database error happens before the controller has returned and the exception is propagated back")
+  fun `should create induction but fail to create timeline event given timeline database exception`() {
```
nathanrussell-moj-digital, line 113: The changes to this test are the exact same changes I made in this test in the ThreadPoolTaskExecutor based PR, and the changes were made for the same reason.

However, in this coroutine based implementation the test is indeterminate - sometimes it will pass which means the main request thread has completed and returned before the async timeline event database write has failed.
But other times the test will fail, which means the async timeline event database write failed before the main request thread completed, and the exception propagated back to the controller and its global exception handler.

Notwithstanding the fact that actually we should be validating our inputs better such that this scenario can never happen, the fact that it's indeterminate means that the timeline event service is not properly "fire and forget"

Also, its actually the source of a memory leak that I've read about online. If the main request thread has completed before the timeline event service fails and throws it's exception ..... where does that exception go? There is talk online about this as a memory leak, but I'm not sure of best approach and strategy for addressing it

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/AsyncActionPlanEventService.kt
```diff
@@ -20,15 +20,15 @@ class AsyncActionPlanEventService(
   private val timelineEventFactory: TimelineEventFactory,
   private val timelineService: TimelineService,
 ) : ActionPlanEventService {
-  override fun actionPlanCreated(actionPlan: ActionPlan) {
-    runBlocking {
+  override suspend fun actionPlanCreated(actionPlan: ActionPlan) {
+    coroutineScope {
```
nathanrussell-moj-digital, line 24: Use the coroutine scope from the suspend function, rather than creating a blocking scope

src/test/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/validator/UpdateGoalRequestReferenceConstraintValidatorTest.kt
```diff
@@ -21,8 +22,11 @@ class UpdateGoalRequestReferenceConstraintValidatorTest {
     UpdateGoalRequest::class.java,
     String::class.java,
     UUID::class.java,
+    Continuation::class.java,
```
nathanrussell-moj-digital, line 25: This is the test that I mentioned above re: the use of the `suspend` function affecting the java byte code.
In this case we are using reflection to get the method from the controller. As part of the reflection call we need to specify the method name and it's arguments, then reflection does it's magic and finds the method.

Because we have marked the method as a `suspend` function in the controller, it has changed the method in the byte code by adding a new arg to it - the `Continuation` argument. Note - this does not change the argument list in the source code or that calling code need to call it with; it's just at the byte code that it's different.

Sad actual times 😢

domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/induction/service/InductionEventService.kt
mikehalmamoj: "This is the way". Kotlin is all about bytecode manipulation I'm afraid. In essence the big win is that converting to async code is as simple as adding the suspend keyword then using async/launch in the method. If you were to try to achieve the same outcomes using Java futures you'd maybe come to appreciate the elegance of Kotlin suspend functions.

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/CreateInductionTest.kt
mikehalmamoj: The problem is that by reusing the parent CoroutineScope you've bumped into [Structured Concurreny](https://kotlinlang.org/docs/coroutines-basics.html#structured-concurrency) - this is where Kotlin keeps track of all child coroutines and cancels both the parent and siblings if any coroutine fails.

Please see [this commit](https://github.com/ministryofjustice/hmpps-education-and-work-plan-api/pull/284/commits/8e3f23b2b9607f404e6a9cbafe6fe99497ecc0bf) - it creates a new CoroutineScope that is not linked to the parent scope and as such achieves the "fire and forget" behaviour you desire.

src/test/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/validator/UpdateGoalRequestReferenceConstraintValidatorTest.kt
mikehalmamoj: Sad times indeed! 😆 I can see why you want unit tests here but is the problem that Kotlin suspend functions are dirty, or is the problem that using reflection in this way is less than ideal? I'd be tempted to try and achieve the same with an integration test. For me that would give more confidence that your constraints are working as intended, this feels a little bit contrived?

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/CreateInductionTest.kt
nathanrussell-moj-digital: Ah, I see. So we use a spring bean that is just for these "fire & forget" coroutine scopes! Got it! That makes sense 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/AsyncActionPlanEventService.kt
nathanrussell-moj-digital: As per @mikehalmamoj's suggested approach to doing fire & forget, this is wrong, and should be the injected scope bean 👍

src/test/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/validator/UpdateGoalRequestReferenceConstraintValidatorTest.kt
nathanrussell-moj-digital: I take your points, though I'm not sure the test is that contrived.
We have a couple of custom constraint validators and generally our view re testing them is to test them 'thoroughly' in unit tests, and test them once in an integration test. Unit tests are cheap and fast; integration tests less so
For example we have a constraint validator that asserts a date is not in the past. Our unit test tests the validator with "today", "tomorrow" and "yesterday" (and it would 'cost' nothing to add more dates/scenarios if the complexity of the validator warranted it).
We also have this date validator covered in one integration test. That test is not about testing whether the validator works in all scenarios, but is about essentially testing whether everything is wired up and that the request is a rejected with a 400 with the correct messages etc.

Back to this validator and why it has a unit test .... whilst the validator itself is simple (does the reference number in the request URI path parameter equals the reference number in the request body object?) the implementation is quite complex in that it needs to be a `SupportedValidationTarget` validator in order to get the controller method args rather than an annotation on a single field. The test needs a reference to the controller class and the method that the annotation is used on; hence the nastiness of reflection in order to get the `Method` object.

mikehalmamoj: All good points. Though I would say that the old testing pyramid is a bit of a different shape these days... more of a trapezoid? 😆 Since Docker came along, integration tests aren't half as expensive as they used to be (though it is a bit painful when you want to run a single integration test locally and you have to wait for Spring to start up). For example we've given up on unit tests where we're interested in the database, integration tests only, mocking repositories is painful and of little value.

PR #279, opened 2024-05-21
RR-797: Conversation create endpoint

leejacobson-moj:


src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -2461,6 +2482,37 @@ components:
...
+      description: A request to create a Prisoner Conversation.
+      type: object
+      properties:
+        prisonId:
+          type: string
+          description: The identifier of the prison that the prisoner is currently resident at.
+          example: BXI
+        type:
+          $ref: '#/components/schemas/ConversationType'
+        note:
```
leejacobson-moj, line 2506: For now `note` is a string and a required field, but it's possible in time this may become an optional string array.

It may be the case that at some point a prisoner "Conversation" could contain items other than notes, and a conversation may eventually contain many notes.

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -262,6 +262,27 @@ paths:
...
+  #
+  # Conversations
+  # --------------------------------------------------------------------------------
+  '/conversations/{prisonNumber}':
+    parameters:
+      - $ref: "#/components/parameters/prisonNumberPathParameter"
+    post:
+      summary: Creates a Learning Work and Progress Conversation for a Prisoner.
+      description: |
+        Creates a Learning Work and Progress Conversation for a Prisoner (if it does not already exist).
```
nathanrussell-moj-digital, line 274: Minor, but I'm not sure about the `(if it does not already exist)` because that suggests if the conversation does exist you'll get some error; and/or there is something in the request that makes it unique in some way.

How about something like:
```
      description: |
        Creates a Learning Work and Progress Conversation for a Prisoner.
        A Prisoner can have many Conversations. Sending the same request will result in a new Conversation being
        saved for the Prisoner.
```

nathanrussell-moj-digital (dismissed):
This is fantastic; and apart from my very minor comment you've done a cracking job here.

One thing we do normally do when we make a swagger spec change is to bump the version number at the top of the file (semantic versioning). But in this case, where we will be making several changes to add 4 new Conversation based endpoints, I'd be happy that we bump the version when we do the last endpoint PR 👍 

Looks really good to me, and I'm happy if you want to merge this, and then crack on with defining the next endpoint 👍

nathanrussell-moj-digital (approved):
LGTM 👍

PR #285, opened 2024-05-23
RR-797: Get all conversations endpoint

leejacobson-moj:


src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -262,6 +262,80 @@ paths:
...
+    put:
+      summary: Updates a Learning Work and Progress Conversation for a Prisoner.
+      description: |
+        Updates a prisoner's Learning Work and Progress Conversation.
+      tags:
+        - Conversations
+      responses:
+        '204':
+          description: The Conversation was updated successfully.
+        '400':
```
nathanrussell-moj-digital, line 333: This `PUT` endpoint probably wants to return a 404 as well?

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -2651,6 +2830,43 @@ components:
...
+        Access-Control-Allow-Headers:
+          schema:
+            type: string
+      content:
+        application/json:
+          schema:
+            type: array
+            description: A List of prisoners' conversations.
+            items:
+              $ref: '#/components/schemas/ConversationResponse'
```
nathanrussell-moj-digital, line 2868: I think that rather than defining the items array here, I think it might be better to create a `#/components/schemas/ConversationsResponse` or `#/components/schemas/ConversationListResponse` (or whatever you want to call it) that contains the `items` array (just like we do currently with responses like `ActionPlanSummaryList` for example)

I think doing it that way might be useful moving forwards when that response type becomes a bit richer perhaps with more properties? I'm thinking that as and when we introduce paging etc, then that response object graph will not only contain an `items` array, but will also contain things like the paging data (page number, page size etc etc)

What do you think?

nathanrussell-moj-digital (approved):
LGTM 👍 

Lets get it merged, then we can both crack on - one of us can be implementing the endpoints in the API codebase, the other can import this spec into the UI codebase to generate the types and then start building out the client and service methods 👍

PR #287, opened 2024-05-28
RR-831 - Refactor Timeline Event contextualInfo to be stored and returned as a map

nathanrussell-moj-digital:
This PR refactors the `contextualInfo` field of a `TimelineEvent` from a single string field into a map.

Currently a timeline event has an optional field for storing some kind of "contextual" info, or meta data, about the timeline event. At the moment its a single string field, and we currently use it to store the goal or step title when the timeline event is relating to a goal or step. And in the case that the timeline event is relating to a prison transfer we store the prison ID that the prisoner is transferring from.

The purpose of the `contextualInfo` field was a bit of an after thought in the design of timeline events, and is really so that the UI can display other related or contextual info about the event itself. When we first added it Matt and I knew it was slightly limiting and that one day we would want to refactor it into some kind of map or key/value pairing.

The work we are currently doing re: adding notes/conversation functionality involves recording new timeline events when the user creates or updates a conversation, and as part of that we will likely need some richer contextual info, so doing this refactoring now is right thing to do.


domain/timeline/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/timeline/TimelineEvent.kt
```diff
@@ -23,9 +23,9 @@ data class TimelineEvent(
    */
   val eventType: TimelineEventType,
   /**
-   * Some useful contextual information about the event. For example a reason for the event.
+   * A map of useful contextual information about the event.
    */
-  val contextualInfo: String?,
+  val contextualInfo: Map<TimelineEventContext, String>?,
```
nathanrussell-moj-digital, line 28: We are changing `contextualInfo` from being a simple string to a `Map<TimelineEventContext, String>`, where `TimelineEventContext` is an enum of values (defined below)

This is the domain class; the rest of the changes in the PR relate to the database, API and mappers etc where we are making essentially the same changes

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/GetTimelineTest.kt
```diff
@@ -155,7 +155,7 @@ class GetTimelineTest : IntegrationTestBase() {
             .hasPrisonId("BXI")
             .wasActionedBy("system")
             .hasNoActionedByDisplayName()
-            .hasContextualInfo("MDI")
+            .hasContextualInfo(mapOf("PRISON_TRANSFERRED_FROM" to "MDI"))
```
nathanrussell-moj-digital, line 158: Here the contextual info is a `Map<String, String>` - you might wonder why its not keyed on `TimelineEventContext` 🤔 

At the API its a `Map<String, String>` because we define the APIs in the swagger spec first and it generates our model classes. See the changes below to the swagger spec, but I couldn't work out how to define a map keyed by a type (rather than keyed by string)

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/mapper/timeline/TimelineEventResourceMapper.kt
```diff
@@ -16,6 +17,14 @@ import java.util.UUID
...
+abstract class TimelineEventResourceMapper {
+  @Mapping(target = "contextualInfo", expression = "java( buildContextualInfo(timelineEventDomain) )")
+  abstract fun fromDomainToModel(timelineEventDomain: TimelineEvent): TimelineEventResponse
+
+  protected fun buildContextualInfo(timelineEventDomain: TimelineEvent): Map<String, String> =
+    if (timelineEventDomain.contextualInfo !== null) {
+      timelineEventDomain.contextualInfo!!.mapKeys { it.key.toString() }
+    } else {
+      emptyMap()
+    }
```
leejacobson-moj, line 29: Maybe it's a syntax preference thing, but you could use an Elvis operator here instead,

```suggestion
  protected fun buildContextualInfo(timelineEventDomain: TimelineEvent): Map<String, String> =
    timelineEventDomain.contextualInfo?.mapKeys { it.key.toString() } ?: emptyMap()
```

leejacobson-moj (approved):
Looks good. Approved 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/mapper/timeline/TimelineEventResourceMapper.kt
nathanrussell-moj-digital: Darn it! I've only just seen this comment after merging it!
You are right, that is a better syntax. I forgot about the elvis operator, and started off trying with a ternary which kotlin doesnt support! hence the if/else !
I'll change it to use the elvis operator in my next PR 👍

PR #292, opened 2024-05-30
RR-797: Update conversation endpoint implementation

leejacobson-moj:
## Overview

Adds the implementation for the update conversation endpoint.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/JpaConversationPersistenceAdapter.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/JpaConversationPersistenceAdapter.kt
@@ -27,6 +27,7 @@ class JpaConversationPersistenceAdapter(
     val conversationEntity = conversationRepository.findByReference(updateConversationDto.reference)
     return if (conversationEntity != null) {
       conversationEntityMapper.updateEntityFromDto(conversationEntity, updateConversationDto)
+      conversationEntity.updateLastUpdatedAt() // force the main Induction's JPA managed fields to update
       val persistedEntity = conversationRepository.saveAndFlush(conversationEntity)
       conversationEntityMapper.fromEntityToDomain(persistedEntity)
     } else {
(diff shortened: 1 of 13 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/JpaConversationPersistenceAdapter.kt
```diff
@@ -27,6 +27,7 @@ class JpaConversationPersistenceAdapter(
     val conversationEntity = conversationRepository.findByReference(updateConversationDto.reference)
     return if (conversationEntity != null) {
       conversationEntityMapper.updateEntityFromDto(conversationEntity, updateConversationDto)
+      conversationEntity.updateLastUpdatedAt() // force the main Induction's JPA managed fields to update
```
leejacobson-moj, line 30: After our conversation @nathanrussell-moj-digital I took a look at Inductions to see if it had a similar problem and found this line was forcing the JPA managed fields to update even when the parent entity itself had no updates. This seemed like the easiest solution to the problem I was having so replicated here.

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/ConversationController.kt
```diff
@@ -29,12 +32,26 @@ class ConversationController(
   @ResponseStatus(HttpStatus.CREATED)
   @PreAuthorize(HAS_EDIT_AUTHORITY)
   @Transactional
-  fun createInduction(
+  fun createConversation(
```
leejacobson-moj, line 35: Missed this copy and paste typo in the last PR.

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -2731,8 +2731,6 @@ components:
           type: string
           description: The identifier of the prison that the prisoner is currently resident at.
           example: BXI
-        type:
-          $ref: '#/components/schemas/ConversationType'
```
leejacobson-moj, line 2735: I think I included this by accident. There's no user need to update the conversation type (at least not currently).

nathanrussell-moj-digital (approved):
LGTM 👍

PR #295, opened 2024-06-04
RR-797: Get conversation endpoint implementation

leejacobson-moj:
## Overview

Adds the implementation for the get conversation endpoint.

This PR also removes the `/{prisonNumber}` path param from the get and update conversation endpoints since it wasn't necessary.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/conversation/ConversationEntityMapper.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/conversation/ConversationEntityMapper.kt
@@ -36,5 +36,7 @@ interface ConversationEntityMapper {
   @ExcludeReferenceField
   @Mapping(target = "note.content", source = "noteContent")
   @Mapping(target = "note.updatedAtPrison", source = "prisonId")
+  @Mapping(target = "type", ignore = true)
+  @Mapping(target = "prisonNumber", ignore = true)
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/conversation/ConversationEntityMapper.kt
```diff
@@ -36,5 +36,7 @@ interface ConversationEntityMapper {
   @ExcludeReferenceField
   @Mapping(target = "note.content", source = "noteContent")
   @Mapping(target = "note.updatedAtPrison", source = "prisonId")
+  @Mapping(target = "type", ignore = true)
+  @Mapping(target = "prisonNumber", ignore = true)
```
leejacobson-moj, line 40: This was needed to fixed those mapstruct warnings.

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -1599,6 +1599,18 @@ components:
...
+        - prisonNumber
+        - note
+        - createdBy
+        - createdByDisplayName
+        - createdAt
+        - createdAtPrison
+        - updatedBy
+        - updatedByDisplayName
+        - updatedAt
+        - updatedAtPrison
```
leejacobson-moj, line 1613: Not 100% sure why, but this fixed those mapstruct write errors I was encountering. I think maybe when the fields are nullable mapstruct will try to call the constructor with null values then attempt to update the individual fields which fails since they are immutable.

nathanrussell-moj-digital (approved):
LGTM 👍

PR #298, opened 2024-06-11
RR-845: Add "not relevant" option to previous work experience

james-baker-digital-justice:


src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
nathanrussell-moj-digital, line 1: I forgot to mention, but we also bump the version of the swagger spec whenever we make changes to it (it's easier to track changes when devs of consumers come to us saying "I'm using your API v1.8.1". From that we can look in the git history etc to see how the API might have changed since then etc)

Anyway, not to worry in this instance as we are targeting the integration branch, so as long as one of our PRs for the integration branch for this epic bumps it then we will be fine 👍 (I'll do it in the PR I'm working through at the moment)

domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/induction/PreviousWorkExperiences.kt
```diff
@@ -49,3 +49,9 @@ enum class WorkExperienceType {
   CLEANING_AND_MAINTENANCE,
   OTHER,
 }
+
+enum class HasWorkedBeforeDomain {
```
nathanrussell-moj-digital, line 53: Minor comment on this (and the enum at the entity level) .... is there a reason why you've suffixed the classname with `Domain` and `Entity`?
I don't particularly mind, but if it was to differentiate the classes in some way, their package hierarchy should do that sufficiently for us? 🤔 

If however it was to do with import clashes in classes like the mappers, where you are mapping to/from `HasWorkedBefore` as an API type to a domain type for example .... kotlin has you covered here 😁 
With kotlin you can do import aliasing, so you'd be able to do this:
```
import uk.gov.justice.digital.hmpps.domain.learningandworkprogress.induction.HasWorkedBefore as HasWorkedBeforeDomain
...
import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.HasWorkedBefore
```

(I don't mind though if you want to name them this way for explicitness 👍 )

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/entity/induction/PreviousWorkExperiencesEntity.kt
```diff
@@ -51,7 +51,7 @@ class PreviousWorkExperiencesEntity(
   @field:NotNull
   var reference: UUID? = null,
 
-  var hasWorkedBefore: Boolean? = null,
+  var hasWorkedBefore: HasWorkedBeforeEntity? = null,
```
nathanrussell-moj-digital, line 54: I think you're going to need a couple of annotations here:
```
@field:NotNull
@Enumerated(value = EnumType.STRING)
```
I think the `@field:NotNull` is largely syntactic sugar. It's a validation annotation, but we don't (currently) validate the entities, instead relying on JSR303 validation on the REST API request.
Having said that I think it's useful to have it on the class if nothing else as a hint to other devs that this field is not intended to be null (and we probably have a non-null constraint on the table in it's original DDL 🤔 )

The other annotation, `@Enumerated(value = EnumType.STRING)`, is more important.
Without it we'll be storing the enum's ordinal index in the database column, rather than it's name. We want the enum name stored in the database, because it then makes more sense when querying the database directly (also remembering that the entire database is exported to the Reporting Platform. They don't need to know that 0 == YES, 1 == NO and 2 == NOT_RELEVANT 🤣 )

nathanrussell-moj-digital (commented):
Looks good; though a minor comment about a couple of annotations that I think we need on the property in the entity class

domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/induction/PreviousWorkExperiences.kt
james-baker-digital-justice: Yeah it was to distinguish them for imports really, there is a clash in a mapper but also when importing into tests, etc. you have to pick 1 of 3. TBH if there wasn't precedent for duplicating them I'd just use the auto-generated one in all 3 cases given they have the same values on only specialise if there is some sort of drift.

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/entity/induction/PreviousWorkExperiencesEntity.kt
james-baker-digital-justice: There is no not-null constraint currently on this field although the mapping I did would set all nulls to NO. I prefer it isn't null in the kotlin type also if it's not null in the DB. Don't know why others don't do that?

james-baker-digital-justice: `@field:NotNull` but also `HasWorkedBeforeEntity? = null,` 🤔 😂

nathanrussell-moj-digital: The reason we don't make the fields non-nullable (IIRC 🤔 ) is because of the way JPA works, specifically when retrieving records from the database and creating entity class instances. IIRC it uses reflection and (I think) uses a no-args constructor (even though we dont have one - we use the `kotlin-jpa` plugin which provides a no-args constructor) and reflection to set the individual fields. I think I remember this correctly (from a previous non-MoJ project a while back), but I think there are some circumstances where if the fields are non-nullable kotlin gets very grumpy! 🤣 

You'll also notice our entity classes are regular kotlin classes, and not `data` classes. Again, this was specific best-practice all relating to using kotlin classes with JPA.

See an [example here](https://github.com/JetBrains/jpa-buddy-kotlin-entities) (even though they give an example of `data` classes - I'm sure we had trouble with that), but notice the use of `nullable = false` on the `@Column` annotation, even though the kotlin field is nullable (and ` var` rather than `val` so it's value can be set after construction)

domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/induction/PreviousWorkExperiences.kt
nathanrussell-moj-digital: Hopefully you wouldnt be able to do that; not in the case of the domain anyway.
The way we've organised the modules and the dependencies etc, then domain module can't see the main application so therefore wouldn't be able to see & import the classes from the main module (either the generated or the entity classes)

This was very deliberate. The reason the domain is separate is that some of this stuff doesn't actually belong in our repo at all! Timeline for example ... it would be much better if there were a common/central HMPPS timeline service; and maybe all we'd need to do was import a jar as a dependency if we wanted to do "timeline" stuff
Likewise goals and steps .... When they move to OnePlan we wont own that domain, but we will (almost certainly) still need some REST API endpoints relating to them in some way.

Anyway, 3 layers (albeit with duplication) & mappers FTW 😁 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/entity/induction/PreviousWorkExperiencesEntity.kt
nathanrussell-moj-digital: I forgot to mention above, the best-practice I'm referring to was from [JPA Buddy](https://jpa-buddy.com/) which is a fantastic plug-in for intellij and will generate your entity classes etc (and their website also has a load of guides and best practice recommendations on the subject 👍 )

nathanrussell-moj-digital (approved):
LGTM 👍 
Thanks for making the suggested changes 👍

PR #299, opened 2024-06-11
RR-848 - Removed deprecated /ciag/induction endpoints and all associated classes

nathanrussell-moj-digital:
This PR removes the deprecated `/ciag/inductions/` REST API endpoints, and all associated classes.
The only "ciag" endpoint that is left is `/ciag/inductions/list` which the UI still consumes, but we are likely to remove that as well at some point in the future.


src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
nathanrussell-moj-digital, line 1: The bulk of the change is here in the swagger spec.
The version has been bumped (semantic versioning - arguably it should be `2.0.0` ?)

And the API endpoints and all associated types have been removed.

Because the types have been removed this has caused most of the other change in this PR in respect of mappers and test fixtures data being deleted 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/CiagInductionController.kt
```diff
@@ -1,75 +1,26 @@
...
-  fun updateInduction(
-    @Valid
-    @RequestBody
-    request: UpdateCiagInductionRequest,
-    @PathVariable @Pattern(regexp = PRISON_NUMBER_FORMAT) prisonNumber: String,
-  ) {
-    inductionService.updateInduction(updateInductionRequestMapper.toUpdateInductionDto(prisonNumber, request))
-  }
-
   @PostMapping("/list")
```
nathanrussell-moj-digital, line 24: The only remaining "ciag" API endpoint is now `/ciag/inductions/list` 👍

james-baker-digital-justice (approved):
Nice

PR #302, opened 2024-06-12
NO_JIRA - fixed deprecated assertion method calls following update to hmpps spring-boot-gradle:6

nathanrussell-moj-digital:
This PR fixes a load of build warnings about the use of a deprecated version of the assertJ method `catchThrowableOfType`
This has come about since the recent update to HMPPS `spring-boot-gradle-plugin:6.0.0` which presumably brings with it a later version of Spring which in turn brings a later version of assertJ as a transitive dependency 🤷‍♂️ 

Anyway, this PR changes the previous use of `catchThrowableOfType` to a different method overload (same method, it's just the args are in a different order)

It also commits the change to `.trivyignore` which again is the result of updating to HMPPS `spring-boot-gradle-plugin:6.0.0` 

And also some changes to the use of some JWTS code, again based on updating `spring-boot-gradle-plugin` which brings in a newer version of JWTS which deprecates some of the stuff we were using previously!

### DIFF
--- a/.trivyignore
+++ b/.trivyignore
@@ -10,6 +10,3 @@
 # Suppression for h2 2.1.214 password on command line vulnerability
 #   can be suppressed as we only run h2 locally and not on build environments
 CVE-2022-45868
-# Suppression for logback-classic and logback-core as we don't let third parties control our appenders.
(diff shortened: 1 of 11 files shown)


james-baker-digital-justice (approved):
LGTM

PR #307, opened 2024-06-20
RR-865 - Added `hasWorkedBeforeNotRelevantReason` field to API, domain and database

nathanrussell-moj-digital:
This PR is the API side of the fix to provide the free text reason field for when the prisoner has answered "Have you worked before" as `NOT_RELEVANT`

It's an optional field in the API, and is therefore nullable in the domain and entity classes, and the database table.
Therefore this change can safely be merged and deployed even though the UI part of this fix has not been done yet.


james-baker-digital-justice (approved):
LGTM

PR #308, opened 2024-06-21
RR-856 Archive APIs

james-baker-digital-justice:
Add APIs to archive and unarchive a goal.


domain/personallearningplan/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/personallearningplan/service/GoalService.kt
```diff
@@ -84,6 +100,73 @@ class GoalService(
       }
   }
 
+  /**
+   * Archives a [Goal], identified by its `prisonNumber` and `goalReference`, from the specified [ArchiveGoalDto].
+   */
+  fun archiveGoal(prisonNumber: String, archiveGoalDto: ArchiveGoalDto): Either<ArchiveGoalProblem, Goal> {
```
james-baker-digital-justice, line 106: Happy to change this from Either/Problem style to some other but just wanted to demo it.

domain/personallearningplan/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/personallearningplan/service/GoalPersistenceAdapter.kt
```diff
@@ -29,4 +31,14 @@ interface GoalPersistenceAdapter {
    * Updates a [Goal] identified by its `prisonNumber` and [UpdateGoalDto.reference]
    */
   fun updateGoal(prisonNumber: String, updatedGoalDto: UpdateGoalDto): Goal?
+
+  /**
+   * Archives a [Goal] identified by its `prisonNumber` and [ArchiveGoalDTO.reference]
+   */
+  fun archiveGoal(prisonNumber: String, archiveGoalDto: ArchiveGoalDto): Goal?
```
james-baker-digital-justice, line 38: I feel like we should add prison number to the DTO as it's required to make the request/identify the goal

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -1444,6 +1504,38 @@ components:
       required:
         - status
 
+    ArchiveGoalRequest:
+      title: ArchiveGoalRequest
+      type: object
+      description: A request to archive an individual Goal.
+      properties:
+        goalReference:
```
james-baker-digital-justice, line 1512: feels a bit weird to duplicate the goal reference between URL and body but prison number is only in the URL. again, did it this way to match existing pattern

domain/personallearningplan/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/personallearningplan/Goal.kt
```diff
@@ -2,9 +2,10 @@ package uk.gov.justice.digital.hmpps.domain.personallearningplan
 
 import uk.gov.justice.digital.hmpps.domain.personallearningplan.GoalStatus.ARCHIVED
 import uk.gov.justice.digital.hmpps.domain.personallearningplan.GoalStatus.COMPLETED
+import uk.gov.justice.digital.hmpps.domain.personallearningplan.dto.ReasonToArchiveGoal
 import java.time.Instant
 import java.time.LocalDate
-import java.util.UUID
+import java.util.*
```
nathanrussell-moj-digital, line 8: I don't know how your code formatter and linter has let you get away with this?
I was sure the configured lint rules objected to wildcard imports, and the code formatter was configured to not use them (or only use wildcards when there were at least 5?) or something like that?

Anyway, I guess the linter is happy 👍  ...... but its sending my OCD bananas and is eating me alive ....! 🤣

nathanrussell-moj-digital: (Don't worry; I'll try not to lose too much sleep over it! 🤣 )

domain/personallearningplan/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/personallearningplan/service/GoalService.kt
nathanrussell-moj-digital: It's definitely good to see this as a demo 👍 , and for the most part I can understand what it is doing, but I'll check out the branch in my IDE to have a proper play with it

I'm struggling with a couple of bits in this class .... the method chaining from `toOption` to `toEither` then a series of `flatMap` operations ...
It seems weird that you have to construct the ~exception~ problem class twice . eg. once on line 111 and again on line 115
And the calls to `right` and `left` (lines 115 and 118) are hard to read/follow

And the call to `Unit.right()` on line 147?

domain/personallearningplan/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/personallearningplan/service/GoalService.kt
```diff
@@ -84,6 +100,73 @@ class GoalService(
...
+      .flatMap { validateGoalIsValidForUnarchiving(it, prisonNumber, goalReference) }
+      .flatMap {
+        goalPersistenceAdapter.unarchiveGoal(prisonNumber, unarchiveGoalDto)?.right() ?: GoalToBeUnarchivedCouldNotBeFound(
+          prisonNumber,
+          goalReference,
+        ).left()
+      }
+  }
+
+  private fun validateArchiveDto(
```
nathanrussell-moj-digital, line 140: I know that not having to declare the return types on kotlin functions is a big plus for the language, but I think in cases like these 3 private methods it would add some value, especially when reading the code outside of an IDE

I honestly can't tell you from reading the code what these 3 functions return. At a guess I would say an `Either<Goal, ArchiveGoalProblem>` ? 🤷‍♂️

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/GoalController.kt
```diff
@@ -63,4 +75,65 @@ class GoalController(
...
+  @PreAuthorize(HAS_EDIT_AUTHORITY)
+  @Transactional
+  fun <T> archiveGoal(
+    @Valid
+    @RequestBody
+    archiveGoalRequest: ArchiveGoalRequest,
+    @PathVariable @Pattern(regexp = PRISON_NUMBER_FORMAT) prisonNumber: String,
+    @PathVariable goalReference: UUID,
+  ): ResponseEntity<Any> {
+    return goalService.archiveGoal(
```
nathanrussell-moj-digital, line 90: I'm really not sure about this 🤔 , but I know this was the big win that you were after, in that you wanted to be able to see/follow the flow of the code, rather than it being hidden away in the global exception handler.

If we adopted this approach across all of our controllers we're going to add so much code 🤔
AFAIK the "best practice / recommended" (by spring evangelists) approach is to use a `@ControllerAdvice` - hence our global exception handler class. 
Given HMPPS is a Spring house I wonder what the benefits of deviating too far from standard practices might be?
Might we be in danger of violating the principal of least surprise? IE. other devs might be expecting a more standard/conventional approach of using a `@ControllerAdvice` class?

domain/personallearningplan/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/personallearningplan/Goal.kt
james-baker-digital-justice: I've got everything set to a large number so it shouldn't be doing it. For some reason it's java.util so there must be a separate setting somewhere

domain/personallearningplan/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/personallearningplan/service/GoalService.kt
james-baker-digital-justice: Close, it's the other way round because left = bad, right = good

domain/personallearningplan/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/personallearningplan/service/GoalService.kt
james-baker-digital-justice: Yes, appreciate it might be a bit jarring if you are not used to an either monad.

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/GoalController.kt
james-baker-digital-justice: Using Exceptions for control flow is generally considered an anti-pattern (I don't know why spring thought this was a good idea). It's an even worse offence in Kotlin as there are no checked exceptions so you lose the benefit of documenting your "bad" path in the method signature and, not that we have any, they do not play nicely with coroutines either, you must be very careful.
Appreciate this is a spring shop but to me, least surprise means I can see all the error cases in the controller and they match the swagger spec. If you use ControllerAdvise it happens in another class and is only testable via integration test instead of POJO unit tests. The code in the ControllerAdvice will be almost identical.

domain/personallearningplan/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/personallearningplan/service/GoalService.kt
james-baker-digital-justice: pushed some updates with return types and using the constructors rather than short left/right methods.

domain/personallearningplan/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/personallearningplan/service/GoalService.kt
```diff
@@ -84,6 +99,71 @@ class GoalService(
...
   }
 
+  /**
+   * Archives a [Goal], identified by its `prisonNumber` and `goalReference`, from the specified [ArchiveGoalDto].
+   */
+  fun archiveGoal(prisonNumber: String, archiveGoalDto: ArchiveGoalDto): Either<ArchiveGoalProblem, Goal> {
+    val goalReference = archiveGoalDto.reference
+    log.info { "Archiving Goal with reference [$goalReference] for prisoner [$prisonNumber] because [${archiveGoalDto.reason}] with description [${archiveGoalDto.reasonOther ?: ""}]" }
+
+    return goalPersistenceAdapter.getGoal(prisonNumber, goalReference)
```
nathanrussell-moj-digital, line 109: Oooohhhh, how does `goalPersistenceAdapter.getGoal` return something that has a `.orProblem` method?
```
return goalPersistenceAdapter.getGoal(prisonNumber, goalReference)
      .orProblem(GoalToBeArchivedCouldNotBeFound(prisonNumber, goalReference))
      ....
```

As far as I can tell in this PR the `getGoal` method in the persistence adapter has not been changed and returns a nullable `Goal` ?
```
  /**
   * Returns a [Goal] identified by its `prisonNumber` and `goalReference` if found, otherwise `null`.
   */
  fun getGoal(prisonNumber: String, goalReference: UUID): Goal?
```

Has the inclusion of the `arrows` library essentially added an extension function or similar on every type? 😬

nathanrussell-moj-digital: Doh! scrap that!, I've just seen where you add the extension function towards the end of this class
Panic over! 🤣

nathanrussell-moj-digital: Not quite panic over .... the ext function you added below is this:
```
private fun <T> Goal?.orProblem(problem: T): Either<T, Goal> = this?.right() ?: problem.left()
```

`this` in  `this?.right()` is the `Goal`; therefore something somewhere has added a `right` function to the `Goal` (and presumably also a `left` function)
Clicking into the `right` function it takes me to the `arrows` source code:
```
public fun <A> A.right(): arrow.core.Either<kotlin.Nothing, A> { /* compiled code */ }
```
It seems that every class now has a `left` and `right` function (and all the others that presumably `arrows` has augmented into them) 🤔 . EG. using typeahead I can see that the `UUID` class now has these functions:

Screenshot 2024-06-25 at 08 28 26

nathanrussell-moj-digital (commented):
I think there's some good stuff and interesting approaches in this PR.

I think the `arrows` library looks interesting, so thank you for introducing it to me - I will definitely dig into it a little more. Though the use of extension functions on (what would seem) every type is making me pause for thought a little. I guess we'd want to make sure we understand what it is doing and any potential implications, and just make sure we are happy with this. And maybe the "we" in this context is the wider HMPPS dev community given that we will be handing this project over to them at some point (rather than just our team or the Farsight dev teams) (though I do accept that we did not do this when we introduced mapstruct into this project)

And no doubt `arrows` has some great use cases; I'm just wondering whether this is one of them .... 🤔 

I am a little worried about deviating from the spring approach of `@ControllerAdvice` for exception handling and returning an appropriate response. It's not that what you're proposing is "wrong" as such - it totally works, and would equally well whether we use `arrows`, exceptions in a `try/catch`, or some other library/technique.
And I get what you are wanting from this - to be able to "see" and follow/click through etc the code in the IDE at both design and runtime; to be able to unit test (rather than integration test) the controller etc
I get it ... but it does deviate from the Spring approach of keeping the controllers really slim and using a `@ControllerAdvice` ... and I wonder how wise that is 🤔 

I think what we should do is keep this branch and PR as a reference, but not merge it at this time (perhaps give the PR the label of `Do not merge` ?)
I think we should create a new branch and PR that delivers the archive/unarchive API endpoints that we need for the story (based on what you have in this PR, but keeping to the existing patterns and approaches for now). Lets get that done and merged so we can at least keep the momentum of the epic that we are trying to deliver.
And perhaps we go out to the wider HMPPS dev community and canvas opinion on not using `@ControllerAdvice` and secondly introducing `arrows` as a project dependency.

Does tis work for you?

---

As a side note, and this is less about conforming to HMPPS standards etc; but the introduction of a library into our Domain has broken one of the core principles of how we envisaged the Domain working a little 😬 
Our Domains are separate gradle modules, where the main app depends on the Domain module. This was very deliberate as it could be argued that PLP doesn't "own" some of the domain (specifically `Timeline`, and with the advent of `OnePlan` the `Goal` itself!)
Notwithstanding who "owns" a given domain, rather than build a "dumb" crud application where the REST API ends up being tightly coupled (mirrored) to the database schema, we wanted to adopt something more akin to a hexagonal architecture or clean architecture (what we've ended up with is based on those two ideas). (We had been bitten before on a previous project which resulted in some very uncomfortable coupling between the REST and SQS APIs)
Kinda going back to who owns the domain, but we also envisaged the Domain being a standalone library jar (assuming one day HMPPS have a means of publishing jars). 
Our thought is that the Domain should simply model business types and behaviour, but should be 100% agnostic to use case / framework / etc. If a Spring application wants to do something with the Goal domain it can pull in the jar and implement the persistence adapter. Equally if a Quarkus application wanted to .... or a Jersey application .... or even an application that was not built around any framework .... It just a library and all of this should be possible.
To do this we consider the Domain "pure". We dont use any framework libraries in the Domain dependencies ; and the externally facing API (ie. the service classes) are "pure". They operate on standard java types, or the types exposed by the Domain.
What you're proposing here, but using the `arrows` `Either` type as the return type of the service class methods breaks this. It means that any application that depends on the Domain must also use `arrows`. And there might be several reasons why a project team might not want to do that.
If we go ahead with this, and then one day our Domain modules are moved to some other team to manage and publish as a library jar, then it means that all consuming code will need to use `arrows`.

domain/personallearningplan/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/personallearningplan/service/GoalService.kt
james-baker-digital-justice: Yeah left() and right() are just extension functions provided by arrow that create Either.Left or Either.Right wrappers around your type. They are much more concise for things like that orProblem method I created. It's a more functional style so appreciate it might take some getting used to.

nathanrussell-moj-digital (approved):
LGTM 👍 
Sorry if this has felt like a battle to get it over the line, but I think its been some valuable discussion

PR #312, opened 2024-06-27
RR-828 - Timeline and telemetry events for archiving and unarchiving goals

nathanrussell-moj-digital:
This PR  makes the API changes to create Timeline events for archiving and unarchiving goals

Whilst not written or mentioned in the story we actually record 2 types of events. We record a `TimelineEvent` in out database and return it in the `GET /timelines/{prisonNumber}` API; but we also send an event to App Insights

The story (RR-828) only talks about Timeline (because it's written in context of the UI Timeline, and therefore App Insights is not of interest); but I've also added the code to send the App Insights events as well for completeness


domain/personallearningplan/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/personallearningplan/service/GoalService.kt
```diff
@@ -99,39 +99,48 @@ class GoalService(
...
   fun archiveGoal(prisonNumber: String, archiveGoalDto: ArchiveGoalDto): ArchiveGoalResult {
     val goalReference = archiveGoalDto.reference
     log.info { "Archiving Goal with reference [$goalReference] for prisoner [$prisonNumber] because [${archiveGoalDto.reason}] with description [${archiveGoalDto.reasonOther ?: ""}]" }
 
-    val existingGoal = goalPersistenceAdapter.getGoal(prisonNumber, goalReference)
-    return if (existingGoal == null) {
-      GoalToBeArchivedCouldNotBeFound(prisonNumber, goalReference)
-    } else if (checkReasonIsSpecifiedIfOther(archiveGoalDto)) {
-      ArchiveReasonIsOtherButNoDescriptionProvided(
+    if (checkReasonIsSpecifiedIfOther(archiveGoalDto)) {
```
nathanrussell-moj-digital, line 109: Little bit of tidying, but I've moved the call to `checkReasonIsSpecifiedIfOther` to before we try to retrieve the goal from the persistenceAdapter because we can do that validation early. Fail early and save a call to the database 👍

domain/personallearningplan/src/test/kotlin/uk/gov/justice/digital/hmpps/domain/personallearningplan/service/GoalServiceTest.kt
```diff
@@ -247,11 +247,13 @@ class GoalServiceTest {
       val prisonNumber = aValidPrisonNumber()
       val goalReference = aValidReference()
       val archiveGoal = aValidArchiveGoalDto(goalReference)
-      given(goalPersistenceAdapter.getGoal(prisonNumber, goalReference)).willReturn(null)
+      given(goalPersistenceAdapter.getGoal(any(), any())).willReturn(null)
```
nathanrussell-moj-digital, line 250: Our general pattern/approach with mockito is to setup our `given`s with `any()` args; and then `verify` the call was made with the expected args.
This is something we (me and the previous dev team) have done for years. The thinking being that if you setup `given`s with specific args and then the implementation changes in some way in respect of the args that the method is called with, you'll (hopefully) end up with a broken test (which is good), but it'll be broken most likely because of an NPE or similar which can be tricky to track down.
If you setup the `given`s with `any` arg matchers, then the call to the dependency will return the expected data no matter if the implementation changes in some way in respect of the args that the method is called with
So the test will pass down to the main `assertThat` (ie. asserting the return value from the method under test)
Then when we call `verify` on the mocks with the specific expected args, we'll get a clear failure message (and the test will fail) if the mock was called with different args

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/IntegrationTestBase.kt
```diff
@@ -135,12 +136,18 @@ abstract class IntegrationTestBase {
...
 
   fun getTimeline(prisonNumber: String): TimelineResponse =
-    webTestClient.get()
-      .uri(GET_TIMELINE_URI_TEMPLATE, prisonNumber)
-      .bearerToken(aValidTokenWithViewAuthority(privateKey = keyPair.private))
-      .exchange()
-      .returnResult(TimelineResponse::class.java)
-      .responseBody.blockFirst()!!
+    run {
+      wiremockService.stubGetPrisonTimelineFromPrisonApi(
```
nathanrussell-moj-digital, line 140: Minor fix here to stub the call to prison API to get an empty set of prison movements for the prisoner
This is because the call to our `GET /timelines/{prisonNumber}` (ie. what the web test client is doing) calls prison api to get prisoner movements, and also queries our timeline database.
Without the stub to prison api, whilst it didnt break any tests because we have suitable error handling in the timeline service class, it did result in quite a lot of log noise when the tests run

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/ArchiveGoalTest.kt
```diff
@@ -78,6 +87,28 @@ class ArchiveGoalTest : IntegrationTestBase() {
           .hasArchiveReason(ReasonToArchiveGoal.OTHER)
           .hasArchiveReasonOther(reasonOther)
       }
+
+    await.untilAsserted {
```
nathanrussell-moj-digital, line 91: We can now assert that the Timeline Event was correctly recorded, and that the telemetry (App Insights) event was sent

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/TelemetryEventTypeResolver.kt
```diff
@@ -37,12 +33,6 @@ class TelemetryEventTypeResolver {
       updateTypes.add(GOAL_UPDATED)
     }
 
-    // check if the goal's status was updated
```
nathanrussell-moj-digital, line 40: As mentioned above, the goal status is no longer changed via a goal update operation ( we no longer support it at the AP) hence all this code can be removed,

james-baker-digital-justice (approved):
LGTM

PR #317, opened 2024-07-02
RR-800 - Added request validation for `prisonId` fields

nathanrussell-moj-digital:
This PR implements format validation for the `prisonId` field in our API request bodies.

This problem actually came up as part of our appSec security scan where they identified that submitting a request with a malformed prisonId cause and exposed a SQL exception (that they quite rightly said could be used to garner information and prepare an attack)
Whilst we acknowledged that lack of validation was the root cause it also identified the fact that our code that wrote to the timeline database (which is the db write that failed) was not async fire and forget.
At the time we chose to use this problem to fix our async code so that it was fire and forget and that any errors were not thrown back and returned in the main request thread. Hence the test that this PR now removes/changes.

We wrote this tech-debt ticket to address when we had time to fix/implement the validation.
This is what this PR now implements 👍

### DIFF
--- a/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/CreateInductionTest.kt
+++ b/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/CreateInductionTest.kt
@@ -110,32 +110,30 @@ class CreateInductionTest : IntegrationTestBase() {
   }
 
   @Test
-  fun `should create induction but fail to create timeline event given timeline database exception`() {
+  fun `should fail to create induction given malformed prison id`() {
     // Given
     val prisonNumber = aValidPrisonNumber()
 
-    val invalidPrisonId = "longer than 3 chars" // created_at_prison column on timeline_event table is set to 3 characters, so this value results in a database exception
+    val invalidPrisonId = "does not meet pattern of 3 upper case letters"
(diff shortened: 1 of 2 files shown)


src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/CreateInductionTest.kt
```diff
@@ -110,32 +110,30 @@ class CreateInductionTest : IntegrationTestBase() {
   }
 
   @Test
-  fun `should create induction but fail to create timeline event given timeline database exception`() {
```
nathanrussell-moj-digital, line 113: This was the original test we wrote. Whilst it didnt address the validation issue, it did ensure that timeline events were written to the database asynchronously, and that any errors were not thrown back to the main request thread.

Now that we have validation in place (via the changes in the swagger spec) the test as written fails, and has been rewritten to assert a 400 BAD_REQUEST

I see no value in implementing this same test for all the API endpoints that accept a `prisonId` in the request body (otherwise we'd be writing "request validation" tests for every field in every API request)

james-baker-digital-justice (approved):
Personally I prefer to avoid really strict validation of inputs beyond something that's an actual technical constraint e.g., length, enum. 
See numerous discussions on the internet for "how to validate an email address" :)
Anyway, I don't see the harm in this case so LGTM

PR #319, opened 2024-07-08
RR-759: Prevent session creation on auth failure

james-baker-digital-justice:
APSEC reported that a JSESSIONID cookie is being set in some scenarios.  We make no use of the session and therefore session hijack is not an issue, however it could conceivably be a denial of service attack vector.

If unauthenticated, Spring security was rejecting the request and creating the session so that it could store the original URL and redirect back to it if this was part of an authentication flow. That is something not supported in our auth model and can be disabled.


nathanrussell-moj-digital (approved):
Cool; LGTM 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/config/ResourceServerConfiguration.kt
```diff
@@ -40,4 +42,11 @@ class ResourceServerConfiguration {
 
     return http.build()
   }
+
+  @Bean
```
nathanrussell-moj-digital, line 46: Just a thought; I wonder if this is the type of thing we should contribute back to the main template project? It would then save all future projects having this picked up by AppSec and having to make the same change as us? 🤔 
Just a thought;

PR #325, opened 2024-07-18
NO-JIRA - Change the start time of the generic-data-analytics-extractor job to a time when the database is up!

nathanrussell-moj-digital:
This PR changes the start time of the `generic-data-analytics-extractor` job in `dev` and `preprod` to 07:00 Monday-Friday which is a time when the database is up!
Previously the job ran at 01:00 7 days a week. However in `dev` and `preprod` we terminate resources (like the database!) out of working hours to save a few pennies, so the job could never attach to the database! Doh! 🤣 

This does not change the time in `prod`, which is still set to run at 01:00 every day.

See discussion here: https://mojdt.slack.com/archives/C05BDTUL64E/p1721288755297809


james-baker-digital-justice (approved):
LGTM

PR #327, opened 2024-07-31
RR-900 - Refactor PreviousQualifications so that they are associated with the prisoner and not the induction

nathanrussell-moj-digital:
This PR refactors `PreviousQualifications` so that they are associated with the Prisoner rather than the Induction.
This is an enabler which will allow us to design and create the REST APIs to create/read/update a Prisoner's previous qualifications (eg: GCSEs etc) before the Induction is created.

Whilst this is quite a big PR 😬 , it is essentially a change to the database tables and entity classes to add `prisonNumber` to `PreviousQualifications` and remove the ID of the `PreviousQualifications` from the `Induction`
Fundamentally that's it, but clearly there was some change needed in respect of the actual process of creating and updating Inductions, as previously it relied on JPA and the object relationships between `Induction` and `PreviousQualifications`. Now that we have removed that relationship in the entity classes JPA wont work in this respect any longer, so we need some code to write the Previous Qualifications ourselves.

At the end of the day this is a database schema change. In the spike and story we had agreed that the REST API for Inductions should not change; but extending that train of thought, the domain should not change in that the business still talk about qualifications being part of the induction (ie. qualifications should still be a child of an induction at both the API and domain) 
Because of this the right place to make this change is in the implementation of the `InductionPersistenceAdapter` - ie. `JpaInductionPersistenceAdapter`


domain/learningandworkprogress/src/testFixtures/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/induction/InductionBuilder.kt
```diff
@@ -7,12 +7,12 @@ fun aFullyPopulatedInduction(
   reference: UUID = UUID.randomUUID(),
   prisonNumber: String = "A1234AB",
   workOnRelease: WorkOnRelease = aValidWorkOnRelease(),
-  previousQualifications: PreviousQualifications = aValidPreviousQualifications(),
+  previousQualifications: PreviousQualifications? = aValidPreviousQualifications(),
```
nathanrussell-moj-digital, line 10: Minor correction to the builder here - these properties are all nullable 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/JpaInductionPersistenceAdapter.kt
```diff
@@ -7,35 +7,45 @@ import uk.gov.justice.digital.hmpps.domain.learningandworkprogress.induction.Ind
 import uk.gov.justice.digital.hmpps.domain.learningandworkprogress.induction.dto.CreateInductionDto
 import uk.gov.justice.digital.hmpps.domain.learningandworkprogress.induction.dto.UpdateInductionDto
 import uk.gov.justice.digital.hmpps.domain.learningandworkprogress.induction.service.InductionPersistenceAdapter
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.induction.PreviousQualificationsEntity
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.mapper.induction.InductionEntityMapper
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.mapper.induction.PreviousQualificationsEntityMapper
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.repository.InductionRepository
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.repository.PreviousQualificationsRepository
```
nathanrussell-moj-digital, line 14: This class is where the lions share of the change is in respect of writing the PreviousQualifications when we create or update an Induction.

There is a little bit of logic that needs to happen which is all thoroughly tested in the unit test class. Some of it I've written "java style" with `if` statements as I felt the more condensed kotlin approach was harder to read. But happy to refactor this if there is a better approach 👍 

The logic is:

### When Creating an Induction
* If the prisoner has no qualifications already recorded (ie. before Induction) and the create induction DTO contains qualifications
  * Create the Induction and create the PreviousQualifications
* If the prisoner has qualifications already recorded (ie. before Induction) and the create induction DTO contains qualifications
  * Create the Induction and update the PreviousQualifications
* If the prisoner does not have qualifications already recorded (ie. before Induction) and the create induction DTO does not contain qualifications
  * Create the Induction and do not create PreviousQualifications
* If the prisoner has qualifications already recorded (ie. before Induction) and the create induction DTO does not contain qualifications
  * Create the Induction and delete the existing PreviousQualifications

### When Updating an Induction
* If the prisoner has no qualifications already recorded and the update induction DTO contains qualifications
  * Update the Induction and create the PreviousQualifications
* If the prisoner has qualifications already recorded and the create induction DTO contains qualifications
  * Update the Induction and update the PreviousQualifications
* If the prisoner does not have qualifications already recorded and the create induction DTO does not contain qualifications
  * Update the Induction and do not create or update PreviousQualifications
* If the prisoner has qualifications already recorded and the create induction DTO does not contain qualifications
  * Create the Induction and do not update PreviousQualifications

This last scenario may sound surprising but is specifically mentioned in the swagger spec for updating an Induction:
> If a field (or "child" object) is provided in the request, it will completely overwrite the existing value. Conversely, if a field is not included in the request, then the existing value will remain untouched (including its updatedAt/updatedBy values).
And there is an integration test already in place that proves this functionality.

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/entity/induction/PreviousQualificationsEntity.kt
```diff
@@ -36,8 +36,6 @@ import java.util.UUID
  * Holds details about a Prisoner's educational qualifications, including where relevant, the grades achieved in each
  * subject.
  *
- * Note that the list of `qualifications` can be empty, but `educationLevel` is mandatory (but only if the Prisoner has
- * been asked about their education).
```
nathanrussell-moj-digital, line 40: This comment is no longer true/relevant (since we did the question set changes a few weeks back) 
(and to be fair that comment should not really have been related to the database layer; it was more relevant at the API and possibly domain layers)

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/induction/InductionEntityMapper.kt
```diff
@@ -61,11 +61,26 @@ abstract class InductionEntityMapper {
   @Mapping(target = "updatedAtPrison", source = "prisonId")
   abstract fun fromCreateDtoToEntity(dto: CreateInductionDto): InductionEntity
 
-  @Mapping(target = "lastUpdatedBy", source = "updatedBy")
-  @Mapping(target = "lastUpdatedByDisplayName", source = "updatedByDisplayName")
-  @Mapping(target = "lastUpdatedAt", source = "updatedAt")
-  @Mapping(target = "lastUpdatedAtPrison", source = "updatedAtPrison")
-  abstract fun fromEntityToDomain(persistedEntity: InductionEntity): Induction
+  fun fromEntityToDomain(inductionEntity: InductionEntity, previousQualificationsEntity: PreviousQualificationsEntity?): Induction =
```
nathanrussell-moj-digital, line 64: Here's an example of where MapStruct mappers don't need to be generated classes (or more specifically dont need to be entirely generated)
In respect of this change I started off by adding more `@Mapping` annotations to the method signature to reflect how the mapping needs to work. It ended up being an unnecessary amount of annotations for what it was worth, and was easier and neater to simply write the implementation as I have done here.

The point I guess I'm making is that we can treat these mapping functions on a case by case basis. Where there's value in writing the implementation ourselves then we can. Equally where it's easier/neater to just write the method sig and let MapStruct generate the implementation then we can do that as well.
It's flexible 👍

james-baker-digital-justice: Looks good, I agree with providing an impl. My tolerance for keeping mapstruct on some of these is probably lower than yours and if it was up to me, anything more than a couple of annotations or any `java foo(entity, dto)` type mappings should be purged.

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/JpaInductionPersistenceAdapter.kt
james-baker-digital-justice: When the new standalone flow for qualifications is added I assume that can be used and we can deprecate previous quals from the update induction flow?

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/JpaInductionPersistenceAdapter.kt
```diff
@@ -46,4 +56,55 @@ class JpaInductionPersistenceAdapter(
...
+  }
+
+  private fun createOrUpdatePreviousQualifications(createInductionDto: CreateInductionDto, prisonNumber: String): PreviousQualificationsEntity? {
+    val previousQualificationsEntity = previousQualificationsRepository.findByPrisonNumber(prisonNumber)
+
+    if (createInductionDto.previousQualifications == null) {
+      // If previousQualifications on the DTO is null it means the CIAG is creating an Induction with the specific intent of the prisoner having no previous qualifications.
+      // In this case we should delete the previousQualifications entity if it exists (because regardless of any previously added qualifications for the prisoner, the CIAG is
+      // saying that the Induction should be created with no qualifications)
+      previousQualificationsEntity?.also { previousQualificationsRepository.delete(it) }
```
james-baker-digital-justice, line 91: This makes me slightly nervous but I assume we'll have loaded the qualifications during the induction and the CIAG would've explicitly deleted them during induction?

src/main/resources/db/migration/common/V2024.06.30.0001__associate_qualifications_with_prisoner.sql
```diff
@@ -0,0 +1,24 @@
...
+(
+    prison_number
+);
+
+--- 2. Populate `previous_qualifications.prison_number` from the associated induction record
+UPDATE previous_qualifications pq
+SET prison_number = (
+    SELECT i.prison_number
+    FROM induction i
+    WHERE i.previous_qualifications_id = pq.id
```
james-baker-digital-justice, line 19: Looks correct to me

james-baker-digital-justice (dismissed):
A few questions & comments but LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/JpaInductionPersistenceAdapter.kt
nathanrussell-moj-digital: Yes, I agree, I'm equally nervous about it; but it does feel like the right outcome.

The scenario is that a prisoner has had qualifications added before induction
Then, at some point after, the CIAG does the induction. As part of doing the induction their previously entered quals are retrieved and pre-setup and displayed on the Qualifications page of the induction.

* If the CIAG is happy to accept them, they can move on through the induction, then at the end when we call the API it will be a case of the prisoner already has quals, and the DTO has quals, and they are the same
* If the CIAG wants to edit them (remove one or more, add some, etc) they can, and they move on through the induction. Then at the end when we call the API it will be a case of the prisoner already has quals, and the DTO has quals, and they are different so the quals in the database get updated
* If the CIAG is not happy with them all and removes them all, then this is the use case we are talking about here. In this case the existing quals for the prisoner should be deleted.

Personally I think it's a real edge case in respect of numbers of people actually doing it, but I think we do need to support it.
I think I'll add a log statement or two so that we can at least retrospectively see that people have done it.

src/main/resources/db/migration/common/V2024.06.30.0001__associate_qualifications_with_prisoner.sql
nathanrussell-moj-digital: Done, and tested locally 👍

james-baker-digital-justice (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/JpaInductionPersistenceAdapter.kt
nathanrussell-moj-digital: I've added some logging, which has highlighted there are 3 cases where you can end up with no qualifications
Of those 3 only one of them is the JPA `delete` that you've highlighted in this comment. 
The more I think about it the more I think it wont happen in practice (the other 2 will)
I think it's right to keep the code in because it _could_ happen, but only if we changed how our UI creates and sends the Create Induction message (so that it sends a null PreviousQualifications object - at the moment I dont think it ever would based on the question set changes we made a few weeks back)
But we might change that user journey (again 🤷‍♂️ ), or our API might get called by something other than our UI ? 🤷‍♂️ 

So I think it's right to have the code coverage for that eventuality

PR #328, opened 2024-08-01
RR-901 - Swagger spec for new pre-prison-education API endpoints

nathanrussell-moj-digital:
Swagger spec for new pre-prison-education API endpoints

Screenshot 2024-08-01 at 17 15 38
Screenshot 2024-08-01 at 17 15 54
Screenshot 2024-08-01 at 17 16 04

### DIFF
--- a/src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
+++ b/src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   title: Education and Work Plan API
-  version: '1.14.0'
+  version: '1.15.0'
   description: Education and Work Plan API
   contact:
     name: Learning and Work Progress team
@@ -17,6 +17,77 @@ paths:
   # RESTful style endpoints start here
   # --------------------------------------------------------------------------------
   #
+
+  #
+  # Education
+  # --------------------------------------------------------------------------------
+  '/person/{prisonNumber}/education':
+    parameters:
+      - $ref: "#/components/parameters/prisonNumberPathParameter"
+    description: | 
+      API endpoints to create, read and update an individual's education.  
+      
+      Personal Learning Plans currently considers "education" as meaning an individual's Highest Level of Education and
+      achieved qualifications. Typically the term "qualifications" is used to refer to things like formal school,
+      college and university education. It can in theory include other education providers, but specifically excludes
+      education that has been taken and delivered whilst in-prison. In-prison education is mastered and returned by
+      the Curious API.
+      
+      An individual's education record contains their highest level of education and their achieved qualifications.
+      An individual cannot have more than one record of education.
+    get:
+      summary: Returns an individual's education record.
+      description: | 
+        Returns an individual's education which includes their highest level of education, and a collection of 
+        qualifications that have been achieved.
+      operationId: get-education
+      tags:
+        - Education
+      responses:
+        '200':
+          $ref: '#/components/responses/Education'
+        '404':
+          $ref: '#/components/responses/404ErrorResponse'
+    post:
+      summary: Creates a record of an individual's education.
+      description: |
+        Creates a record of an individual's education which includes their highest level of education, and a
+        collection of qualifications that have been achieved.  
+        
+        A `CONFLICT 409` response is returned if there is already an education record for the individual.
+      operationId: create-education
+      tags:
+        - Education
+      requestBody:
+        $ref: '#/components/requestBodies/CreateEducation'
+      responses:
+        '201':
+          description: The Education record was created successfully.
+        '400':
+          $ref: '#/components/responses/400ErrorResponse'
+        '409':
+          $ref: '#/components/responses/409ErrorResponse'
+    put:
+      summary: Updates an individual's education.
+      description: |
+        Updates an individual's education.  
+        
+        The entire education record is updated as one. Updating just the highest level of education, or individual
+        qualifications (including adding, removing or updating) is not currently supported.
+
+        A `NOT_FOUND 404` response is returned if the specified individual does not currently have an education record.
+      operationId: update-education
+      tags:
+        - Education
+      requestBody:
+        $ref: '#/components/requestBodies/UpdateEducation'
+      responses:
+        '201':
+          description: The Education record was successfully updated.
+        '400':
+          $ref: '#/components/responses/400ErrorResponse'
+        '404':
+          $ref: '#/components/responses/404ErrorResponse'
 
   #
   # Action Plans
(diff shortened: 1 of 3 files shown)


src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -449,6 +520,54 @@ components:
           type: string
       required:
         - status
+    AuditedResource:
```
nathanrussell-moj-digital, line 523: I've created a new type, which is essentially a base class / super class.
We've copy and pasted these 8 fields into pretty much every response type, so in an attempt to reduce the size of the swagger spec file I've introduced this, then will reference it where we use those fields.
At the moment I've only referenced it in the new response types I've introduced here, plus one other place as an example, but we can create future PRs to use this type as a base type over time 👍

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -1304,6 +1423,8 @@ components:
       title: ConversationResponse
       type: object
       description: Represents a Conversation with a prisoner within the context of their Learning and Work Progress.
+      allOf:
```
nathanrussell-moj-digital, line 1426: This is the existing type that I've changed to use the `AuditedResource` base type; we can do a similar thing for other types in the future.
I didnt want to do them all in this PR so as to try to keep this PR reasonably focussed on the new pre-prison education API endpoints 👍

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
james-baker-digital-justice: I'm not sure how this translates into the generated code if this is just a swagger optimisation and it comes out flattened then I think it's fine. Otherwise I would suggest we should favour composition over inheritance and we could have this as it's own type with an `auditInfo` or similar field on each object.

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -18,6 +18,77 @@ paths:
   # --------------------------------------------------------------------------------
   #
 
+  #
+  # Pre-prison Education
+  # --------------------------------------------------------------------------------
+  '/person/{prisonNumber}/pre-prison-education':
```
james-baker-digital-justice, line 24: I'm not 100% sold on the name pre-prison-education although I can't quite put my finger on why it doesn't feel right. Perhaps because some of it may have been delivered on previous visits to prison? The terminology used in the new designs is "educational qualifications history" which could be shortened to education-history maybe?

It might just be an artefact of having "education qualifications" mastered by two different services when in reality if I'm preparing someone for work I only care if a person has a Maths GCSE or not and not whether it was earned in prison or not (although showing development in prison might also be a factor) 🤷‍♂️

james-baker-digital-justice (dismissed):
A couple of comments but nothing to halt the progress of the feature

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
nathanrussell-moj-digital: Yeah, that's a fair challenge 👍 
I've asked the question on the plp-dev-ba channel; hopefully Aliki or Ben will be able to provide some clarity on this 👍

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
nathanrussell-moj-digital: In respect of the generated code, I guess it's loosely inheritance, but without the formality of extending a class. Using your term, the generated types are "flattened" - eg:
```
package uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model

import java.util.Objects
import com.fasterxml.jackson.annotation.JsonProperty
    import jakarta.validation.constraints.*
    import jakarta.validation.Valid
    import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.resource.validator.*
    import io.swagger.v3.oas.annotations.media.Schema

        /**
 * Represents a Conversation with a prisoner within the context of their Learning and Work Progress.
 * @param reference A unique reference reference for this ConversationResponse.
 * @param prisonNumber The ID of the prisoner
 * @param note A note for this conversation
 * @param createdBy The DPS username of the person who created this resource.
 * @param createdByDisplayName The display name of the person who created this resource.
 * @param createdAt An ISO-8601 timestamp representing when this resource was created.
 * @param createdAtPrison The identifier of the prison that the prisoner was resident at when this resource was created.
 * @param updatedBy The DPS username of the person who last updated this resource.
 * @param updatedByDisplayName The display name of the person who last updated this resource.
 * @param updatedAt An ISO-8601 timestamp representing when this resource was last updated. This will be the same as the created date if it has not yet been updated.
 * @param updatedAtPrison The identifier of the prison that the prisoner was resident at when this resource was updated.
 */
data class ConversationResponse(

    @Schema(example = "814ade0a-a3b2-46a3-862f-79211ba13f7b", required = true, description = "A unique reference reference for this ConversationResponse.")
    @get:JsonProperty("reference", required = true) val reference: java.util.UUID,

    @Schema(example = "A1234BC", required = true, description = "The ID of the prisoner")
    @get:JsonProperty("prisonNumber", required = true) val prisonNumber: kotlin.String,

    @Schema(example = "Pay close attention to Peter's behaviour.", required = true, description = "A note for this conversation")
    @get:JsonProperty("note", required = true) val note: kotlin.String,

    @Schema(example = "asmith_gen", required = true, description = "The DPS username of the person who created this resource.")
    @get:JsonProperty("createdBy", required = true) val createdBy: kotlin.String,

    @Schema(example = "Alex Smith", required = true, description = "The display name of the person who created this resource.")
    @get:JsonProperty("createdByDisplayName", required = true) val createdByDisplayName: kotlin.String,

    @Schema(example = "2023-06-19T09:39:44Z", required = true, description = "An ISO-8601 timestamp representing when this resource was created.")
    @get:JsonProperty("createdAt", required = true) val createdAt: java.time.OffsetDateTime,

    @Schema(example = "BXI", required = true, description = "The identifier of the prison that the prisoner was resident at when this resource was created.")
    @get:JsonProperty("createdAtPrison", required = true) val createdAtPrison: kotlin.String,

    @Schema(example = "asmith_gen", required = true, description = "The DPS username of the person who last updated this resource.")
    @get:JsonProperty("updatedBy", required = true) val updatedBy: kotlin.String,

    @Schema(example = "Alex Smith", required = true, description = "The display name of the person who last updated this resource.")
    @get:JsonProperty("updatedByDisplayName", required = true) val updatedByDisplayName: kotlin.String,

    @Schema(example = "2023-06-19T09:39:44Z", required = true, description = "An ISO-8601 timestamp representing when this resource was last updated. This will be the same as the created date if it has not yet been updated.")
    @get:JsonProperty("updatedAt", required = true) val updatedAt: java.time.OffsetDateTime,

    @Schema(example = "BXI", required = true, description = "The identifier of the prison that the prisoner was resident at when this resource was updated.")
    @get:JsonProperty("updatedAtPrison", required = true) val updatedAtPrison: kotlin.String
) {

}
```

Whilst this is working for simple types like `ConversationResponse` it's not working in the array structure (which is where I really wanted it! 🤣 )
I need to play with a few options .... (and the discussion about API naming, and whether this API is just "pre-prison" or all quals will play into this)

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -1877,27 +1980,47 @@ components:
...
+            Note: This is a free format value and there is no type or enum. Therefore values can be "A", "B", "C" etc,
+            but also "1", "2", "3", "Pass", "Distinction", "Merit", "First class honours" etc. It is up to the consumer
+            to interpret this data as necessary.
+          example: Distinction
       required:
         - subject
         - level
         - grade
 
+    QualificationLevel:
```
nathanrussell-moj-digital, line 1999: I've moved this out of `AchievedQualification` and made it it's own enum type; to prevent it being duplicated in both `AchievedQualification`  and `AchievedQualificationResponse`

nathanrussell-moj-digital:
@james-baker-digital-justice re: the [slack conversation](https://mojdt.slack.com/archives/C05E3G0PU65/p1722843152434969) with the BAs where I was asking about naming, I had a call with Lydia and Ben about it where we now have an understanding that:

* Any qualifications whilst in prison are all recorded in Curious
* It is possible to do "school" qualifications in prison, typically Maths or English GCSEs. They will be recorded in Curious
* Anyone in Young Offenders who enter prison before they take they GCSEs in school will take them in prison. They will be recorded in Curious. (though those in YOI do not use the PLP induction process anyway)
* Any qualifications recorded in Curious are also sent to LRS
* LRS holds all* qualifications, including those recorded in Curious because they were taken in-prison. ("all" in this sense means those after a certain date, and also does not include foreign nationals)
* The reason prisons want to know about qualifications is to get an understanding of a prisoner's functional skills levels (typically just Maths and English)
  * If the prisoner has evidenced qualifications on Maths and/or English (either in Curious, LRS or PLP) then it is used to determine the prisoner's basic ability in Maths and English.
  * Prisons typically wont record (and dont care about) other qualifications. IE. they wont record a Pottery GCSE
  * The only non-Maths/English GCSEs that they _might_ record might be something like Metalwork or Woodwork where the prisoner wants to get on a related activity
  * Prisons may record higher level qualifications such as (evidenced) degrees, as obtaining a degree implies a sufficient grounding in Maths and English
* PLP shouldn't really be capturing/mastering qualifications, but there is nowhere else at the moment

Recording and seeing a prisoner's highest level of education and (relevant - see above) qualifications, along with qualifications and functional skills assessments pulled from Curious, and quals pulled from LRS (if we ever do that) allows the prison staff to get an understanding of the prisoner's Maths and English levels at a glance.

Anyway, based on the above, I've renamed the API path and related types, and have dropped the term "pre-prison". It is now simply "education" and I've documented the fact that it does not include quals etc achieved whilst in prison.

james-baker-digital-justice (approved):
LGTM - much better name

PR #330, opened 2024-08-06
RR-901 - Rename HighestEducationLevel to EducationLevel

nathanrussell-moj-digital:
This PR is an enabler for the work we are about to get into in respect of prisoner education and qualifications.

We already have some related functionality and code in the API, where part of it models a prisoner's education level. 
On screen this is captured as the prisoners Highest Level of Education, but this is very much a UI/presentation concern.
In respect of the API, domain and database it has been modelled within an enclosing type that models their education and qualifications, and there we (correctly) refer to it simply as `educationLevel`
However, some early development in this area named the enum as `HighestEducationLevel`
This is clearly wrong - the enum is an enumeration of education levels, so should be called `EducationLevel`

This PR renames the enums `HighestEducationLevel` to simply `EducationLevel` (there are 3 of them, one of which is generated from the swagger spec)


joelstobart-moj (approved):
LGTM

PR #331, opened 2024-08-07
RR-900 - Fix bug when updating induction with new qualifications

nathanrussell-moj-digital:
This PR fixes a bug when updating an induction that previous had no qualifications, but the update now contains qualifications.

Specifically the scenario is:
* An `Induction` exists, and when that induction was first created no `previousQualifications` (Highest Level of Education, or Qualifications) were added
(Highest Level of Education and Qualifications used to be entirely optional at the UI, so there will be existing `Induction` records in the database where the corresponding `PreviousQualificationsEntity` is null
* The `Induction` is subsequently updated in the UI by adding previous qualifications (either Highest Level of Education, or Qualifications)

This always used to work, but now that `PreviousQualificationsEntity` contains the `prisonNumber` field, we now have a bug in the above scenario because `prisonNumber` is not mapped.

All other scenarios involving `Induction`s (creating and updating) and `previousQualifications` (adding new or updating existing) work as the `prisonNumber` is either correctly mapped in the create flows, or already exists in the update flows.

This scenario is a bit of a weird one as its a combination of an update flow (updating the existing Induction) and a create flow (creating previousQualifications on the already existing Induction)

### DIFF
--- a/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/UpdateInductionTest.kt
+++ b/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/UpdateInductionTest.kt
@@ -22,6 +22,7 @@ import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.Error
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.HasWorkedBefore
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.TimelineEventType
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.assertThat
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.induction.aValidAchievedQualification
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.induction.aValidCreateInductionRequestForPrisonerLookingToWork
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.induction.aValidCreateInductionRequestForPrisonerNotLookingToWork
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.induction.aValidFutureWorkInterestsResponse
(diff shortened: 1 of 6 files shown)


src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/UpdateInductionTest.kt
```diff
@@ -448,6 +449,55 @@ class UpdateInductionTest : IntegrationTestBase() {
     }
   }
 
+  @Test
+  fun `should update induction for prisoner who has no qualifications to start with but wants to add qualifications`() {
```
nathanrussell-moj-digital, line 453: Integration test for this specific circumstance to prove the fix

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/JpaInductionPersistenceAdapter.kt
```diff
@@ -89,8 +90,16 @@ class JpaInductionPersistenceAdapter(
...
+        CreatePreviousQualificationsDto(
+          prisonNumber = prisonNumber,
+          educationLevel = educationLevel!!,
+          qualifications = qualifications,
+          prisonId = prisonId,
+        )
+      }
       return previousQualificationsRepository.saveAndFlush(
-        previousQualificationsMapper.fromUpdateDtoToNewEntity(updateInductionDto.previousQualifications)!!,
+        previousQualificationsMapper.fromCreateDtoToEntity(createPreviousQualificationsDto)!!,
```
nathanrussell-moj-digital, line 102: This is the core fix.
Previously we had a mapper that took a `UpdatePreviousQualificationsDto` instance and returned a new un-persisted `PreviousQualificationsEntity` instance.
At compile and unit test time this worked; but the resultant `PreviousQualificationsEntity` was missing `prisonNumber` because that field does not exist in `UpdatePreviousQualificationsDto` (it doest belong in `UpdatePreviousQualificationsDto` as it's not a property we use to reference/locate the right record, and its not a property we support updating)
At runtime this failed when JPA tried to persist the `PreviousQualificationsEntity` because `prisonNumber` is a non-nullable column.

Essentially the mapper is wrong, in that it doesn't make sense to create a new `PreviousQualificationsEntity` from an `UpdatePreviousQualificationsDto` (create a new thing from something that describes an update 🤔 👎 )

We already have a mapper that creates a new `PreviousQualificationsEntity` but from a `CreatePreviousQualificationsDto` which contains all of the required data, so my fix is to use that mapper instead, and to delete the other mapper entirely

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/actionplan/GoalEntityMapper.kt
```diff
@@ -36,13 +36,18 @@ abstract class GoalEntityMapper {
...
+   * `archiveReason` and `archiveReasonOther` are not mapped as they have no corresponding fields in the source `CreateGoalDto` instance
+   * because it is not possible to create a new goal that is already in an archived state.
    */
   @BeanMapping(qualifiedByName = ["addNewStepsDuringCreate"])
   @ExcludeJpaManagedFieldsIncludingDisplayNameFields
   @GenerateNewReference
   @Mapping(target = "createdAtPrison", source = "prisonId")
   @Mapping(target = "updatedAtPrison", source = "prisonId")
-  @Mapping(target = "steps", ignore = true)
+  @Mapping(target = "steps", ignore = true) // Steps are mapped in via the AfterMapping method addNewStepsDuringCreate
```
nathanrussell-moj-digital, line 48: Fixed some mapstruct warning whilst I was here about unmapped properties 👍 

(@james-baker-digital-justice , you'll be glad to know, I agree, there are more annotations than what would be lines of code if these were hand-written. I'm not aiming to do anything about this in this PR, but I have created a story for it - RR-917)

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/induction/PreviousQualificationsEntityMapper.kt
```diff
@@ -102,20 +103,6 @@ abstract class PreviousQualificationsEntityMapper {
     return existingQualifications
   }
 
-  @BeanMapping(qualifiedByName = ["addNewQualificationsDuringUpdate"])
```
nathanrussell-moj-digital, line 105: Two fewer mapstruct generated mapper functions 😝

james-baker-digital-justice (approved):
LGTM - still feels like it could be avoided by making the fields actually not-nullable types. I know you said it's something to do with the JPA mapping although in CRDS we had actual non-null types in the JPA entities so might be worth investigating at some point.

nathanrussell-moj-digital:
> LGTM - still feels like it could be avoided by making the fields actually not-nullable types. I know you said it's something to do with the JPA mapping although in CRDS we had actual non-null types in the JPA entities so might be worth investigating at some point.

Yes, agreed. If the JPA entity class had non-nullable fields then this would have been caught much earlier on (at original dev/compile time). I think we will should get to revisit this at some point 👍 🤞

PR #332, opened 2024-08-08
RR-920 - Basic setup to listen to HMPPS Domain Events

nathanrussell-moj-digital:
This PR adds the basic setup and config to consume and process HMPPS Domain Events, specifically  `prison-offender-events.prisoner.received` and `prison-offender-events.prisoner.released`

The actual logic for how we consume each of these events will be in subsequent PRs - this is just the basic infrastructure

The listener bean is disabled in `preprod` and `prod`, but is enabled in `dev`

There are unit tests in this PR. Whilst there are no integration tests yet, the integration test setup including circleci setup has been done.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,34 +23,23 @@ parameters:
     type: string
     default: -Xmx512m -XX:ParallelGCThreads=2 -XX:ConcGCThreads=2 -Djava.util.concurrent.ForkJoinPool.common.parallelism=2 -Dorg.gradle.daemon=false -Dkotlin.compiler.execution.strategy=in-process -Dorg.gradle.workers.max=1
 
-executors:
-  integration-test-environment:
-    docker:
-      - image: cimg/openjdk:<< pipeline.parameters.java-version >>
-      - image: cimg/postgres:<< pipeline.parameters.postgres-version >>
-        environment:
-          POSTGRES_PASSWORD: admin_password
-          POSTGRES_USER: admin
-          POSTGRES_DB: education_and_work_plan_api_db
-      - image: localstack/localstack:<< pipeline.parameters.localstack-version >>
-        environment:
-          SERVICES: sns,sqs
-          DATA_DIR: /tmp/localstack/data
-          DOCKER_HOST: unix:///var/run/docker.sock
-          AWS_EXECUTION_ENV: True
-          DEFAULT_REGION: eu-west-2
-          TMPDIR: /private
+jobs:
+  validate:
+    executor:
+      name: hmpps/java_localstack_postgres_with_db_name
+      jdk_tag: << pipeline.parameters.java-version >>
+      java_options: << pipeline.parameters.java-options >>
+      postgres_tag: << pipeline.parameters.postgres-version >>
+      postgres_db: "education_and_work_plan_api_db"
+      postgres_username: "admin"
+      postgres_password: "admin_password"
+      localstack_tag: << pipeline.parameters.localstack-version >>
+      services: "sqs,sns"
     environment:
-      _JAVA_OPTIONS: << pipeline.parameters.java-options >>
       DB_SERVER: localhost
       DB_NAME: education_and_work_plan_api_db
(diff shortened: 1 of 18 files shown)


.circleci/config.yml
```diff
@@ -16,21 +16,41 @@ parameters:
...
-jobs:
-  validate:
-    executor:
-      name: hmpps/java_postgres
-      jdk_tag: << pipeline.parameters.java-version >>
-      postgres_tag: << pipeline.parameters.postgres-version >>
-      postgres_db: "education_and_work_plan_api_db"
-      postgres_username: "admin"
-      postgres_password: "admin_password"
+executors:
```
nathanrussell-moj-digital, line 26: I couldnt find an hmpps orb that is java and postgres and localstack (there is one for java and postgres; and one for localstack; but not one for everything); and I couldn't workout how to configure the job with multiple orbs.
So I've gone with this approach

build.gradle.kts
```diff
@@ -5,7 +5,7 @@ import org.jlleitschuh.gradle.ktlint.tasks.KtLintFormatTask
 import org.openapitools.generator.gradle.plugin.tasks.GenerateTask
 
 plugins {
-  id("uk.gov.justice.hmpps.gradle-spring-boot") version "6.0.1"
+  id("uk.gov.justice.hmpps.gradle-spring-boot") version "6.0.2"
```
nathanrussell-moj-digital, line 8: Bumped a handful of versions whilst I was here, but the significant changes here are pulling in the hmpps sqs starter, and the localstack testcontainers dependencies.

helm_deploy/hmpps-education-and-work-plan-api/values.yaml
```diff
@@ -45,6 +46,14 @@ generic-service:
       DPR_USER: "DPR_USER"
       DPR_PASSWORD: "DPR_PASSWORD"
 
+    # Inbound SQS config
```
nathanrussell-moj-digital, line 49: These secrets are already setup in CPE and our namespaces

.circleci/config.yml
matt-crowson-moj: One should exist, see [here](https://github.com/ministryofjustice/hmpps-book-a-video-link-api/blob/main/.circleci/config.yml#L20) for an example.

nathanrussell-moj-digital: @matt-crowson-moj - Amazing; thank you; that looks to be just the ticket 👍 
I'll change it to use that from your example. Thanks

src/integrationTest/resources/application-integration-test.yml
```diff
@@ -17,6 +17,24 @@ spring:
         jwt:
           public-key-location: classpath:local-public-key.pub
 
+hmpps.sqs:
+  enabled: true
+  provider: localstack
+  queues:
+    educationandworkplan:
+      queueName: education-and-work-plan-queue
+      subscribeFilter: "{\"eventType\":[\"prison-offender-events.prisoner.received\", \"prison-offender-events.prisoner.released\"]}"
```
matt-crowson-moj, line 26: You may want to check you are listening to the right release event.  There are actually two (domain) events from different sources.

I think you probably want this want this `prisoner-offender-search.prisoner.released`???

The one you are using here is delayed in production (I think by approx 45 mins).  Perhaps double check with the Syscon team?

nathanrussell-moj-digital: Yeah, this came up in a slack conversation with Andy Lee and Andy Marke: https://mojdt.slack.com/archives/C69NWE339/p1722937869998369

They mentioned the delay factor as well.

As I understand the purpose of the delay, and our use case for want to know when people are released, I _think_ this is the right one to use. (We want to know when they really have been released, rather than potentially having to deal with a release then admission event in fairly quick succession.

Unless I've misunderstood what the delay is for? (it's quite possible! 🤣 )

matt-crowson-moj (approved):
LGTM

PR #335, opened 2024-08-13
RR-901 - Update swagger spec and introduce test fixtures

nathanrussell-moj-digital:
This PR is part of a larger piece of work (RR-901, specifically implementing the GET endpoint for returning a prisoner's education/qualifications outside of an induction). I've done the work, but it represents a large PR (40 files), so felt it would be easier for review purposes to break it up and do PRs for each logic section. 
This PR is the first in that process.

---

This PR updates the GET education endpoint definition in the swagger spec, and adds the text fixture classes we will use.

A previous PR had defined the GET endpoint, but on subsequent analysis we determined would could not fulfil all of the data items in the response. Specifically the "display name" of the user who created/updated the education record, and the prisonId where it was created/updated. We are not able to (easily and in a reasonable amount of dev time) provide these data items, so I have removed them from the `EducationResponse` type in the swagger spec.

The other part of this PR is the introduction of the test fixture classes for `EducationResponse`. A "test data builder" class to build instances for the purpose of tests, and an assertJ assertion class. Both will be used in subsequent PRs when we introduce the functionality that requires tests. This is just to get them in in the first place

### DIFF
--- a/src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
+++ b/src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   title: Education and Work Plan API
-  version: '1.15.1'
+  version: '1.15.2'
   description: Education and Work Plan API
   contact:
     name: Learning and Work Progress team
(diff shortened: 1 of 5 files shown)


src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -2016,10 +2016,39 @@ components:
     AchievedQualificationResponse:
       title: AchievedQualificationResponse
       type: object
-      description: A qualification that a Prisoner has achieved, including the audit fields of who created/updated it.
+      description: A qualification that a Prisoner has achieved, including the audit fields of who created/updated it and when.
       allOf:
         - $ref: '#/components/schemas/AchievedQualification'
-        - $ref: '#/components/schemas/AuditedResource'
```
nathanrussell-moj-digital, line 2022: We can no longer use `AuditedResource` for this type (`AchievedQualificationResponse`) because it contains the display name and prison ID fields, and these are the fields we cannot provide without a lot of rework.

Rather the extend `AuditedResource`, now we are simply defining the fields that we can provide

lynnette-orr (approved):

PR #336, opened 2024-08-13
RR-920 - Fix enums and message field cardinality

nathanrussell-moj-digital:
This PR adds some missing enum values and fixes field cardinality in respect of the HMPPS Domain Events that we listen to.

At the moment we are seeing a couple of problems in `prod` where the messages cannot be deserialized into our types. Most of these seem to be from the missing enum value `BEING_TRANSFERRED`, and some are based on the fact that we have the `details` field as non-nullable where in reality it can be null.

eg:
Screenshot 2024-08-13 at 16 38 57

The offender events swagger spec (where I was pointed at by the HMPPS dev community initially) is misleading / wrong, so I will separately raise a PR on that to correct it.
Screenshot 2024-08-13 at 16 40 40

Notice the missing enum value `BEING_TRANSFERRED`, the fact that `details` is incorrectly spelt as `detail` and none of the fields describe whether they are nullable or not.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/AdditionalInformation.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/AdditionalInformation.kt
@@ -15,9 +15,9 @@ sealed class AdditionalInformation {
   data class PrisonerReceivedAdditionalInformation(
     val nomsNumber: String,
     val reason: Reason,
-    val details: String,
-    val currentLocation: Location,
-    val currentPrisonStatus: PrisonStatus,
+    val details: String?,
+    val currentLocation: Location?,
+    val currentPrisonStatus: PrisonStatus?,
     val prisonId: String,
     val nomisMovementReasonCode: String,
   ) : AdditionalInformation() {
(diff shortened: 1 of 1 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/AdditionalInformation.kt
```diff
@@ -15,9 +15,9 @@ sealed class AdditionalInformation {
   data class PrisonerReceivedAdditionalInformation(
     val nomsNumber: String,
     val reason: Reason,
-    val details: String,
-    val currentLocation: Location,
-    val currentPrisonStatus: PrisonStatus,
+    val details: String?,
```
nathanrussell-moj-digital, line 18: By digging into the source code of `prisoner-offender-events` I found that these three properties are all nullable
https://github.com/ministryofjustice/prison-offender-events/blob/22dd9fdba858ae227d6facc19195d13921651792/src/main/kotlin/uk/gov/justice/hmpps/offenderevents/services/ReleasePrisonerReasonCalculator.kt#L77-L79

carloveo-moj (approved):
LGTM

PR #337, opened 2024-08-14
RR-901 - Move qualification related classes out of Induction domain into Education domain

nathanrussell-moj-digital:
Currently a prisoner's education and qualifications are modelled in our domain as being part of the Induction, because (up until now) it has always been understood that that is the relationship.

The business are now asking us for new functionality whereby a prisoner's education and qualification history can be entered and viewed outside of the context of an Induction. 

This PR refactors the classes related to education out of the `induction` domain package, and into a new package `education`.
Whilst it looks like a lot of change, the reality is that only a small number of files have moved , and the rest of the change is mostly import changes.


domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/education/EducationLevel.kt
```diff
@@ -0,0 +1,14 @@
+package uk.gov.justice.digital.hmpps.domain.learningandworkprogress.education
+
+/**
+ * An enumeration of Education Levels
```
nathanrussell-moj-digital, line 4: Appears to be a new enum, but has just been moved out of `PreviousQualifications.kt` into it's own source file.

domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/education/PreviousQualifications.kt
```diff
@@ -24,20 +26,15 @@ data class PreviousQualifications(
...
-  SECONDARY_SCHOOL_LEFT_BEFORE_TAKING_EXAMS,
-  SECONDARY_SCHOOL_TOOK_EXAMS,
-  FURTHER_EDUCATION_COLLEGE,
-  UNDERGRADUATE_DEGREE_AT_UNIVERSITY,
-  POSTGRADUATE_DEGREE_AT_UNIVERSITY,
-  NOT_SURE,
-}
-
 data class Qualification(
+  val reference: UUID?,
```
nathanrussell-moj-digital, line 30: As well as refactoring the domain, we do need these extra fields in the `Qualification` domain object

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/mapper/induction/QualificationsResourceMapper.kt
```diff
@@ -3,29 +3,80 @@ package uk.gov.justice.digital.hmpps.educationandworkplanapi.app.resource.mapper
...
+      educationLevel = request.educationLevel?.let { toEducationLevel(it) } ?: EducationLevelDomain.NOT_SURE,
+      qualifications = request.qualifications?.let { toQualifications(it) } ?: emptyList(),
+    )
 
   @Mapping(target = "updatedBy", source = "lastUpdatedBy")
   @Mapping(target = "updatedByDisplayName", source = "lastUpdatedByDisplayName")
   @Mapping(target = "updatedAt", source = "lastUpdatedAt")
   @Mapping(target = "updatedAtPrison", source = "lastUpdatedAtPrison")
-  fun toPreviousQualificationsResponse(previousQualifications: PreviousQualifications?): PreviousQualificationsResponse?
+  abstract fun toPreviousQualificationsResponse(previousQualifications: PreviousQualifications?): PreviousQualificationsResponse?
```
nathanrussell-moj-digital, line 39: Because `Qualification` now has some new fields this mapper class needs to change.
Currently we are using `mapstruct` to generate mapper implementations but there are some concerns over it's use, so we are exploring the approach of writing the mappers by hand.
The changes in this class represent a "half way house" , in that I'm only changing the methods in here that directly impact the changes we've made to `Qualification`

domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/education/PreviousQualifications.kt
```diff
@@ -24,20 +26,15 @@ data class PreviousQualifications(
...
   val subject: String,
   val level: QualificationLevel,
   val grade: String,
+  val createdBy: String?,
```
matt-crowson-moj, line 34: Obviously I don't know your code base so this may be standard throughout and a genuine reason.

Why would createdBy and createdAt ever be null?

nathanrussell-moj-digital: Good challenge 👍 
Yes, they could both be null.
We have a concept of a "domain" in our codebase, which sits between the REST API model classes, and the JPA entity classes.
When a new Qualification (or any other domain class) is created, it is not created until it has been persisted to the database; and we use the Spring annotations on the fields on the JPA entity classes to set the createdBy and createdAt values.

So, when a new Qualification is created, it comes in as a request in the REST API. That request object contains the data associated with a Qualification (grade, level, subject), but does not contain the DPS user or created time (ie. we dont expect the UI client to tell the API when something was created or by who)
The REST API request model object is mapped to this domain object (and therefore wont have createdBy and createdAt values, hence them being nullable)
The domain object is then mapped to the JPA entity which using the Spring annotations will set createdBy and createdAt

On the way back out, eg a REST API GET request, the JPA entity is mapped to the domain class, and this time it will have createdBy and createdAt populated. The domain class is then mapped to the REST API response model class where createdBy and createdAt are not nullable 👍

PR #340, opened 2024-08-16
RR-929 - Fix to allow specific endpoints to be called without a bearer token

nathanrussell-moj-digital:
This PR fixes a problem where some endpoints that are intended to be anonymous cannot be hit without a bearer token on the request (and are therefore not anonymous at all! 🙄 )

This has come about as a query from the SAR team because our service is not being correctly registered in the HMPPS Service Catalogue.

On digging into this, the Service Catalogue is maintained by some code that (amongst other things) hits the service's swagger UI url (to assert it has a swagger spec):
Screenshot 2024-08-16 at 14 52 51

We have a HTTP interceptor that extracts fields from the JWT bearer token to add to the App Insights context, and for URIs that we knew should be anonymous had set them up in the exclusion list. We had added `/swagger-ui/index.html`, but not `/swagger-ui.html`
Because `/swagger-ui.html` was not in the exclusion list, the interceptor was being invoked which resulted in a NPE because there was no bearer token on the request.

This PR fixes this by adding `/swagger-ui.html` to the exclusion (and also adding explanatory comments); and also adds some new integration tests to call the endpoints that we specifically exclude to assert they work without a bearer token.


src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/AnonymousAccessRequestTest.kt
```diff
@@ -0,0 +1,62 @@
...
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.ValueSource
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.IntegrationTestBase
+
+/**
+ * Integration tests that make HTTP requests against the API for resources that are designed to be accessed anonymously
+ * without a bearer token.
+ */
+class AnonymousAccessRequestTest : IntegrationTestBase() {
```
nathanrussell-moj-digital, line 12: This looks like a new test class, but is actually a rename of `GetSwaggerSpecTest` with a few extra non swagger related tests (hence the rename)

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/AnonymousAccessRequestTest.kt
```diff
@@ -0,0 +1,62 @@
...
+      .isOk
+  }
+
+  @ParameterizedTest
+  @ValueSource(
+    strings = [
+      "/queue-admin/retry-all-dlqs",
+    ],
+  )
+  fun `should make PUT request to HMPPS SQS queue housekeeping endpoints given request with no bearer token`(uri: String) {
```
matt-crowson-moj, line 55: Probably doesn't really need to be parameterized for one value? meh!

matt-crowson-moj (approved):
LGTM

PR #341, opened 2024-08-19
RR-901 - Return AchievedQualificationResponse as part of REST API InductionResponse

nathanrussell-moj-digital:
This PR changes the REST API `GET /inductions/{prisonNumber}` response so that it returns qualifications as collection of `AchievedQualificationResponse` rather that a collection of `AchievedQualification`

`AchievedQualificationResponse` as a type already exists and is the same as `AchievedQualification` except that it also contains the reference and audit fields (who/when created & updated). 

The reason we need this is part of supporting the changes we need to make for Qualifications and the change to the user journey. We need to show the audit fields (who/when created & updated) for each Qualification; and also we will need to be able to have better control over how we edit the prisoner's Qualifications, so having the reference for each Qualification will allow us to edit individual Qualifications.

For the UI this is a non breaking change as it is simply adding a few new fields to the response body object graph.
I will create a PR on the UI shortly to make use of these new fields.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/mapper/induction/QualificationsResourceMapper.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/mapper/induction/QualificationsResourceMapper.kt
@@ -3,12 +3,14 @@ package uk.gov.justice.digital.hmpps.educationandworkplanapi.app.resource.mapper
 import org.mapstruct.Mapper
 import org.mapstruct.Mapping
 import org.mapstruct.NullValueMappingStrategy
+import org.springframework.beans.factory.annotation.Autowired
 import uk.gov.justice.digital.hmpps.domain.learningandworkprogress.education.PreviousQualifications
 import uk.gov.justice.digital.hmpps.domain.learningandworkprogress.education.Qualification
 import uk.gov.justice.digital.hmpps.domain.learningandworkprogress.education.dto.CreatePreviousQualificationsDto
 import uk.gov.justice.digital.hmpps.domain.learningandworkprogress.education.dto.UpdatePreviousQualificationsDto
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.resource.mapper.InstantMapper
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.AchievedQualification
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.AchievedQualificationResponse
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.CreatePreviousQualificationsRequest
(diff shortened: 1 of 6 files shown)


src/testFixtures/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/resource/model/induction/PreviousQualificationsResponseBuilder.kt
```diff
@@ -1,17 +1,19 @@
...
 import java.time.OffsetDateTime
 import java.util.UUID
 
 fun aValidPreviousQualificationsResponse(
   reference: UUID = UUID.randomUUID(),
   educationLevel: EducationLevel = EducationLevel.SECONDARY_SCHOOL_TOOK_EXAMS,
-  qualifications: List<AchievedQualification> = listOf(
-    aValidAchievedQualification(),
-    anotherValidAchievedQualification(),
+  qualifications: List<AchievedQualificationResponse> = listOf(
```
nathanrussell-moj-digital, line 14: The type `PreviousQualificationsResponse` is now generated (generated from the swagger spec) with a `qualifications` property of a list of `AchievedQualificationResponse` rather than `AchievedQualification` as before

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -1049,7 +1049,7 @@ components:
           type: array
           description: A list of the Prisoner's previous qualifications. Can be empty but not null.
           items:
-            $ref: '#/components/schemas/AchievedQualification'
+            $ref: '#/components/schemas/AchievedQualificationResponse'
```
nathanrussell-moj-digital, line 1052: This change to the swagger spec is for the type `PreviousQualificationsResponse` and it's property `qualifications`. It is now a list of `AchievedQualificationResponse` rather than `AchievedQualification`.
`AchievedQualificationResponse` already exists as a type in this swagger spec and is basically the same as `AchievedQualification` but with additional reference and audit fields.

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/mapper/induction/QualificationsResourceMapper.kt
```diff
@@ -55,6 +61,21 @@ abstract class QualificationsResourceMapper {
   fun toQualifications(achievedQualifications: List<AchievedQualification>): List<Qualification> =
     achievedQualifications.map { toQualification(it) }
 
+  fun toAchievedQualificationResponse(qualification: Qualification): AchievedQualificationResponse =
```
nathanrussell-moj-digital, line 64: New mapper method to map from `Qualification` (the domain object) to `AchievedQualificationResponse` (the REST API response model object)

matt-crowson-moj (approved):
LGTM

PR #342, opened 2024-08-21
RR-901 - Pass optional reference in when creating or updating quals

nathanrussell-moj-digital:
This PR makes a change to the REST API spec to allow the reference number to be optionally passed in when creating qualifications as part of the Induction. This allows for qualifications to already exist (created by an earlier process by the Education team), and for them to be included in the Induction where they will be treated as existing Qualifications rather than created as new (and therefore duplicate) Qualifications.

(This PR is the API and basic mapping only, not the full implementation yet)

### DIFF
--- a/src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
+++ b/src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   title: Education and Work Plan API
-  version: '1.15.3'
+  version: '1.15.4'
   description: Education and Work Plan API
   contact:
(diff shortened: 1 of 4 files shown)


src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -974,7 +974,7 @@ components:
         reference:
           type: string
           format: uuid
-          description: A unique reference reference for this WorkOnReleaseResponse.
```
nathanrussell-moj-digital, line 977: Took the opportunity to correct the typo of double "reference" 😁

lynnette-orr (approved):

PR #343, opened 2024-08-23
RR-901 - Implementation to update qualifications identified by their reference number

nathanrussell-moj-digital:
This PR changes the API implementation re: creating and updating individual qualifications as part of the create or update Induction requests.

Previously when creating an Induction it was assumed there were no child elements of the induction already recorded in the system. (eg: pre-existing qualifications recorded before the Induction was created) This is still true because at time of writing it is not possible to create qualifications (or other resources) outside of the context of the parent Induction.
But we will soon be changing that, so that you will be able to create Qualifications before in the induction.

Therefore, to support that, the Create Induction implementation has been changed so that for any qualifications in the Create Induction request; if they have no reference number they are assumed to be new Qualifications to be recorded. For any that have a reference number they are assumed to either be an update to delete.

The same is now true of the Update Induction implementation.

This does not change the API (we've already done that in previous PRs) and is a non-breaking change to the UI. It simply changes the internal implementation


src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/UpdateInductionTest.kt
```diff
@@ -314,10 +316,6 @@ class UpdateInductionTest : IntegrationTestBase() {
       workOnRelease = aValidUpdateWorkOnReleaseRequestForPrisonerLookingToWork(
         reference = persistedInduction.workOnRelease.reference,
       ),
-      previousQualifications = aValidUpdatePreviousQualificationsRequest(
```
nathanrussell-moj-digital, line 317: The qualifications stuff in this test has been removed as the test itself isnt really about qualifications.
I've added a new test at the end that updates an induction with a specific focus on the various changes to qualifications that can be made in the update request 👍

lynnette-orr (approved):

PR #349, opened 2024-09-03
Small changes to make the build work

stevomcallister:
Couple of small changes - 

1. updated swagger so that the try it now should work with a valid JWT
2. updated docker compose so that the missing prison api url is there for dev.


nathanrussell-moj-digital (approved):
LGTM 👍

PR #350, opened 2024-09-04
RR-938 - Added maxLength properties to Induction fields

nathanrussell-moj-digital:
This PR updates the swagger spec to add `maxLength` properties to all the Induction fields that should have a length constraint.

This PR goes hand in hand with the [corresponding UI PR](https://github.com/ministryofjustice/hmpps-education-and-work-plan-ui/pull/764) that adds UI validation to the UI forms.
Basically it's belt'n'braces / doing the right thing. 

With this change the swagger spec now clearly communicates the field length constraints:
Screenshot 2024-09-04 at 10 51 26

and the generated model classes get the JSR annotation which will be validated by Spring, with the request being rejected with a 400 if the constraint is violated:
Screenshot 2024-09-04 at 10 51 46

### DIFF
--- a/src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
+++ b/src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   title: Education and Work Plan API
-  version: '1.15.4'
+  version: '1.16.0'
   description: Education and Work Plan API
   contact:
     name: Learning and Work Progress team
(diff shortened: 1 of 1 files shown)


src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   title: Education and Work Plan API
-  version: '1.15.4'
+  version: '1.16.0'
```
nathanrussell-moj-digital, line 4: Bump the version of the swagger spec (semver)
Technically this is a breaking change to the API (though it is actually fixing/improving things 😁 ), so this feels like the right version number change

stevomcallister (approved):

PR #352, opened 2024-09-06
refactored endpoints

stevomcallister:
API - Refactor archiveGoal, unarchiveGoal and getGoals controller & service code

Rewrote the above endpoints to follow the same pattern as other endpoints. Added a few additional exceptions and updated global exception handler. Removed redundant code and did some minor refactoring of GoalPersistenceAdaptor.


src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/ArchiveGoalTest.kt
```diff
@@ -178,7 +178,7 @@ class ArchiveGoalTest : IntegrationTestBase() {
   }
 
   @Test
-  fun `should return 400 if other reason without description`() {
+  fun `should return 409 if other reason without description`() {
```
nathanrussell-moj-digital, line 181: I'm not sure about this .... 🤔 
To me if feels like it should be a 400 BAD REQUEST, in that the user has tried to archive a goal with the reason of OTHER, but has not provided a reason. From our swagger spec:
Screenshot 2024-09-09 at 14 25 31

I'm not sure this feels like a 409 ? , but if it were we would want to update the swagger spec to describe this behaviour.

Keep it as a 400? What do you think?

domain/personallearningplan/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/personallearningplan/Exceptions.kt
```diff
@@ -27,3 +22,28 @@ class ActionPlanAlreadyExistsException(message: String) : RuntimeException(messa
  */
 class GoalNotFoundException(val prisonNumber: String, val goalReference: UUID) :
   RuntimeException("Goal with reference [$goalReference] for prisoner [$prisonNumber] not found")
+
+/**
+ * Thrown when a Goal cannot be arhived/unarchived due to the current state.
+ */
+class InvalidGoalStateException(val prisonNumber: String, val goalReference: UUID, val status: String, val action: String) :
```
nathanrussell-moj-digital, line 29: Can we make this a little more type safe perhaps?
Maybe make the `status` field `uk.gov.justice.digital.hmpps.domain.personallearningplan.GoalStatus` ?
And the `action` field - could we make that one of the values in the companion object? (would have to be an enum I guess)

domain/personallearningplan/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/personallearningplan/Exceptions.kt
```diff
@@ -27,3 +22,28 @@ class ActionPlanAlreadyExistsException(message: String) : RuntimeException(messa
...
+  RuntimeException("Could not $action goal with reference [$goalReference] for prisoner [$prisonNumber]: Goal was in state [$status] that can't be ${action}d") {
+
+  companion object {
+    const val ARCHIVE = "archive"
+    const val UNARCHIVE = "unarchive"
+  }
+}
+
+/**
+ * A general business exception that takes a string and will have a http code of 409 CONFLICT.
```
nathanrussell-moj-digital, line 39: Minor, but the way we've modelled and segregated the domain was quite deliberate and was to keep it "pure". So things like persistence and (in this case) transport concerns/terminology didn't end up "poluting" the domain.

(ie. our plan was that the domain modelled the business domain and it's behaviour, but with no regard to any given persistence or transport/api implementation. Then, at the application layer, we could catch/handle the exception and do something relevant to the transport/api . EG. in a spring-web environment we'd use a ControllerAdvice and handle it into some given response code/body. But we could also use the same domain for an SQS listener, or a console based app, or a quarkus app (if we ever moved away from spring) etc etc; and handle the exception to something relevant to the transport)

Saying `A general business exception that takes a string and will have a http code of 409 CONFLICT.` is not strictly true - the exception doesn't have the HTTP code, it's how it's handled downstream that will determine whether its a HTTP response code of some sort, or an error message in a console app, or a rejected and requeued message in an SQS app etc ...

domain/personallearningplan/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/personallearningplan/Exceptions.kt
```diff
@@ -27,3 +22,28 @@ class ActionPlanAlreadyExistsException(message: String) : RuntimeException(messa
...
+ */
+
+class BusinessException(val text: String) :
+  RuntimeException(text)
+
+/**
+ * A general not found exception that takes a string and will have a http code of 404 NOT_FOUND.
+ */
+class NotFoundException(val text: String) :
```
nathanrussell-moj-digital, line 48: Kinda related to my comment above, but we've tried to keep our exceptions quite specific, even if that means we end up with a few that ultimately mean a similar thing and are mapped to the same http response code.

`NotFoundException` feels a little vague a generic 🤔 Thinking about how you are currently using this new exception, what are your thoughts on it being called something like `PrisonerHasNoGoalsException` with a constructor arg for the `prisonNumber` rather than the reason text.
```
class PrisonerHasNoGoalsException(val prisonNumber: String) :
  RuntimeException("Prisoner $prisonNumber has no goals")

nathanrussell-moj-digital (commented):
A couple of comments that might be worth thinking about

Also, I wonder if this PR would also allow us to delete the (terribly named!) `ReturnAnErrorException` and associated code? 🤞

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/ArchiveGoalTest.kt
stevomcallister: 400 is fine by me - we are using 409s for the status update changes so I aligned them but really I think you're correct in that 400 is more appropriate.

nathanrussell-moj-digital (approved):
LGTM 👍

PR #360, opened 2024-09-16
Config change

stevomcallister:


nathanrussell-moj-digital (dismissed):
Happy to approve to keep this moving and fix the 500/404 issue 👍 

Not sure about re-using the prison-api client though ... I hear what you say about not needing one because all we need is a token that is issued/signed by hmpps-auth .... but I'm concerned that someone reading this code in 6 months or so will not find it clear etc
```
          manage-users-api:
            provider: hmpps-auth
            client-id: ${prison-api.client.id}
            client-secret: ${prison-api.client.secret}
            authorization-grant-type: client_credentials
            scope: read
```
IE. they might assume it's a typo or similar 🤔 

I think it's OK for now to move this ticket forwards, but I think we do want to get the HAAR team to create a manage-api client with the appropriate secrets in our namespaces eg: follow suit from this in our helm config
```
    hmpps-education-and-work-plan-api:
      APPINSIGHTS_INSTRUMENTATIONKEY: "APPINSIGHTS_INSTRUMENTATIONKEY"
      PRISON_API_CLIENT_ID: "PRISON_API_CLIENT_ID"
      PRISON_API_CLIENT_SECRET: "PRISON_API_CLIENT_SECRET"
```
Even if that client has no roles I think it feels cleaner / easier to understand; and it means that should one day someone decide that manage-api does need to use roles it will mean that we are not getting the role `ROLE_MANAGE_API__USER__RO` or similar added to our prison-api client (which really will make no sense 🤣 )

nathanrussell-moj-digital (approved):

PR #368, opened 2024-09-26
RR-981 - Added induction_schedule table and associated domain and persistence code

nathanrussell-moj-digital:
PR to add the induction_schedule table and associated domain and persistence code, which we'll be needing for KPI1

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/entity/induction/InductionScheduleEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/entity/induction/InductionScheduleEntity.kt
@@ -1,0 +1,103 @@
+package uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.induction
+
+import jakarta.persistence.Column
+import jakarta.persistence.Entity
+import jakarta.persistence.EntityListeners
+import jakarta.persistence.GeneratedValue
+import jakarta.persistence.Id
+import jakarta.persistence.Table
+import jakarta.validation.constraints.NotNull
+import org.hibernate.Hibernate
+import org.hibernate.annotations.CreationTimestamp
+import org.hibernate.annotations.UpdateTimestamp
+import org.hibernate.annotations.UuidGenerator
+import org.springframework.data.annotation.CreatedBy
+import org.springframework.data.annotation.LastModifiedBy
+import org.springframework.data.jpa.domain.support.AuditingEntityListener
+import java.time.Instant
+import java.time.LocalDate
+import java.util.UUID
+
+/**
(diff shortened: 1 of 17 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/entity/induction/InductionScheduleEntity.kt
```diff
@@ -0,0 +1,103 @@
...
+
+  @Column(updatable = false)
+  val scheduleCalculationRule: InductionScheduleCalculationRule,
+
+  @Column(updatable = true)
+  var deadlineDate: LocalDate,
+
+  @Column(updatable = true)
+  var scheduleStatus: InductionScheduleStatus,
+) {
```
nathanrussell-moj-digital, line 42: I'm trying something slightly different here than what we have currently with our other entity classes. This is based on comments from James Baker about our existing entity classes ...

The existing entity classes are all standard classes (non-data classes), and the fields are all constructor args .... but they are all nullable, even though as far as our application/domain etc were concerned they might have been mandatory fields.

The reason was based on (some probably old? 🤷‍♂️ ) kotlin "best practice" from JpaBuddy, which basically said to make the fields nullable and to make then all `var` (even if they were not updatable) so that JPA could instantiate instances with the default no-args constructor, and then it uses the setters or reflection to set the values.

James said you dont need to do that anymore, and that JPA will find the right constructor etc; and that it would be better to use kotlin as intended in respect of non-nullable and/or non-mutable fields.

So here's my stab at it, which did seem to work in my POC this morning 👍 

Using a `data` class, and all the fields in the constructor are the ones that are fundamentally part of the entity - ie. the "business" data of the entity. Fields that once set should not be mutated (such as `prisonNumber`) are `val`, and those that can be updated (eg: `deadlineDate`) are `var`

Then there are other fields that are not in the constructor. These are the JPA managed fields, such as the `id` and `createdBy` etc fields. These cannot be on the constructor because their values are not known at runtime when the instance is instantiated (ie. when creating a new one), and it would be weird for our application code to be providing these values.
These fields are all nullable, because as a kotlin class it is valid for these fields to be null at some point in time, but I'm using the `NotNull` annotation to give devs a clue they are not meant to be null.
And the fields are all `val` so that they can be set (by JPA) after instantiation via the setters

nathanrussell-moj-digital: If this works and we like the approach, we could look to migrate our other entity classes to this pattern over time.
If nothing else it will tidy our code elsewhere as currently we use the `!!` notation to tell the compiler that whilst the field is nullable we know better and it does have a value. We can tidy all that up 👍

stevomcallister (approved):
looks great ⭐️

PR #369, opened 2024-09-27
RR-983 - Added prisoner-search-api client class and supporting code

nathanrussell-moj-digital:
This PR adds a client for calling `prisoner-search-api`
Currently nothing uses it - this PR just introduces the class, the basic wiring, and the config.

As part of this I've asked the HAAR team to create new client creds for us (in all 3 envs) with the appropriate role (ROLE_PRISONER_SEARCH)
Whilst I was talking to HAAR I also took the opportunity to ask them to create client creds for manage-users-api 👍 
Using the dev client creds they created, I have manually tested using them against the dev APIs, all works fine 👍 

Also updated the readme in a number of places.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -221,7 +221,7 @@ tasks.register<GenerateTask>("buildPrisonApiModel") {
   )
   globalProperties.set(
     mapOf(
-      "models" to "",
+      "models" to "PrisonerInPrisonSummary,PrisonPeriod,SignificantMovement,TransferDetail",
     ),
   )
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/client/prisonersearch/PagedPrisonerResponse.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/client/prisonersearch/PagedPrisonerResponse.kt
@@ -1,0 +1,27 @@
+package uk.gov.justice.digital.hmpps.educationandworkplanapi.app.client.prisonersearch
+
+import java.time.LocalDate
+
+data class PagedPrisonerResponse(
+  val last: Boolean,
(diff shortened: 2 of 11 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/client/prisonersearch/PagedPrisonerResponse.kt
```diff
@@ -0,0 +1,27 @@
+package uk.gov.justice.digital.hmpps.educationandworkplanapi.app.client.prisonersearch
+
+import java.time.LocalDate
+
+data class PagedPrisonerResponse(
```
nathanrussell-moj-digital, line 5: There's some weirdness in the prisoner-search-api swagger spec which prevents us codegen'ing the models as we might otherwise do (some fields in the spec are marked as mandatory, but they are missing in the responses (in dev) so null pointers a-plenty!; plus there was some weird inheritance going on which made the codegen wobble!)

Anyway, given we are only interested in 3 classes, and of those a significant subset of the fields, I decided it was easiest just to create them here.

build.gradle.kts
```diff
@@ -221,7 +221,7 @@ tasks.register<GenerateTask>("buildPrisonApiModel") {
   )
   globalProperties.set(
     mapOf(
-      "models" to "",
+      "models" to "PrisonerInPrisonSummary,PrisonPeriod,SignificantMovement,TransferDetail",
```
nathanrussell-moj-digital, line 224: Not related to prisoner-search-api, but I decided to modify the codegen for prison-api so that it only generates the models that we are actually using (prison-api is massive and it was generating loads of models - I'm hoping this change might make a (small) improvement in build times? 🤞 )

stevomcallister (approved):

PR #375, opened 2024-09-30
drop note column

stevomcallister:


nathanrussell-moj-digital (commented):
Do we also need to get rid of the transient `notes` field on `GoalEntity`?
And the `notes` field on the domain `Goal` class?
🤔

stevomcallister:
> Do we also need to get rid of the transient `notes` field on `GoalEntity`? And the `notes` field on the domain `Goal` class? 🤔

Sadly not. Due to the way the the goals are created (from an action plan) in bulk as a list, we have no way of keeping track of the goal notes. If I was to rewrite the goal code to save them individually then I could remove from the entity etc. But as it stands this is the only way to do it.

nathanrussell-moj-digital:
> > Do we also need to get rid of the transient `notes` field on `GoalEntity`? And the `notes` field on the domain `Goal` class? 🤔
> 
> Sadly not. Due to the way the the goals are created (from an action plan) in bulk as a list, we have no way of keeping track of the goal notes. If I was to rewrite the goal code to save them individually then I could remove from the entity etc. But as it stands this is the only way to do it.

Sad times 😢 
OK, we might want to look at this again at some point as I think it makes for something fairly confusing and not massively obvious in respect of handing over to another team. I suspect as we do further work in this area, eg: supporting Notes for Archiving and Completing goals etc; then that work might help in some ways 🤔 🤞 

Let's merge it as is, which at least allows us to move forwards 👍

nathanrussell-moj-digital (approved):

PR #376, opened 2024-10-01
Fix for concurrent modification exeption

stevomcallister:
Issue is that when an action plan is created that has multiple goals it sometimes fails in dev with a concurrent modification exception. I've added a test that should replicate the issue but unfortunately has not failed for me. 

Perceived wisdom from the internet says to copy the list before doing anything with it - so that is my proposed fix. Just a shame I can't be more confident with a test that reproduces the problem.


nathanrussell-moj-digital (approved):
🤞

PR #382, opened 2024-10-04
archive note changes

stevomcallister:
Changes to Archive Goal such that there is an optional parameter of note - this will then create a Note linked to the goal and a note type of GOAL_ARCHIVAL. 

Reactivating will hard delete this note. 

Additional change to the getGoal endpoint to return this single note as an archiveNote String - although I think it would be better to return as a list of notes - together with note type... watch this space.

### DIFF
--- a/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/GetGoalTest.kt
+++ b/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/GetGoalTest.kt
@@ -2,6 +2,8 @@ package uk.gov.justice.digital.hmpps.educationandworkplanapi.app.resource
 
 import org.junit.jupiter.api.Test
 import org.springframework.http.HttpStatus
+import org.springframework.http.MediaType
+import org.springframework.test.web.reactive.server.WebTestClient
 import uk.gov.justice.digital.hmpps.domain.aValidPrisonNumber
 import uk.gov.justice.digital.hmpps.domain.aValidReference
 import uk.gov.justice.digital.hmpps.domain.anotherValidPrisonNumber
@@ -9,11 +11,16 @@ import uk.gov.justice.digital.hmpps.educationandworkplanapi.aValidTokenWithAutho
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.aValidTokenWithNoAuthorities
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.IntegrationTestBase
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.bearerToken
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.ArchiveGoalRequest
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.ErrorResponse
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.GoalResponse
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.ReasonToArchiveGoal
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.actionplan.aValidArchiveGoalRequest
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.actionplan.aValidCreateGoalRequest
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.actionplan.assertThat
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.assertThat
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.withBody
+import java.util.UUID
 
 class GetGoalTest : IntegrationTestBase() {
   companion object {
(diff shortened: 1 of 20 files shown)


nathanrussell-moj-digital:
> although I think it would be better to return as a list of notes - together with note type... watch this space.

I like that idea quite a lot 👍 
Though we'll need to be careful as it will be a breaking change to the API will will definitely impact our UI, and possibly (we'll need to check) DPS Prisoner Profile 🤔

nathanrussell-moj-digital:
Moving forwards we'll want to make sure we have suitable jira tickets, so that the branches and commit messages we create can reference the ticket number. 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/repository/NoteRepository.kt
```diff
@@ -1,16 +1,23 @@
...
   fun findAllByPrisonNumber(prisonNumber: String): List<NoteEntity>
 
-  fun findAllByEntityReferenceAndEntityType(entityReference: UUID, entityType: EntityType): List<NoteEntity>
+  fun findAllByEntityReferenceAndEntityTypeAndNoteType(entityReference: UUID, entityType: EntityType, noteType: NoteType): List<NoteEntity>
 
   fun findByReference(entityReference: UUID): NoteEntity
+
+  fun findAllByEntityReference(entityReference: UUID): NoteEntity
+
+  @Modifying
```
nathanrussell-moj-digital, line 21: Minor, but is `@Modifying` meant to be used with a `@Query` annotation?

From our old friend [bealdung](https://www.baeldung.com/spring-data-jpa-modifying-annotation):
```
We should note that executing a delete query with @Query works differently from Spring Data JPA’s deleteBy name-derived query methods. The latter first fetches the entities from the database and then deletes them one by one. This means that the life-cycle method @PreRemove will be called on those entities. However, with the former, a single query is executed against the database.
```

With that in mind is it worth adding the `@Query` annotation? It would certainly suggest a small performance improvement in that it will delete the record(s) in one operation, rather than n+1 ? 
(I know we are only deleting 1 note at a time by this method, so any gains will be negligible )

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -821,6 +821,10 @@ components:
         archiveReasonOther:
           type: string
           description: Describes the reason for archiving if 'OTHER' is selected, it is mandatory in this scenario.
+        archiveNote:
```
nathanrussell-moj-digital, line 824: We normally bump the version of the swagger spec when we make any changes (semver)
This is a non breaking change, so perhaps just a patch bump?
(line 4 of the swagger spec)

src/test/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/GoalControllerTest.kt
```diff
@@ -8,6 +8,7 @@ import org.mockito.kotlin.any
 import org.mockito.kotlin.given
 import org.mockito.kotlin.verify
 import uk.gov.justice.digital.hmpps.domain.aValidPrisonNumber
+import uk.gov.justice.digital.hmpps.domain.learningandworkprogress.note.service.NoteService
```
nathanrussell-moj-digital, line 11: I think we can (and should?) delete this unit test altogether ...

Its the only controller that has a unit test class; everything else is tested by the integration tests.

The original thinking was that the controllers were so lightweight and slim that unit tests offered no value (especially considering the integrations which are the real tests we are interested in)

Then we added that error handling stuff, where the service returned an object, and depending on that object the controller would either return the resource, or return an error response resource.

Given that we've since removed that and reverted back to "standard" exception throwing and a controller advice, I really cant see any gain in this unit test class?

What do you think?

src/testFixtures/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/resource/model/actionplan/GoalResponseAssert.kt
```diff
@@ -157,6 +157,16 @@ class GoalResponseAssert(actual: GoalResponse?) :
     return this
   }
 
+  fun hasArchiveNote(expected: String?): GoalResponseAssert {
```
nathanrussell-moj-digital, line 160: I love assertJ 😁 
Just sayin' 😁 🤣

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/GetGoalTest.kt
```diff
@@ -142,5 +149,71 @@ class GetGoalTest : IntegrationTestBase() {
...
+
+    // Then
+    val actual = response.responseBody.blockFirst()
+    assertThat(actual)
+      .hasReference(goalReference)
+    // archive note should not be null.
+    assertThat(actual).hasArchiveNote(archiveNote)
+  }
+
+  fun archiveAGoal(
```
nathanrussell-moj-digital, line 202: I think we already have this method in `IntegrationTestBase`, line 231 ?
(I was going to suggest putting it there, then other tests can make use of it in the same way you are here, but then I found we already have it 🤣 )

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/repository/NoteRepository.kt
stevomcallister: Force of habit putting @Modifiying on updates. I'll change it to a query though since that's what the Spring elders recommend.

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/GetGoalTest.kt
stevomcallister: gah - didn't see that although I did look.

stevomcallister:
> > although I think it would be better to return as a list of notes - together with note type... watch this space.
> 
> I like that idea quite a lot 👍 Though we'll need to be careful as it will be a breaking change to the API will will definitely impact our UI, and possibly (we'll need to check) DPS Prisoner Profile 🤔

I was still going to return the existing goal notes as the note field so that it didn't break anything - maybe with an @deprected flag.

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/GoalController.kt
```diff
@@ -59,7 +65,13 @@ class GoalController(
     @PathVariable @Pattern(regexp = PRISON_NUMBER_FORMAT) prisonNumber: String,
     @PathVariable goalReference: UUID,
   ): GoalResponse {
-    return goalResourceMapper.fromDomainToModel(goalService.getGoal(prisonNumber, goalReference))
+    val response = goalResourceMapper.fromDomainToModel(goalService.getGoal(prisonNumber, goalReference))
+    // Get the archive note and update the response if present
+    noteService.getNotes(response.goalReference, EntityType.GOAL, NoteType.GOAL_ARCHIVAL)
+      .firstOrNull()?.let { archiveNote ->
+        return response.copy(archiveNote = archiveNote.content)
+      }
```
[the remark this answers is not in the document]
nathanrussell-moj-digital: Yes, I think you are right, but we need to be careful not to make this a breaking change (or at least coordinate breaking changes with the UIs).
Lets do it like this for now, and create a plan/tickets etc to implement this change, both at the API and the UIs (our UI, and also possibly DPS Prisoner Profile)

nathanrussell-moj-digital:
> > > although I think it would be better to return as a list of notes - together with note type... watch this space.
> > 
> > 
> > I like that idea quite a lot 👍 Though we'll need to be careful as it will be a breaking change to the API will will definitely impact our UI, and possibly (we'll need to check) DPS Prisoner Profile 🤔
> 
> I was still going to return the existing goal notes as the note field so that it didn't break anything - maybe with an @deprected flag.

Thats a good idea 👍 
I'd suggest perhaps doing it in a different branch and PR though (all part of the same jira ticket though if you like) (I'm very much a fan of little and often / small incremental changes ; rather than overly large PRs if we can help it 😁 👍 )

src/test/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/GoalControllerTest.kt
stevomcallister: It does seem a bit pointless. I'm all for removing tests for the sake of tests.

PR #383, opened 2024-10-07
implementing archive notes

stevomcallister:
Changes to Archive Goal such that there is an optional parameter of note - this will then create a Note linked to the goal and a note type of GOAL_ARCHIVAL. 

Reactivating will hard delete this note.

New parameter of goalNotes on goal - with note content and note type. This will return all notes - including the goal note and completion note (not yet implemented) 

I have deprecated the single solitary note on goal, since this is still used in the Front end.


nathanrussell-moj-digital (approved):
LGTM 👍

PR #386, opened 2024-10-09
RR-998 - Added displayName fields to NotesResource and related mappers

nathanrussell-moj-digital:
This PR maps the createdByDisplayName and updatedByDisplayName fields in Note in the REST API responses to be mapped from the user service (ie. look them up from the API call)
We need to make this change on all REST API responses where the resource contains createdByDisplayName and updatedByDisplayName fields - this PR just does Notes

In order to do this change I have:
* Renamed Notes to NotesResponse (consistent with our other response model types)
* added the createdByDisplayName and updatedByDisplayName fields to NoteResponse
* added the other audit related fields to NoteResponse (not strictly necessary for this PR, but I was there anyway and it made sense to do them)
* created and used a new mapper to map NoteDto domain instances into NoteResource instances, including looking up the displayname fields from the user service
*

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/client/manageusers/ManageUsersApiClient.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/client/manageusers/ManageUsersApiClient.kt
@@ -7,14 +7,14 @@ import org.springframework.web.reactive.function.client.WebClientResponseExcepti
 
 @Component
 class ManageUsersApiClient(@Qualifier("manageUsersApiWebClient") private val manageUsersApiClient: WebClient) {
-  fun getUserDetails(username: String): UserDetailsDto? {
+  fun getUserDetails(username: String): UserDetailsDto {
     return try {
       manageUsersApiClient
         .get()
         .uri("/users/{username}", username)
         .retrieve()
         .bodyToMono(UserDetailsDto::class.java)
-        .block()
+        .block()!!
     } catch (e: WebClientResponseException.NotFound) {
       UserDetailsDto(username, false, "$username not found")
     } catch (e: Exception) {
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/note/NoteMapper.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/note/NoteMapper.kt
@@ -8,7 +8,6 @@ import uk.gov.justice.digital.hmpps.domain.learningandworkprogress.note.dto.Enti
 import uk.gov.justice.digital.hmpps.domain.learningandworkprogress.note.dto.NoteType as DomainNoteType
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.note.EntityType as EntityEntityType
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.note.NoteType as EntityNoteType
-import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.NoteType as ResourceNoteType
 
 object NoteMapper {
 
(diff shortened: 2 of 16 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/client/manageusers/ManageUsersApiClient.kt
```diff
@@ -7,14 +7,14 @@ import org.springframework.web.reactive.function.client.WebClientResponseExcepti
 
 @Component
 class ManageUsersApiClient(@Qualifier("manageUsersApiWebClient") private val manageUsersApiClient: WebClient) {
-  fun getUserDetails(username: String): UserDetailsDto? {
+  fun getUserDetails(username: String): UserDetailsDto {
```
nathanrussell-moj-digital, line 10: Minor change to the signature here as this function will either return a UserDetailsDto or throw a ManageUsersApiException
I dont think there is a case / code path where it can return null?

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/note/NoteMapper.kt
```diff
@@ -55,13 +54,6 @@ object NoteMapper {
       EntityNoteType.GOAL_COMPLETION -> DomainNoteType.GOAL_COMPLETION
     }
 
-  fun toResourceModel(noteType: DomainNoteType) =
```
nathanrussell-moj-digital, line 58: This method is now in my new mapper
To be fair this method probably shouldn't have been here (my bad, I should have seen it at PR time) because it is mapping a domain type into a REST resource type; whereas this mapper class is in the jpa package so should not be concerned with REST API stuff! 😁

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/GoalController.kt
```diff
@@ -133,16 +133,14 @@ class GoalController(
     @Valid @RequestBody archiveGoalRequest: UnarchiveGoalRequest,
     @PathVariable @Pattern(regexp = PRISON_NUMBER_FORMAT) prisonNumber: String,
     @PathVariable goalReference: UUID,
-  ): Goal {
```
nathanrussell-moj-digital, line 136: Minor tidying - giving the method is annotated with `@ResponseStatus(HttpStatus.NO_CONTENT)` and the swagger spec defines it as returning no body, the function shouldn't really return anything 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/mapper/note/NoteResourceMapper.kt
```diff
@@ -0,0 +1,38 @@
...
+      createdByDisplayName = userService.getUserDetails(note.createdBy!!).name,
+      createdAt = instantMapper.toOffsetDateTime(note.createdAt)!!,
+      createdAtPrison = note.createdAtPrison,
+      updatedBy = note.lastUpdatedBy!!,
+      updatedByDisplayName = userService.getUserDetails(note.lastUpdatedBy!!).name,
+      updatedAt = instantMapper.toOffsetDateTime(note.lastUpdatedAt)!!,
+      updatedAtPrison = note.lastUpdatedAtPrison,
+    )
+
+  private fun toNoteType(noteType: DomainNoteType): ApiNoteType =
```
nathanrussell-moj-digital, line 32: This is basically the function that was in the other mapper class above ☝️

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/ManageUserService.kt
```diff
@@ -2,11 +2,12 @@ package uk.gov.justice.digital.hmpps.educationandworkplanapi.app.service
...
 import org.springframework.stereotype.Component
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.client.manageusers.ManageUsersApiClient
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.client.manageusers.UserDetailsDto
 
 @Component
 class ManageUserService(
   private val manageUsersApiClient: ManageUsersApiClient,
 ) {
-  fun getUserDetails(username: String) =
+  fun getUserDetails(username: String): UserDetailsDto =
```
nathanrussell-moj-digital, line 11: I know with kotlin we dont need to do this type of thing and it will infer the types .... but sometimes I like to see it on the method sig so I know what it's returning (rather than clicking into the implementation etc to find out, which is what I had to do in this case)

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/GoalController.kt
stevomcallister: Should this be the same with Archive Goal?

stevomcallister (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/GoalController.kt
nathanrussell-moj-digital: Good shout; yes, I've just pushed a commit that does that too 👍

stevomcallister (approved):

PR #388, opened 2024-10-10
Delete goal note when update note is blank

stevomcallister:
When the the goal note is set to be a blank string or null and there is an existing goal note - then delete it.


nathanrussell-moj-digital (approved):
LGTM 👍

PR #397, opened 2024-10-18
RR-1011 - Mapped displayName fields in WorkOnReleaseResourceMapper via user lookup

nathanrussell-moj-digital:
This PR changes `WorkOnReleaseResourceMapper` (one of the mappers used by `InductionResourceMapper`) to lookup and map the displayName fields from the user service

### DIFF
--- a/src/integrationTest/resources/simulations/__files/user-asmith_gen.json
+++ b/src/integrationTest/resources/simulations/__files/user-asmith_gen.json
@@ -1,0 +1,10 @@
+{
+  "username": "asmith_gen",
+  "active": true,
+  "name": "Alex Smith",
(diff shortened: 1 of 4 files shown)


src/integrationTest/resources/simulations/__files/user-asmith_gen.json
```diff
@@ -0,0 +1,10 @@
+{
```
nathanrussell-moj-digital, line 1: Needed a new wiremock mapping for "Alex Smith" as one of the integration tests expects that value based on values in one of the test data builders (if I had changed the test data builder I likely would have broken other tests and ended up chasing my tail!)

PR #406, opened 2024-10-23
RR-1011 - Mapped displayName fields in QualificationsResourceMapper via user lookup

stevomcallister:


nathanrussell-moj-digital (commented):
Minor, but can we change the signature of the method that maps to the Response object so that it does not take or return a nullable:
```
fun toPreviousQualificationsResponse(previousQualifications: PreviousQualifications): PreviousQualificationsResponse =
```

And then we'll need to put the logic re: whether to map or not in the main InductionMapper (like you've done in the other's you've done so far)

That way we have a consistent place where the logic happens, and these mappers are consistent in respect of their method sigs 👍

PR #411, opened 2024-10-29
process Admission domain event

stevomcallister:


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/PrisonerReceivedIntoPrisonEventService.kt
```diff
@@ -2,30 +2,61 @@ package uk.gov.justice.digital.hmpps.educationandworkplanapi.app.messaging
...
+          inductionService.createInductionSchedule(
+            CreateInductionScheduleDto(
+              deadlineDate = calculateDeadlineDate(inboundEvent),
+              prisonNumber = inboundEvent.prisonNumber(),
+              scheduleCalculationRule = InductionScheduleCalculationRule.NEW_PRISON_ADMISSION,
+            ),
+          )
+        } catch (inductionAlreadyExistsException: InductionAlreadyExistsException) {
+          log.info { "An induction already exists for prisoner with id ${inboundEvent.prisonNumber()}" }
+          // update existing deadline date
```
nathanrussell-moj-digital, line 36: I think this comment and todo needs to live in the catch block for InductionScheduleAlreadyExistsException 🤔 

If a prisoner already has an Induction (ie. InductionAlreadyExistsException), it means they have already had an induction (perhaps via their previous sentence) and what we need to do in this case is to kick off the Review process by creating / updating a ReviewSchedule (which we dont have yet)

The other catch block is the one you want I think. If a prisoner is admitted to prison and they already have an InductionSchedule, then that is the case by which we need to update the deadline date 👍 

But thinking about it a bit more .... I think maybe the service method `inductionService.createInductionSchedule` might need some fettling with ... 🤔 

I think the various scenarios are for prisoner received event are:
* Prisoner does not have an Induction or an InductionSchedule
  * They are a brand new prisoner, never been seen before. We need to create an InductionSchedule record for them
* Prisoner already has an Induction (whether they have an InductionSchedule is largely irrelevant)
  * They are a re-admitting / re-offending prisoner, and from when they were in prison before they had a PLP Induction
  * We need to treat this as a Review process, and update (or create) the prisoner's ReviewSchedule record (doesn't exist yet, is loosely in scope of the work I am currently doing)
* Prisoner does not have an Induction, but does have an InductionSchedule
  * They are a re-admitting / re-offending prisoner, and had an InductionSchedule from when they were in prison previously, but their actual Induction was never done
  * We need to update the deadline date of their InductionSchedule

I _think_ they are the scenarios, and I think we might need to change the service method in some way to support this (might be as easy and changing the order in which the exceptions are thrown 🤔 )

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/PrisonerReceivedIntoPrisonEventService.kt
```diff
@@ -2,30 +2,61 @@ package uk.gov.justice.digital.hmpps.educationandworkplanapi.app.messaging
...
+          // TODO
+        } catch (inductionScheduleAlreadyExistsException: InductionScheduleAlreadyExistsException) {
+          log.info { "An induction schedule exists for prisoner with id ${inboundEvent.prisonNumber()}" }
+        }
       }
 
       TRANSFERRED -> {
         log.info { "Processing Prisoner Received Into Prison Event with reason ${additionalInformation.reason}" }
+        // Should already have an induction but Create induction or update existing deadline date if one exists
```
nathanrussell-moj-digital, line 45: I think the logic here might be identical to whatever we write for the Admission event above 🤔 
I _think_ in terms of whether they have an Induction and/or InductionSchedule, and what we need to do .... I think it's the same 🤔 
We should check that assumption though

nathanrussell-moj-digital:
The following enum (in the domain) might help with the various reasons why a prisoner's InductionSchedule deadline date might get re-calculated:
```
enum class InductionScheduleCalculationRule(val existingPrisonerWhenScheduleCreated: Boolean) {
  NEW_PRISON_ADMISSION(false),
  EXISTING_PRISONER_LESS_THAN_6_MONTHS_TO_SERVE(true),
  EXISTING_PRISONER_BETWEEN_6_AND_12_MONTHS_TO_SERVE(true),
  EXISTING_PRISONER_BETWEEN_12_AND_60_MONTHS_TO_SERVE(true),
  EXISTING_PRISONER_INDETERMINATE_SENTENCE(true),
  EXISTING_PRISONER_ON_REMAND(true),
  EXISTING_PRISONER_UN_SENTENCED(true),
}

```
The existing prisoner ones are basically where we need to treat it as a Review

PR #416, opened 2024-11-01
RR-1065 - REST API implementation

nathanrussell-moj-digital:
REST API implementation for the GET endpoint to get a prisoner's review schedule and completed reviews.


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/entity/induction/InductionScheduleEntity.kt
```diff
@@ -43,27 +42,22 @@ data class InductionScheduleEntity(
   @Id
   @GeneratedValue
   @UuidGenerator
-  @field:NotNull
```
nathanrussell-moj-digital, line 46: These not null annotations caused me grief in the integration test (for review schedules and completed reviews), but I've removed them from InductionSchedule here as well to prevent similar future issues when we get there

stevomcallister (approved):

PR #423, opened 2024-11-06
RR-1088 - Service class for looking up prisoners via the prisoner search api

nathanrussell-moj-digital:


stevomcallister (approved):
Looks good

PR #425, opened 2024-11-06
RR-1088 - Refactor NoteEntity as a data class, make content field non-nullable

nathanrussell-moj-digital:
This PR makes a minor refactoring to the `NoteEntity` class to make it a data class, and to it's `content` field to make it non-nullable (and the same change in the DTO, mappers etc)

I've needed to make this change because it was biting me a little in respect of the "Submit Review" API endpoint I am implementing because it needs to persist a NoteEntity as part of submitting the Review
(To be fair, it wasn't strictly necessary, I could have coded around it, but it's neater this way)

### DIFF
--- a/domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/note/dto/UpdateNoteDto.kt
+++ b/domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/note/dto/UpdateNoteDto.kt
@@ -4,6 +4,6 @@ import java.util.UUID
 
 data class UpdateNoteDto(
   val reference: UUID,
-  val content: String?,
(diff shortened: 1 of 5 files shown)


domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/note/dto/UpdateNoteDto.kt
```diff
@@ -4,6 +4,6 @@ import java.util.UUID
 
 data class UpdateNoteDto(
   val reference: UUID,
-  val content: String?,
+  val content: String,
```
nathanrussell-moj-digital, line 7: A note's content cannot be null - you shouldnt be able to update a note with null content 👍

stevomcallister (approved):

PR #426, opened 2024-11-06
Create outbound message on the creation/update of an induction schedule

stevomcallister:


src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/IntegrationTestBase.kt
```diff
@@ -149,8 +161,15 @@ abstract class IntegrationTestBase {
...
   }
-  val domainEventQueueDlqClient by lazy { domainEventQueue.sqsDlqClient }
   val domainEventQueueClient by lazy { domainEventQueue.sqsClient }
+  val domainEventQueueDlqClient by lazy { domainEventQueue.sqsDlqClient }
+
+  val inductionScheduleEventQueue by lazy {
+    hmppsQueueService.findByQueueId("inductionscheduleeventqueue")
+      ?: throw MissingQueueException("HmppsQueue testeventqueue not found")
+  }
+  val testInductionScheduleEventQueueClient by lazy { inductionScheduleEventQueue.sqsClient }
```
stevomcallister, line 171: in order to test that the outbound message was put on the topic we need to create a queue that we can then read the message back from - this is her.

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/IntegrationTestBase.kt
```diff
@@ -325,4 +353,37 @@ abstract class IntegrationTestBase {
       .expectStatus()
       .isCreated
   }
+
```
stevomcallister, line 356: some helper methods for reading from the inductionSchedule queue

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/InductionScheduleUpdateEventPusher.kt
```diff
@@ -0,0 +1,65 @@
...
+import uk.gov.justice.hmpps.sqs.HmppsQueueService
+import uk.gov.justice.hmpps.sqs.HmppsTopic
+import uk.gov.justice.hmpps.sqs.publish
+import java.net.URI
+import java.time.Instant
+import java.time.LocalDateTime
+import java.time.ZoneId
+
+@Component
+class InductionScheduleUpdateEventPusher(
```
stevomcallister, line 18: I would like to make this more generic in a future iteration - so that we can send review messages via the same mechanism but taking baby steps for now.

nathanrussell-moj-digital: Sounds like a plan - just a minor question on the class name though - "Pusher" ? or "Publisher" ?

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/InductionScheduleUpdateEventPusher.kt
```diff
@@ -0,0 +1,65 @@
...
+  private val hmppsQueueService: HmppsQueueService,
+  private val objectMapper: ObjectMapper,
+  @Value("\${service.base-url}") private val serviceBaseUrl: String,
+) {
+  companion object {
+    val log: Logger = LoggerFactory.getLogger(this::class.java)
```
nathanrussell-moj-digital, line 24: Minor, but we use `KotlinLogging` to get our logger instance, and we do it outside of the class definition:
```
package uk.gov.justice.digital.hmpps.educationandworkplanapi.app.messaging

import mu.KotlinLogging
import org.springframework.stereotype.Service
import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.messaging.AdditionalInformation.PrisonerReleasedAdditionalInformation

private val log = KotlinLogging.logger {}

@Service
class PrisonerReleasedFromPrisonEventService {
```

Why? I hear you ask 🤣 
I believe its a) more kotlin specific, and the various log methods can take lambas etc
and b) (probably more important) is the fully qualified class name that the logs appear under.
With your logger in the companion object, I believe your log statements will be for:
```
uk.gov.justice.digital.hmpps.educationandworkplanapi.app.messaging.InductionScheduleUpdateEventPusher$Companion
```
IE. it has `$Companion` suffixed at the end because the JVM doesnt know what a companion object is and sees it as a nested class (I think)

( We use a slightly outdated version, where the gradle dependency is from `io.github.microutils` - the newer version is from `io.github.oshai`, but it's the same thing
https://github.com/oshai/kotlin-logging/wiki )

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/InductionScheduleUpdateEventPusher.kt
```diff
@@ -0,0 +1,65 @@
...
+    const val EVENT_TYPE = "plp.induction-schedule.updated"
+    const val CONTENT_TYPE = "text/plain;charset=UTF-8"
+  }
+
+  internal val eventTopic by lazy { hmppsQueueService.findByTopicId("domainevents") as HmppsTopic }
+
+  fun sendEvent(prisonerNumber: String, occurredAt: Instant) {
+    val cne = HmppsDomainEvent(
```
nathanrussell-moj-digital, line 32: `cne` ? - could we call it `domainEvent` or similar?

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/InductionScheduleUpdateEventPusher.kt
```diff
@@ -0,0 +1,65 @@
...
+      serviceBaseUrl,
+      occurredAt
+        .atZone(ZoneId.of("Europe/London")).toLocalDateTime(),
+    )
+    log.info("Pushing induction schedule updated event to event topic person reference $prisonerNumber")
+    val publishResponse = eventTopic.publish(
+      cne.eventType,
+      objectMapper.writeValueAsString(cne),
+      attributes = mapOf(
+        "eventType" to MessageAttributeValue.builder().dataType("String").stringValue(cne.eventType).build(),
```
nathanrussell-moj-digital, line 43: I could be wrong, but I vaguely recall some chatter about an update to the HMPPS SQS Spring Boot Starter which abstracted the AWS-ness away from client code like ours

I _think_ they implemented builders or similar that handle the specific impl classes of things like `MessageAttributeValue` etc.

We are on the latest version of that dependency, so would be worth looking to see if they have added something similar like that, that can make our code a little neater etc

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/InductionScheduleUpdateEventPusher.kt
```diff
@@ -0,0 +1,65 @@
...
+          .stringValue(CONTENT_TYPE).build(),
+      ),
+    )
+    log.debug("Sent case induction schedule updated with message id {}", publishResponse.messageId())
+  }
+
+  data class HmppsDomainEvent(
+    val version: Int = 1,
+    val eventType: String = EVENT_TYPE,
+    val description: String = "A prisoner learning plan induction schedule created or amended",
```
nathanrussell-moj-digital, line 54: I think this is OK for the moment, but I suspect the class will get refactored as and when we need to start sending similar events for review schedule changes etc (eg. the description you have here wont work for review schedule updates)
But it's cool for now 👍

(you mentioned above about making this class more generic, and I think this will fall into that thinking as and when 👍 )

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/InductionScheduleUpdateEventPusher.kt
stevomcallister: ha! a lot of this has been cribbed from other projects (cne = case note event Ooops!)

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/InductionScheduleUpdateEventPusher.kt
stevomcallister: yeah copy paste I'll sort it

nathanrussell-moj-digital (approved):
LGTM 👍

nathanrussell-moj-digital (approved):

PR #431, opened 2024-11-12
RR-1088 - Persistence adapter etc for persisting a new Completed Review

nathanrussell-moj-digital:
Once this PR is approved and merged I should be able to create a service method to create a completed review, whose implementation will use all the stuff that these recent PRs have been adding (ie. it will need to check a ReviewSchedule exists, create the new Completed Review, call the methods to work out the calculation rule and next schedule, and then update the ReviewSchedule with the next schedule etc.
And once we've got the service method in place then it will be an easy job to create the controller method which will call it 👍

### DIFF
--- a/domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/review/SentenceType.kt
+++ b/domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/review/SentenceType.kt
@@ -1,0 +1,14 @@
+package uk.gov.justice.digital.hmpps.domain.learningandworkprogress.review
+
+enum class SentenceType {
+  RECALL,
+  DEAD,
+  INDETERMINATE_SENTENCE,
+  SENTENCED,
+  CONVICTED_UNSENTENCED,
+  CIVIL_PRISONER,
+  IMMIGRATION_DETAINEE,
+  REMAND,
+  UNKNOWN,
+  OTHER,
+}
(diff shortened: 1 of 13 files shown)


domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/review/SentenceType.kt
```diff
@@ -0,0 +1,14 @@
+package uk.gov.justice.digital.hmpps.domain.learningandworkprogress.review
+
+enum class SentenceType {
```
nathanrussell-moj-digital, line 3: I've moved this out of `ReviewService` (below) into a source file of it's own
(still not sure about the name, but moving it out of ReviewService feels like a step in the right direction)

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/note/NoteMapper.kt
```diff
@@ -26,17 +26,17 @@ object NoteMapper {
...
       lastUpdatedBy = noteEntity.updatedBy,
       lastUpdatedAt = noteEntity.updatedAt,
-      lastUpdatedAtPrison = noteEntity.updatedAtPrison.orEmpty(),
-      noteType = toNoteType(noteEntity.noteType!!),
-      entityType = toEntityType(noteEntity.entityType!!),
-      entityReference = noteEntity.reference!!,
+      lastUpdatedAtPrison = noteEntity.updatedAtPrison,
+      noteType = toNoteType(noteEntity.noteType),
+      entityType = toEntityType(noteEntity.entityType),
+      entityReference = noteEntity.entityReference,
```
nathanrussell-moj-digital, line 39: As well as the changes for the non-nullable fields (which were made non-nullable in a PR a couple of days ago), there was a bug on this line re: the mapping of entityReference. The unit tests in `ReviewEntityMapperTest` below highlighted it 👍

stevomcallister (approved):

PR #436, opened 2024-11-12
allow all on dev

stevomcallister:
Disable the ingress allowlist for the API **on `dev` only**
This is to allow the Integration API to consume the API's swagger spec

As per the following conversation with Matt Ryall this is not ideal but we dont have other options in the short term:
image


nathanrussell-moj-digital (approved):
Approved on the basis that we don't have other options at the moment and time is not our friend right now either.

I suspect we might get challenged on this one day so I've updated the description to include the conversation/justification from Matt Ryall as rationale/evidence of this change

PR #437, opened 2024-11-14
RR-1088 - Changes to support multiple review schedules

nathanrussell-moj-digital:
PR to allow multiple review schedules for each prisoner, thus allowing a history of review schedules

The Review Schedule now has a new status `COMPLETED`, and the Review itself now has a new field which is the reference of the corresponding review schedule.


domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/review/Exceptions.kt
```diff
@@ -7,7 +7,7 @@ class ReviewScheduleNotFoundException(val prisonNumber: String) :
   RuntimeException("Review Schedule not found for prisoner [$prisonNumber]")
 
 /**
- * Thrown when a Review Schedule already exists for a given Prisoner.
+ * Thrown when a prisoner already has an active Review Schedule. A Prisoner cannot have more than 1 active Review Schedules.
  */
-class ReviewScheduleAlreadyExistsException(val prisonNumber: String) :
-  RuntimeException("Review Schedule already exists for prisoner [$prisonNumber]")
+class ActiveReviewScheduleAlreadyExistsException(val prisonNumber: String) :
```
nathanrussell-moj-digital, line 12: Now that a prisoner can have many ReviewSchedules, we now have the concept of one (and only one) of them being the active one. You'll see this in the code below, but the "active" Review Schedule is considered the one that is not COMPLETED.
IE. a prisoner can have many COMPLETED Review Schedules, but only one that is not COMPLETED. This one record can be SCHEDULED or one of the EXEMPT_ statuses

domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/review/service/ReviewPersistenceAdapter.kt
```diff
@@ -12,7 +12,7 @@ interface ReviewPersistenceAdapter {
   fun getCompletedReviews(prisonNumber: String): List<CompletedReview>
 
   /**
-   * Creates and returns a new [CompletedReview] using the specified [CreateCompletedReviewDto] and deadlineDate.
+   * Creates and returns a new [CompletedReview] using the specified [CreateCompletedReviewDto] and [ReviewSchedule].
    */
-  fun createCompletedReview(createCompletedReviewDto: CreateCompletedReviewDto, deadlineDate: LocalDate): CompletedReview
+  fun createCompletedReview(createCompletedReviewDto: CreateCompletedReviewDto, reviewSchedule: ReviewSchedule): CompletedReview
```
nathanrussell-moj-digital, line 17: The impl now needs the ReviewSchedule when creating a Completed Review, because we now need the reference of the schedule (as well as the deadline date, which we can get from the review schedule)

domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/review/service/ReviewSchedulePersistenceAdapter.kt
```diff
@@ -20,7 +20,23 @@ interface ReviewSchedulePersistenceAdapter {
...
+   */
+  fun getActiveReviewSchedule(prisonNumber: String): ReviewSchedule?
+
+  /**
+   * Retrieves a Prisoner's latest [ReviewSchedule]. The latest (most recently updated) [ReviewSchedule] is returned
+   * irrespective of status.
+   * Returns `null` if no [ReviewSchedule]s exist.
    */
-  fun getReviewSchedule(prisonNumber: String): ReviewSchedule?
+  fun getLatestReviewSchedule(prisonNumber: String): ReviewSchedule?
```
nathanrussell-moj-digital, line 41: We need to be able to get both the active schedule, but also the latest schedule.

The latest schedule might be the last COMPLETED schedule (in the case of the prisoner having had their last review), and we'll need that to be able to return it through the GET endpoint of the REST API.

src/main/resources/db/migration/common/V2024.11.13.0001__correct_indexes_on_review_and_review_schedule_tables.sql
```diff
@@ -0,0 +1,24 @@
+--- review table indexes
```
nathanrussell-moj-digital, line 1: Hurrah! We are back to normal flyway script number sequences 👍 😁

src/main/resources/db/migration/common/V2024.11.13.0001__correct_indexes_on_review_and_review_schedule_tables.sql
```diff
@@ -0,0 +1,24 @@
...
+DROP INDEX review_prison_number_idx;
```
nathanrussell-moj-digital, line 2: The 2 indexes were previously created incorrectly in `V2024.11.09.0009__add_review_and_review_schedule_tables`:
```
CREATE UNIQUE INDEX review_prison_number_idx ON review
(
    prison_number
);

...

CREATE UNIQUE INDEX review_schedule_prison_number_idx ON review
(
    prison_number
);
```
In that there was a unique index on `review.prison_number` which would have meant a prisoner could only have 1 review
And the unique index that was supposed to be on `review_schedule.prison_number` was actually also on `review` !!
So neither indexes did what they were supposed to! 🤦‍♂️

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/JpaReviewSchedulePersistenceAdapter.kt
```diff
@@ -40,8 +41,21 @@ class JpaReviewSchedulePersistenceAdapter(
   }
 
   @Transactional(readOnly = true)
-  override fun getReviewSchedule(prisonNumber: String): ReviewSchedule? =
-    reviewScheduleRepository.findByPrisonNumber(prisonNumber)?.let {
-      reviewScheduleEntityMapper.fromEntityToDomain(it)
+  override fun getActiveReviewSchedule(prisonNumber: String): ReviewSchedule? =
+    with(
```
stevomcallister, line 45: Lets say a prisoner has been in prison for years. then this will bring back lots of records and then filter on the active one. 

Probably better to create a HQL query that simply returns a single result using a where status != ReviewScheduleStatus.COMPLETED which would return a NonUniqueResultException for the case where the database has got dodgy data and do the filtering in the DB which will perform better

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/JpaReviewSchedulePersistenceAdapter.kt
```diff
@@ -40,8 +41,21 @@ class JpaReviewSchedulePersistenceAdapter(
...
+      if (size > 1) {
+        throw IllegalStateException("A prisoner cannot have more than one active ReviewSchedule. Please investigate the ReviewSchedule data for prisoner $prisonNumber")
+      }
+      firstOrNull()
+        ?.let { reviewScheduleEntityMapper.fromEntityToDomain(it) }
     }
+
+  @Transactional(readOnly = true)
+  override fun getLatestReviewSchedule(prisonNumber: String): ReviewSchedule? =
+    reviewScheduleRepository.getAllByPrisonNumber(prisonNumber)
```
stevomcallister, line 58: Maybe do this in the database rather than the code?

nathanrussell-moj-digital:
Yes, I like the idea of doing that filtering at the DB query, rather than in code .... let me go and have a tinker ... 😁

stevomcallister (approved):

PR #438, opened 2024-11-14
return system as the display name

stevomcallister:
instead of system not found since system doesn't have a user account.


nathanrussell-moj-digital (approved):
LGTM 👍

PR #440, opened 2024-11-15
RR-1088 - Implement "7 days until release" rule for determining the Review Schedule Calculation Rule

nathanrussell-moj-digital:
This PR adds an extra `ReviewScheduleCalculationRule` and associated logic to cover an edge case rule that is documented in the signed off KPI2 rules.

The rules of interest are:
* If the prisoner has exactly 3 months or less left to serve, we do not calculate a new Review Schedule for them. They have had their last review already.
  * This has already been implemented ✅ 
* If the prisoner has over 3 months and 6 months or less left to serve, we calculate their new review schedule as being between 1 and 3 months away. IE. the deadline date is 3 months away.
  * This has already been implemented ✅ 

There is a rule that says prisoners cannot have a review in their final 7 days of prison, which leads us to this edge case ....

If a prisoner has 3 months and 1 day left to serve, they will be in the "3 to 6 months left" bracket, and will will schedule their next review as being between 1 and 3 months away. It means the deadline date is the day before they are released.
This problem is true for 3 months + 1 day, 3 months + 2 days, 3 months + 3 days .... up to 3 months + 7 days.

If the prisoner has 3 months and 8 days left to serve then they problem does not exist as we will schedule their review for between 1 and 3 months, which means the deadline date is 3 months away, which will be 8 days before the prisoner is released, and therefore does not foul the "cannot have a review in the final 7 days" rule.

This PR implements a new `ReviewScheduleCalculationRule` snappily called `BETWEEN_3_MONTHS_AND_3_MONTHS_7_DAYS_TO_SERVE`; and associated logic to calculate the correct Review Schedule Window.


domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/review/ReviewSchedule.kt
```diff
@@ -25,6 +25,7 @@ enum class ReviewScheduleCalculationRule {
   PRISONER_READMISSION,
   PRISONER_TRANSFER,
   BETWEEN_RELEASE_AND_3_MONTHS_TO_SERVE,
+  BETWEEN_3_MONTHS_AND_3_MONTHS_7_DAYS_TO_SERVE,
   BETWEEN_3_AND_6_MONTHS_TO_SERVE,
```
nathanrussell-moj-digital, line 29: Arguably this enum value `BETWEEN_3_AND_6_MONTHS_TO_SERVE` is now incorrect.
The pendant in me is saying it should really be called `BETWEEN_3_MONTHS_8_DAYS_AND_6_MONTHS_TO_SERVE` .... but I really don't know if I can be bothered / know whether it's worth it.

If I were to rename it I would need to rename it here in the domain, in the corresponding entity enum, and in the swagger spec for the resultant REST API enum.

The value is rendered out in one of the REST API responses. Our UI will be consuming that API, but I dont think the verboseness of this value will be an issue to us as this level of detail belongs in our application anyway (and chances are our UI will probably map it to some text that makes it clear anyway)
But (via the integration api), MN will be consuming an API that might contains this value. (we haven't designed it yet). And I don't know if renaming it will make it too verbose / implementation specific for MN (or even if they care about it? 🤔 )

What do you think?

nathanrussell-moj-digital: If we were going to rename it, now would be the time to do it ....

domain/learningandworkprogress/src/test/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/review/service/ReviewServiceTest.kt
```diff
@@ -118,90 +114,4 @@ class ReviewServiceTest {
       verify(reviewPersistenceAdapter).getCompletedReviews(PRISON_NUMBER)
     }
   }
-
-  @Nested
```
nathanrussell-moj-digital, line 122: We dont need these here as they are all covered in `ReviewServiceCreateReviewTest` and `ReviewServiceCalculateReviewWindowTest`

domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/review/ReviewSchedule.kt
stevomcallister: It might be pedantic but it describes the situation and shows that we've considered the situation. So I think the enum should be BETWEEN_3_MONTHS_8_DAYS_AND_6_MONTHS_TO_SERVE even though it is a whopper! 

I don't think anyone will care though so it's not urgent.

stevomcallister (dismissed):

domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/review/ReviewSchedule.kt
nathanrussell-moj-digital: I know it's not urgent, but I think if the pedant in both of us is calling us to do it, then I'm going to do the rename now 😁

stevomcallister (approved):

PR #448, opened 2024-11-22
create a review schedule when an induction already exists

stevomcallister:


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/ciagkpi/PefCiagKpiService.kt
```diff
@@ -31,39 +37,76 @@ class PefCiagKpiService(
...
+  /**
+   * Process an prisoner admission.
+   *
+   * - If the prisoner is brand new create a new Induction schedule
+   * - If the prisoner already has an incomplete induction schedule (this will most likely be
+   * a duplicate message) then do nothing.
+   * - If the prisoner already has had their Induction created. Then this person is likely to
+   * be a coming back into prison and needs to resume their Reviews.
+   */
   override fun processPrisonerAdmission(prisonNumber: String, prisonAdmittedTo: String, eventDate: Instant) {
```
stevomcallister, line 53: The changes in here look scarier than they are. Basically split this function in to more discreet functions. Also an explanation of whats happening in the java doc (kotlin doc sounds weird)

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/ciagkpi/PefCiagKpiService.kt
```diff
@@ -31,39 +37,76 @@ class PefCiagKpiService(
...
     log.info { "Creating or updating induction schedule for prisoner [$prisonNumber]" }
 
-    // Check if an induction schedule already exists.
-    val existingSchedule = inductionSchedulePersistenceAdapter.getInductionSchedule(prisonNumber)
-    if (existingSchedule != null) {
-      log.info { "Induction schedule already exists for prisoner [$prisonNumber], ignoring this message." }
-      return
-    }
+    if (activeInductionScheduleAlreadyExists(prisonNumber)) return
```
stevomcallister, line 56: This was previously if any induction schedule exists then ignore the message. Don't think that was correct there will always be an induction schedule but we only want to return when there is an active induction schedule. 

Although even that may not be correct, I'm kind of assuming this will only happen when there is a duplicate admission message and we would want to drop the message in that case but... what if the induction was not complete before the prisoner was released... Could feasibly happen. Then on readmission months or years later there will still be that active induction schedule lying around with an old deadline date.

nathanrussell-moj-digital (approved):
LGTM, though I think we'll want some unit tests at some point for the new method in `ReviewService`; plus also `PefCiagKpiService` might be running away with itself in respect of class responsibility 🤔 
We can address both of those later though - I think you are right that the important thing in getting this into `dev` ASAP and getting some messages on the queues 👍

PR #457, opened 2024-12-05
RR-1162 - Get all prisoners in prison via paged API call

nathanrussell-moj-digital:
This PR changes the `getAllPrisonersInPrison` method to get all prisoners via paged calls on the prisoner-search API

There is probably a smarter way of doing this using Flux or similar, but that would mean making the whole stack non-blocking (WebFlux) and life is too short! 🤣

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/PrisonerSearchApiService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/PrisonerSearchApiService.kt
@@ -12,15 +12,26 @@ class PrisonerSearchApiService(
   private val prisonerSearchApiClient: PrisonerSearchApiClient,
 ) {
 
-  fun getAllPrisonersInPrison(prisonId: String): List<Prisoner> =
-    prisonerSearchApiClient.getPrisonersByPrisonId(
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/PrisonerSearchApiService.kt
```diff
@@ -12,15 +12,26 @@ class PrisonerSearchApiService(
...
 
-  fun getAllPrisonersInPrison(prisonId: String): List<Prisoner> =
-    prisonerSearchApiClient.getPrisonersByPrisonId(
-      prisonId = prisonId,
-      page = PrisonerSearchApiClient.FIRST_PAGE,
-      pageSize = PrisonerSearchApiClient.DEFAULT_PAGE_SIZE,
-    ).content
+  fun getAllPrisonersInPrison(prisonId: String): List<Prisoner> {
+    var page = 0
+    val pageSize = 250
```
nathanrussell-moj-digital, line 17: I've opted for requests of 250 prisoners at a time  - there might be a more optimal number (or way of coming up with this number), but it seems reasonable enough for now and the use case at hand.

stevomcallister (approved):

PR #461, opened 2024-12-06
update status

stevomcallister:


domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/review/ReviewSchedule.kt
```diff
@@ -35,22 +35,27 @@ enum class ReviewScheduleCalculationRule {
   PRISONER_UN_SENTENCED,
 }
 
-enum class ReviewScheduleStatus(val inScope: Boolean) {
+enum class ReviewScheduleStatus(val inScope: Boolean, val isExclusion: Boolean = false, val isExemption: Boolean = false) {
```
nathanrussell-moj-digital, line 38: Nice addition of the isExclusion property 👍 

We might be able to remove the `inScope` property at some point ... I don't think anything currently uses it. Whilst the concept of a Review Schedule being "in or out of scope" is a thing (hence I added it initially), I _think_ now its a concept that is closer to the MN/Curious implementation than ours. Chances are it's not something we need so can probably remove at some stage 👍

domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/review/service/ReviewScheduleService.kt
```diff
@@ -0,0 +1,135 @@
...
+        ReviewScheduleStatus.SCHEDULED,
+        prisonId,
+        latestReviewDate = newReviewDate,
+        prisonNumber = prisonNumber,
+      ),
+    )
+    performFollowOnEvents(prisonNumber, newStatus, prisonId, newReviewDate)
+  }
+
+  private fun handleSystemException(
```
nathanrussell-moj-digital, line 80: Hmmm, I wonder if we can come up with a better name for this .... it took me quite a while to work out that it was not a (JVM) exception handler! 🤣 

How about something like `updateReviewScheduleFollowingSystemTechnicalIssue` or similar?

Also, I wonder it it would be easier to grok if the method sig did not include `newStatus: ReviewScheduleStatus`, and that `ReviewScheduleStatus.EXEMPT_SYSTEM_TECHNICAL_ISSUE` was used within the method instead?

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/ReviewController.kt
```diff
@@ -75,6 +79,22 @@ class ReviewController(
...
+  @Transactional
+  fun updateLatestReviewScheduleStatus(
+    @Valid
+    @RequestBody createReviewScheduleStatusRequest: CreateReviewScheduleStatusRequest,
+    @PathVariable @Pattern(regexp = PRISON_NUMBER_FORMAT) prisonNumber: String,
+  ) {
+    reviewScheduleService.updateLatestReviewScheduleStatus(
+      prisonNumber,
+      createReviewScheduleStatusRequest.prisonId,
+      createReviewScheduleStatusRequest.status.value,
```
nathanrussell-moj-digital, line 94: Re: my comment on the service class, if we can change the method arg from a sting to the enum type, then you'll need a method here, much like the `toSentenceType` one below.

I know its a few more lines of code, but one of the things I like about kotlin are things like the exhaustive when which means we can be much more type safe in places like this, and we'll get a compilation error if we add a new value in the domain enum but forget to do it in the API (or visa versa)

What do you you think?

nathanrussell-moj-digital (approved):
LGTM; but I wonder if it's worth changing the string method args to instances of the relevant enum (see my comments). What are you thoughts? (happy to approve and merge if you don't feel its necessary)

nathanrussell-moj-digital (approved):
LGTM 👍

PR #469, opened 2024-12-11
Fix Education tests

nathanrussell-moj-digital:


src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/GetEducationTest.kt
```diff
@@ -97,14 +97,12 @@ class GetEducationTest : IntegrationTestBase() {
     assertThat(actual)
       .hasEducationLevel(EducationLevel.SECONDARY_SCHOOL_TOOK_EXAMS)
       .hasNumberOfQualifications(2)
-      .qualification(1) {
-        it.hasSubject("English")
-          .hasLevel(QualificationLevel.LEVEL_3)
+      .qualificationBySubject("English") {
```
nathanrussell-moj-digital, line 100: The core problem is the API spec and impl make no guarantees as to the order in which quals are returned. Therefore asserting by index is flaky
Changed to get the qualification to assert by subject (will only work if the prisoner only has one of any given subject, but it will do for now)

stevomcallister (approved):

PR #470, opened 2024-12-11
Fix timeline tests

nathanrussell-moj-digital:


src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/GetTimelineTest.kt
```diff
@@ -204,12 +204,18 @@ class GetTimelineTest : IntegrationTestBase() {
         ),
       ),
     )
+
     createInduction(prisonNumber, aValidCreateInductionRequest())
+    Thread.sleep(200)
```
nathanrussell-moj-digital, line 209: Perhaps a low-tech solution? but introduced a small sleep/wait between each operation that sets up the data to give the async methods that create the timeline data a change to run in the order we expect them to

stevomcallister (approved):

PR #473, opened 2024-12-12
RR-1049 - Add optional exemption reason when setting Review Schedule as exempted

nathanrussell-moj-digital:
PR thats adds optional exemption reason field when setting Review Schedule as exempted.
Also returns the value in API responses that return Review Schedules.


stevomcallister (approved):
Good stuff!

PR #475, opened 2024-12-13
RR-1170 - Map notes into GoalResponses

nathanrussell-moj-digital:
PR to fix a bug where the response data from the `getGoals` API response does not currently map in the associated notes for each goal.


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/GoalController.kt
```diff
@@ -165,13 +165,28 @@ class GoalController(
...
-  ) =
-    GetGoalsResponse(
+  ): GetGoalsResponse {
+    val response = GetGoalsResponse(
       goalService.getGoals(
         GetGoalsDto(prisonNumber, convertStatuses(statuses)),
       ).map { goal -> goalResourceMapper.fromDomainToModel(goal) },
     )
 
+    // Map each goal response to its corresponding version with notes
```
nathanrussell-moj-digital, line 175: This is the core fix, and is basically what we do in the ActionPlan controller when getting the Action Plan (which includes Goals)

stevomcallister (approved):

PR #478, opened 2024-12-17
adding open api for induction schedule status updates

stevomcallister:


src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -285,6 +285,35 @@ paths:
           $ref: '#/components/responses/404ErrorResponse'
       operationId: get-action-plan-review-schedules
 
+  #
+  # Induction schedule status update
+  # --------------------------------------------------------------------------------
+  '/action-plans/{prisonNumber}/induction/schedule-status':
```
nathanrussell-moj-digital, line 291: We've already got an endpoint `GET /inductions/{prisonerNumber}/induction-schedule` to return the induction schedule. At the moment a prisoner only ever has 1 induction schedule, so if we were to follow that train of thought then perhaps this endpoint should be `PUT /inductions/{prisonerNumber}/induction-schedule` ??

However ... for the same reason that we now have the concept of a history of review schedules, and we have an endpoint to return those histories ..... do you think we'd be better off doing that refactoring first before this PR?

I'm just wondering if we'd be better off parking this one for now, and considering whether we do need the history of changes to the Induction Schedule (even though there will still only be 1 schedule I think 🤔  ); and if we decide we do need that history to implement the DB etc changes, and then change `GET /inductions/{prisonerNumber}/induction-schedule` to `GET /inductions/{prisonerNumber}/induction-schedules` (plural) and making it return the history.
And then obviously the changes up the stack through the integration API etc ...

What do you think? 🤔

PR #480, opened 2024-12-17
RR-1009 - Map `indeterminateSentence` and `recall` fields from prisoner-search-api

nathanrussell-moj-digital:
This is part 1 of a few PRs that will correct the ReviewSchedule calculation following what we have learned recently in respect of indeterminate and recalled prisoners.

We have learnt that there are 2 boolean fields on the `Prisoner` object from `prisoner-search-api` that can help us - `indeterminateSentence` and `recall`

This PR simply maps them into our simplified `Prisoner` pojo object, and updates a couple of tests that make use of the `Prisoner` object

Subsequent PRs will refactor the methods in `ReviewService` that calculate the Review Schedule by making use of these new values.


src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/CreateActionPlanReviewTest.kt
```diff
@@ -126,12 +126,7 @@ class CreateActionPlanReviewTest : IntegrationTestBase() {
   @Test
   fun `should fail to create review given prisoner does not have a review schedule`() {
     // Given
-    val prisonerFromApi = Prisoner(
-      prisonerNumber = prisonNumber,
-      legalStatus = LegalStatus.SENTENCED,
-      releaseDate = LocalDate.now().plusYears(1),
-      prisonId = "BXI",
-    )
+    val prisonerFromApi = aValidPrisoner(prisonerNumber = prisonNumber)
```
nathanrussell-moj-digital, line 129: I've created a test data builder for `Prisoner` 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/client/prisonersearch/PagedPrisonerResponse.kt
```diff
@@ -12,6 +13,10 @@ data class Prisoner(
   val legalStatus: LegalStatus,
   val releaseDate: LocalDate?,
   val prisonId: String?,
+  @field:JsonProperty(value = "indeterminateSentence", defaultValue = "false")
```
nathanrussell-moj-digital, line 16: These are the new fields.
In the `prisoner-search-api` response data they are called `indeterminateSentence` and `recall`, but they are booleans so for readability I have prefixed the properties in our class with `is`

It's also unclear from the `prisoner-search-api` swagger spec as to whether these fields are mandatory or not. For our purposes they need to be mandatory with a definitive `true` or `false` value, so I've taken the approach that they are probably optional in the `prisoner-search-api` response (most things are! 🙄 ), and that the absence of data means `false`

PR #487, opened 2024-12-19
Refactor InboundEventsListenerTest

nathanrussell-moj-digital:
This PR refactors `InboundEventsListenerTest` as follows:

* moves it from the package `uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/messaging` to `uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging` because it's not related to resources (as in REST API resources)
* renames it from `InboundEventsListenerTest` to `PrisonerReceivedEventTest` because the tests in it are specifically related to processing `prisoner.received` events.
  * This was also driven by the ticket I am working on that is to implement a handler for `prisoner.released`, and adding integration tests to the existing class would get massive!
* uses REST API calls to get various resources (rather than calls to the persistence layer to get entities)
  * we had problems with hibernate transactions (that I didn't fully understand !) where the `// When` invocation of the integration test used the running spring app; but the direct db interactions in the `// Given` and `// When` blocks didn't (something like that anyway)
  * we tend to have better assertion classes for the REST API response types, so the assertions are richer and read better
* renamed the test methods to (hopefully) better reflect their purpose & intent


src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/IntegrationTestBase.kt
```diff
@@ -446,6 +454,36 @@ abstract class IntegrationTestBase {
       .returnResult(ActionPlanReviewsResponse::class.java)
       .responseBody.blockFirst()!!
 
+  fun getInductionSchedule(prisonNumber: String): InductionScheduleResponse =
```
nathanrussell-moj-digital, line 457: Added these methods to get the induction schedule, and induction schedule history, via the REST APIs

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/IntegrationTestBase.kt
```diff
@@ -568,6 +607,17 @@ abstract class IntegrationTestBase {
     )
     reviewScheduleHistoryRepository.saveAndFlush(reviewScheduleEntity)
   }
+
+  fun sendDomainEvent(
```
nathanrussell-moj-digital, line 611: Moved this method from the original `InboundEventListenerTest` so that we can use it in other event listener tests

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/PrisonerReceivedEventTest.kt
```diff
@@ -0,0 +1,142 @@
...
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.induction.aValidCreateInductionRequestForPrisonerNotLookingToWork
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.induction.assertThat
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.review.assertThat
+import uk.gov.justice.hmpps.sqs.countMessagesOnQueue
+import java.time.OffsetDateTime
+import java.util.UUID
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.InductionScheduleCalculationRule as InductionScheduleCalculationRuleResponse
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.resource.model.InductionScheduleStatus as InductionScheduleStatusResponse
+
+class PrisonerReceivedEventTest : IntegrationTestBase() {
```
nathanrussell-moj-digital, line 22: This class is what was previously `InboundEventsListenerTest`

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/PrisonerReceivedEventTest.kt
```diff
@@ -0,0 +1,142 @@
...
+  @Test
+  fun `should create new Induction Schedule and send outbound message given 'prisoner received' event for prisoner that does not already have an Induction`() {
```
nathanrussell-moj-digital, line 24: This is the test that used to be called `should send message to service given message is a Notification message`

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/PrisonerReceivedEventTest.kt
```diff
@@ -0,0 +1,142 @@
...
+    createInductionSchedule(
+      prisonNumber = prisonNumber,
+      status = InductionScheduleStatus.COMPLETE,
+    )
+    createInduction(prisonNumber, aValidCreateInductionRequestForPrisonerNotLookingToWork())
+    createActionPlan(prisonNumber)
+
+    // The above calls set the data up but they will also generate events so clear these out before starting the test.
+    // Before clearing the queues though we need to wait until the first "plp.review-schedule.updated" event on the REVIEW queue is received.
+    await untilCallTo {
```
nathanrussell-moj-digital, line 89: This is the crux of the problems I was having.
The `plp.review-schedule.updated` is actually sent twice in this test - once in the `// Given` section as part of the creation of the Induction and Action Plan (this was a recent change, to create the initial Review Schedule on creation of the Induction and 1st Goal; which probably explains why we've not seen this as a problem until now)

Anyway, even though the impl doesnt use an `AsyncEvent....` type class so in theory is not async, the internals of the AWS SNS & SQS etc classes use Completable Futures, so is async!

Because it's async, when we hit the `clearQueues` method we cant guarantee that the SNS message has been sent and mapped (by localstack) to the SQS queue. That is what was happening for me (likely because my machine is slow today and/or localstack is on its knees)

So .... by the time the final assertion block on line 121 was hit:
```
val reviewScheduleEvent = inductionScheduleEventQueue.receiveEvent(QueueType.REVIEW)
```
most times there were 2 messages on the queue!

The fix is to add the `await` until there is at least one message on the queue before clearing the queues.

Phew! Thats a long explanation !

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/PrisonerReceivedEventTest.kt
```diff
@@ -0,0 +1,142 @@
...
+    // wait until the queue is drained / message is processed
+    await untilCallTo {
+      domainEventQueueClient.countMessagesOnQueue(domainEventQueue.queueUrl).get()
+    } matches { it == 0 }
+
+    val actionPlanReviews = getActionPlanReviews(prisonNumber)
+    assertThat(actionPlanReviews)
+      .latestReviewSchedule {
+        it.hasStatus(ReviewScheduleStatus.SCHEDULED)
+          // TODO fix the implementation in PefCiagKpiService.processPrisonerAdmission to make this set the correct calculation rule
```
nathanrussell-moj-digital, line 114: This TODO will also need addressing, but was out of scope for what I was trying to fix here.
But at least the refactoring of this test identified that we need to do it 👍

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -3037,7 +3037,7 @@ components:
           items:
             $ref: '#/components/schemas/InductionScheduleResponse'
       required:
-        - reviewSchedules
+        - inductionSchedules
```
nathanrussell-moj-digital, line 3040: Minor correction here

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/ciagkpi/PefCiagKpiService.kt
```diff
@@ -45,6 +45,14 @@ class PefCiagKpiService(
    * - If the prisoner already has had their Induction created. Then this person is likely to
    * be a coming back into prison and needs to resume their Reviews.
    */
+  /* TODO - the logic is wrong, and needs to be something like:
```
stevomcallister, line 48: This is a worry - have you created a ticket to update it?

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
stevomcallister: oops

stevomcallister (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/ciagkpi/PefCiagKpiService.kt
nathanrussell-moj-digital: Not yet, but will do

nathanrussell-moj-digital: RR-1198

PR #491, opened 2024-12-20
potential fix to telemetry spy bean issue

stevomcallister:
2 fixes in here: 

1. increase the gradle memory args so that we don't see the gradle errors in CI
2. change the telemetry spy bean such that it resets before each test. Fingers crossed it fixes the erroneous telemetry counts.


.circleci/config.yml
```diff
@@ -21,10 +21,11 @@ parameters:
     default: "3.6.0"
   java-options:
     type: string
-    default: -Xmx512m -XX:ParallelGCThreads=2 -XX:ConcGCThreads=2 -Djava.util.concurrent.ForkJoinPool.common.parallelism=2 -Dorg.gradle.daemon=false -Dkotlin.compiler.execution.strategy=in-process -Dorg.gradle.workers.max=1
+    default: -Xmx2g -XX:MaxMetaspaceSize=1g -XX:ParallelGCThreads=2 -XX:ConcGCThreads=2 -Djava.util.concurrent.ForkJoinPool.common.parallelism=2 -Dorg.gradle.daemon=false -Dkotlin.compiler.execution.strategy=in-process -Dorg.gradle.workers.max=1
```
stevomcallister, line 24: This is to fix the gradle memory issues in ci

.circleci/config.yml
```diff
@@ -21,10 +21,11 @@ parameters:
...
 
 jobs:
   validate:
+    resource_class: "medium+"
```
stevomcallister, line 28: needed for the increased memory

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/IntegrationTestBase.kt
```diff
@@ -165,7 +166,7 @@ abstract class IntegrationTestBase {
   @Autowired
   lateinit var reviewScheduleHistoryRepository: ReviewScheduleHistoryRepository
 
-  @SpyBean
+  @MockitoSpyBean(reset = MockReset.BEFORE)
```
stevomcallister, line 169: This should fix the issue with the extra telemetry being reported

PR #498, opened 2025-01-02
RR-1203 - Remove displayName field from Timeline Entity

nathanrussell-moj-digital:
This PR is the first of a few very similar ones that we'll need to do over the next couple of weeks , and is related to the RBAC work.

RBAC is going to become important again soon because we need to have different roles etc for users who can do reviews etc; and in order to do the RBAC work (more specifically, to create granular roles) we need the UI to not use the user token, and instead use the system token. 
We've done this work for the read operations that the UI makes 👍 , but not the write operations yet. In preparation for doing the changes to the write operations (to use the system token) we need to remove the code in the API that writes the "display name" fields to the database entities via the custom annotation that uses the username claim from the token. (ie. before we can get the UI to use the system token that will not carry a username claim, we need to remove the code that reads it and writes it to the database, otherwise we'll get NPEs etc)

As you can see in this PR it's quite a lot of change (but it is mostly deletes, with a little refactoring around the edges), so I'm proposing (and we have tickets) to do each entity (or logical group of entities) at a time (rather than the whole lot in one massive PR)

This PR is the first in this sequence, and addresses the TimelineEvent entity.

### DIFF
--- a/domain/timeline/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/timeline/TimelineEvent.kt
+++ b/domain/timeline/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/timeline/TimelineEvent.kt
@@ -25,7 +25,7 @@ data class TimelineEvent(
   /**
    * A map of useful contextual information about the event.
    */
-  val contextualInfo: Map<TimelineEventContext, String>?,
+  val contextualInfo: Map<TimelineEventContext, String>,
   /**
    * The ID of the Prison where the Prisoner is based at the time of the event.
    */
@@ -34,10 +34,6 @@ data class TimelineEvent(
    * The username of the person who caused this event. Set to 'system' if the event was not actioned by a DPS user.
    */
   val actionedBy: String,
-  /**
-   * The name of the person who caused this event (if applicable).
-   */
-  val actionedByDisplayName: String? = null,
   /**
    * The date and time when the event occurred.
    */
@@ -52,10 +48,9 @@ data class TimelineEvent(
     fun newTimelineEvent(
       sourceReference: String,
       eventType: TimelineEventType,
-      contextualInfo: Map<TimelineEventContext, String>? = null,
+      contextualInfo: Map<TimelineEventContext, String> = emptyMap(),
       prisonId: String,
       actionedBy: String,
-      actionedByDisplayName: String? = null,
       timestamp: Instant = Instant.now(),
       correlationId: UUID = UUID.randomUUID(),
     ) = TimelineEvent(
@@ -65,7 +60,6 @@ data class TimelineEvent(
       contextualInfo = contextualInfo,
       prisonId = prisonId,
       actionedBy = actionedBy,
-      actionedByDisplayName = actionedByDisplayName,
       timestamp = timestamp,
       correlationId = correlationId,
     )
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/entity/timeline/TimelineEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/entity/timeline/TimelineEntity.kt
@@ -10,7 +10,6 @@ import jakarta.persistence.JoinColumn
 import jakarta.persistence.OneToMany
 import jakarta.persistence.OrderBy
 import jakarta.persistence.Table
-import jakarta.validation.constraints.NotNull
 import org.hibernate.Hibernate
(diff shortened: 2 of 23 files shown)


domain/timeline/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/timeline/TimelineEvent.kt
```diff
@@ -25,7 +25,7 @@ data class TimelineEvent(
   /**
    * A map of useful contextual information about the event.
    */
-  val contextualInfo: Map<TimelineEventContext, String>?,
+  val contextualInfo: Map<TimelineEventContext, String>,
```
nathanrussell-moj-digital, line 28: Minor tidying to make this non-nullable; the code that populates this field will now set an empty map rather than null (see below)

domain/timeline/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/timeline/TimelineEvent.kt
```diff
@@ -34,10 +34,6 @@ data class TimelineEvent(
    * The username of the person who caused this event. Set to 'system' if the event was not actioned by a DPS user.
    */
   val actionedBy: String,
-  /**
-   * The name of the person who caused this event (if applicable).
-   */
-  val actionedByDisplayName: String? = null,
```
nathanrussell-moj-digital, line 40: `actionedByDisplayName` is the core field that this PR is removing

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/entity/timeline/TimelineEntity.kt
```diff
@@ -20,34 +19,30 @@ import java.util.UUID
 
 @Table(name = "timeline")
 @Entity
-class TimelineEntity(
-  @Id
-  @GeneratedValue
-  @UuidGenerator
-  var id: UUID? = null,
-
+data class TimelineEntity(
```
nathanrussell-moj-digital, line 22: I've taken the opportunity to refactor the entity class to be a data class, and to move the JPA managed fields (id, createdBy, createdAt etc) out of the constructor args and simply have them as fields.
We already have a couple of entity classes that follow this pattern, and it feels "cleaner", in that the constructor is the public API of this entity, and we would not expect consumers of this code who want to create the entity to have to provide the ID and createdBy/At audit fields (that is the job of the framework / JPA and the lifecycle event listeners)

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/entity/timeline/TimelineEntity.kt
```diff
@@ -20,34 +19,30 @@ import java.util.UUID
...
   @Column(updatable = false)
-  @field:NotNull
-  var reference: UUID? = null,
+  val reference: UUID,
```
nathanrussell-moj-digital, line 24: I've also corrected the cardinality of each field - IE. use the kotlin language as intended and make then non-nullable (and `val` for fields that we do not expect/support changes to be made to)

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/timeline/TimelineEntityMapper.kt
```diff
@@ -1,14 +1,17 @@
...
 
-@Mapper(
-  uses = [
-    TimelineEventEntityMapper::class,
-  ],
-)
-interface TimelineEntityMapper {
-  fun fromEntityToDomain(persisted: TimelineEntity): Timeline
+@Component
+class TimelineEntityMapper(private val timelineEventEntityMapper: TimelineEventEntityMapper) {
```
nathanrussell-moj-digital, line 8: Given that I had to touch the mapper, it made sense to change it from mapstruct to a regular class

src/main/resources/db/migration/common/V2024.12.29.0001__drop_display_name_field_from_timeline_event_table.sql
nathanrussell-moj-digital, line 1: Hmmm, maybe I was "hiding" from the family and doing a little tinkering whilst on my Xmas break! 🤣

stevomcallister (approved):

PR #503, opened 2025-01-03
update induction schedule status 

stevomcallister:
Align the COMPLETE induction schedule status with review schedule which has the status as COMPLETED.


src/testFixtures/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/resource/model/induction/InductionScheduleResponseAssert.kt
```diff
@@ -65,7 +65,7 @@ class InductionScheduleResponseAssert(actual: InductionScheduleResponse?) :
   fun wasUpdatedAtOrAfter(dateTime: OffsetDateTime): InductionScheduleResponseAssert {
     isNotNull
     with(actual!!) {
-      if (updatedAt.isBefore(dateTime)) {
+      if (updatedAt.isBefore(dateTime.truncatedTo(ChronoUnit.MILLIS))) {
```
nathanrussell-moj-digital, line 68: This is intermittently burning us in a few places ( 😢 ) (ie. all the tests that assert dates like this in their various assertJ classes)
This is fine for now, but I was wondering whether we could do this in a way that addresses all the possible places where it might fail .... I was thinking perhaps some kind of extension function on OffsetDate and LocalDate that handles the truncation. Maybe something like `if (updatedAt.isBeforeTruncatedToMillis(datetime))` where the ext function handles the truncation on both operands (ie. `this` and the passed in arg).
But I'm not keen on that name ! 🤣 

Anyway, what you have is fine, my comment was just a thought about how we might be able to address it wider 👍

nathanrussell-moj-digital (approved):
LGTM 👍

PR #508, opened 2025-01-06
RR-1198 - Move getReviewSchedulesForPrisoner to ReviewScheduleService

nathanrussell-moj-digital:
This PR is part of some tidying/refactoring of InductionSchedules and how they are calculated, but also ReviewSchedules where it makes sense (as they are so closely related)

This PR moves the function `getReviewSchedulesForPrisoner` from `ReviewService` to `ReviewScheduleService` which is a more logical place for it.

There are a couple of other "review schedule" related functions that are currently in `ReviewService` that I will look at moving to `ReviewScheduleService` shortly ... 🤔

### DIFF
--- a/domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/review/service/ReviewService.kt
+++ b/domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/review/service/ReviewService.kt
@@ -4,7 +4,6 @@ import mu.KotlinLogging
 import uk.gov.justice.digital.hmpps.domain.learningandworkprogress.review.ActiveReviewScheduleAlreadyExistsException
 import uk.gov.justice.digital.hmpps.domain.learningandworkprogress.review.CompletedReview
 import uk.gov.justice.digital.hmpps.domain.learningandworkprogress.review.ReviewSchedule
-import uk.gov.justice.digital.hmpps.domain.learningandworkprogress.review.ReviewScheduleHistory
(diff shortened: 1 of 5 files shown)


domain/learningandworkprogress/src/test/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/review/service/ReviewScheduleServiceTest.kt
```diff
@@ -249,4 +261,82 @@ class ReviewScheduleServiceTest {
       verifyNoInteractions(reviewScheduleEventService)
     }
   }
+
+  @Nested
```
nathanrussell-moj-digital, line 265: These tests have been moved from `ReviewServiceTest`

domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/review/service/ReviewService.kt
```diff
@@ -55,19 +54,6 @@ class ReviewService(
       prisonNumber,
     )
 
-  /**
-   * Returns a list of [ReviewSchedule] for the prisoner identified by their prison number.
-   */
-  fun getReviewSchedulesForPrisoner(prisonNumber: String): List<ReviewScheduleHistory> {
```
stevomcallister, line 61: I wonder if the method that calls this should also be in it's own controller too - rather than ReviewController?

stevomcallister (approved):

PR #510, opened 2025-01-06
fix to goals test

stevomcallister:
Small tweak to the testing of goals - issue seems to be when the goals are saved they are often saved with the same timestamp which means the ordering cannot be determined. 

The ordering doesn't matter for the testing so this fix addresses this.


nathanrussell-moj-digital (approved):
LGTM 👍

PR #514, opened 2025-01-07
RR-1198 - Move creation of timeline events etc to relevant EventService classes

nathanrussell-moj-digital:
PR to refactor the creation of timeline events, telemetry events, and publishing domain events (for MN to ultimately consume) out of `ReviewScheduleAdapter` and `PefCiagKpiService` and into the relevant EventService interface and impl.

This approach follows the pattern we've already established for doing such things, but more importantly for what I'm trying to do overall, it starts breaking things out of `PefCiagKpiService` which will allow us to refactor it to address the TODOs in it.


src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/CreateActionPlanTest.kt
```diff
@@ -188,8 +188,10 @@ class CreateActionPlanTest : IntegrationTestBase() {
     assertThat(actionPlan)
       .isForPrisonNumber(prisonNumber)
       .hasNumberOfGoals(2)
-      // Because the goals are returned in date created order, the first goal in the response will be Goal 2, because it was the most recently created.
-      .goal(1) {
+      // Because the goals are returned in date created order, the first goal in the response SHOULD be Goal 2, because it was the most recently created.
```
nathanrussell-moj-digital, line 191: Not related to the main changes in this PR, but seeing as I had a failed build I took the opportunity to fix this flaky test which, as we discovered yesterday, is flaky because the order of the goals cannot be asserted reliably.

PR #516, opened 2025-01-07
adding prerelease flag to completed reviews

stevomcallister:


src/main/resources/db/migration/common/V2025.01.07.0002__add_pre_release_field.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE review
```
stevomcallister, line 1: Accidentally added 0001 of this file in a previous commit, which was merged. Thought about fixing the database etc but thought it best to just create an 0002 with the change I wanted in it.

domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/review/service/ReviewService.kt
```diff
@@ -98,6 +98,8 @@ class ReviewService(
           )
         } ?: let {
           // No new ReviewScheduleWindow was calculated, so this was the prisoners last Review before release
+          // update the completed review to have preRelease flag set to true:
```
stevomcallister, line 101: I know we are creating and then updating the review in the same function but this seems like the least complicated way of doing it.

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/CreateActionPlanReviewTest.kt
```diff
@@ -261,4 +262,62 @@ class CreateActionPlanReviewTest : IntegrationTestBase() {
...
+      .expectStatus()
+      .isCreated
+      .returnResult(CreateActionPlanReviewResponse::class.java)
+
+    // Then
+    val reviews = getActionPlanReviews(prisonNumber)
+    assertThat(reviews)
+      .hasNumberOfCompletedReviews(1)
+      .completedReview(1) {
+        it.wasPreRelease()
```
stevomcallister, line 304: This is the only thing I actually wanted to test - does it hurt to test all the other review fields?

domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/review/service/ReviewService.kt
nathanrussell-moj-digital: I kinda get you, but I am struggling with the slight change in pattern, though perhaps its the naming I'm really struggling with 🤔 

I think our existing pattern would say you'd have a method on the persistence adapter - perhaps called `updateCompletedReview` - and you pass into it some DTO or args that identify the entity and the changes you want to make to it. And then it returns a new domain object representing the updated entity.
But I get it - thats a lot of faff for what you want to do here, so I can see why you've done what you have.

I think the problem I have with the name is that its not really telling me what it's doing and/or there's no kdoc to help explain it. I know it's really wordy and verbose but I wouldnt object to something like:
`reviewPersistenceAdapter.markCompletedReviewAsThePrisonersPreReleaseReview(completedReview.reference)`
🤣 

With the persistence adapter interface defining the method as
```
/**
  * Updates the CompletedReview identified by it's reference to mark it as being the prisoner's last review before release.
**/
fun markCompletedReviewAsThePrisonersPreReleaseReview(completedReviewReference: UUID)
```

Or something .... 🤷‍♂️ 
What do you think?

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/repository/ReviewRepository.kt
```diff
@@ -1,11 +1,17 @@
...
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.review.ReviewEntity
 import java.util.UUID
 
 @Repository
 interface ReviewRepository : JpaRepository<ReviewEntity, UUID> {
   fun getAllByPrisonNumber(prisonNumber: String): List<ReviewEntity>
+
+  @Modifying
+  @Query("UPDATE ReviewEntity SET preRelease = TRUE WHERE reference = :reference")
+  fun setPreRelease(reference: UUID)
```
nathanrussell-moj-digital, line 16: As above re: naming on the persistence adapter method; I wonder if this could be named in such a way that others will know what it is doing?

PR #517, opened 2025-01-07
RR-1198 - Move creation of timeline events etc to relevant EventServi…

stevomcallister:
…ce classes


src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/GetTimelineTest.kt
```diff
@@ -190,6 +191,10 @@ class GetTimelineTest : IntegrationTestBase() {
   fun `should get timeline with multiple events in order`() {
     // Given
     val prisonNumber = randomValidPrisonNumber()
+
```
stevomcallister, line 194: Set up the prisoner search wire mock for the prisoner we are using

PR #521, opened 2025-01-08
RR-1198 - Remove CiagKpiService; implement calls to setup InductionSchedule in PrisonerReceivedIntoPrisonEventService

nathanrussell-moj-digital:


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/PrisonerReceivedIntoPrisonEventService.kt
```diff
@@ -35,19 +42,25 @@ class PrisonerReceivedIntoPrisonEventService(
...
+    } catch (e: InductionScheduleAlreadyExistsException) {
+      // Prisoner already has an Induction Schedule
+      when (e.inductionSchedule.scheduleStatus) {
+        COMPLETED -> {
+          // The Induction was completed so we need to reschedule their active Review Schedule if they have one, or create a new Review Schedule.
+          rescheduleOrCreatePrisonersReviewSchedule(nomsNumber)
+        }
+        else -> {
+          // The Induction was not completed so need to reschedule it with a new deadline date.
+          // TODO - reschedule it
```
nathanrussell-moj-digital, line 60: Addressing this TODO should be the last thing we need to do for this ticket/refactoring, at which point I _think_ all of the handling of a prisoner admission in respect of their Induction Schedule and/or Review Schedule _should_ be done 🤞

helm_deploy/hmpps-education-and-work-plan-api/values.yaml
```diff
@@ -36,7 +36,7 @@ generic-service:
     PRISONER_SEARCH_API_URL: https://prisoner-search.prison.service.justice.gov.uk
     HMPPS_SAR_ADDITIONALACCESSROLE: ROLE_EDUCATION_WORK_PLAN_VIEWER
     HMPPS_SQS_ENABLED: true
-    CIAG_KPI_PROCESSING_RULE: "" # Set to PEF or PES depending on Curious support; or empty string to disable CIAG KPI Processing support
+    CIAG_KPI_PROCESSING_RULE: "PEF" # Set to PEF or PES depending on Curious support
```
nathanrussell-moj-digital, line 39: Not strictly related to this PR, more fixing a deployment problem from my last PR! 😬 

The 2 new beans I added in my last PR (`PefInductionScheduleDateCalculationService` and `PesInductionScheduleDateCalculationService`) are dependant on this env var, and where I've wired them in the concept of it being optional (CIAG KPI Processing support being disabled) is no longer a thing!
With this env var set at empty string, neither bean was getting instantiated.
This was not a problem in `dev` because we explicitly set the value to `PEF` in that env, but for preprod and prod we had it set as an empty string. This meant that it does not startup in preprod 😢 

Also see the changes made to `PefInductionScheduleDateCalculationService` and `PesInductionScheduleDateCalculationService` below - we will always get one of the beans now 👍 . If the env var is set to PES we will get the PES bean; and in all other cases (no matter what the env var value is) we will get the PEF bean 👍

PR #522, opened 2025-01-09
initial commit for created/updated at prison

stevomcallister:
How can adding two fields to the induction schedule create so many code changes... 

Anyway, initial code to add in the created/updated at prison id for induction schedules. 

I've tried to keep this pr to the minimum number of changes but it's still huge. I have hardcoded the prison id to 'BXI' in a few places in order to not change too much code in this PR - I will address these in the next one.


src/main/resources/db/migration/common/V2025.01.09.0001__induction_schedule_created_at_prison.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE induction_schedule ADD COLUMN created_at_prison VARCHAR(3) DEFAULT 'BXI';
+ALTER TABLE induction_schedule ADD COLUMN updated_at_prison VARCHAR(3) DEFAULT 'BXI';
+ALTER TABLE induction_schedule_history ADD COLUMN created_at_prison VARCHAR(3) DEFAULT 'BXI';
+ALTER TABLE induction_schedule_history ADD COLUMN updated_at_prison VARCHAR(3) DEFAULT 'BXI';
```
nathanrussell-moj-digital, line 4: Is `DEFAULT` the right keyword/semantics here?
I could be wrong, but I thought DEFAULT sets a default value when the record is created without specifying the value for the column .... which is not really what we want here is it?

I think what we've done elsewhere for this type of change is this type of thing:
```
ALTER TABLE induction_schedule ADD COLUMN created_at_prison VARCHAR(3);
UPDATE induction_schedule SET induction_schedule = "BXI";
ALTER TABLE induction_schedule ALTER COLUMN induction_schedule SET NOT NULL;
```
or something like that ....

nathanrussell-moj-digital (approved):
LGTM 👍

PR #524, opened 2025-01-09
RR-1198 - Reschedule InductionSchedule on prison admission

nathanrussell-moj-digital:
PR to re-schedule an existing non-completed InductionSchedule on prison admission.

It's a bit of a weird edge case, but we need to have it catered for. Basically it's the case where someone has:
* Been admitted to prison
* Gets a scheduled InductionSchedule created
* Prisoner leaves prison before the InductionSchedule has been completed (so is still SCHEDULED or one of the exemption reasons
  * being released _should_ tidy up any InductionSchedules etc, but we've not written that code yet, and I guess there is scope that for whatever reason it might not fire
* Prisoner is re-admitted to prison

In this case they already have their non-completed InductionSchedule from first time round. We cant create them a new one (because a prisoner can only have 1 InductionSchedule), so we need to re-schedule it by setting its status to SCHEDULED and setting the deadline date to the calculated date based on a regular admission (ie. it is not considered an extension of 5 or 10 days like when you clear an exemption. Aliki has confirmed that it should be re-scheduled with the original rules (eg: today + 20 for PEF))


domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/induction/service/InductionScheduleService.kt
```diff
@@ -45,6 +46,32 @@ class InductionScheduleService(
...
+   */
+  fun reschedulePrisonersInductionSchedule(prisonNumber: String, prisonerAdmissionDate: LocalDate): InductionSchedule {
+    val inductionSchedule = inductionSchedulePersistenceAdapter.getInductionSchedule(prisonNumber)
+      ?: throw InductionScheduleNotFoundException(prisonNumber)
+
+    return with(inductionSchedule) {
+      // Validate the status transition
+      inductionScheduleStatusTransitionValidator.validate(prisonNumber, scheduleStatus, SCHEDULED)
+
+      // A rescheduled Induction Schedule must have its due date set as if it were a new Induction Schedule based on the prisoner's admission date
```
nathanrussell-moj-digital, line 66: This is a key bit - the new deadline date is the date as if it were a new Induction Schedule using the date as calculated by the `InductionScheduleDateCalculationService` (and even though we are not creating a new Induction Schedule this seemed like the easiest way to do it)

domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/induction/service/InductionScheduleService.kt
```diff
@@ -61,81 +88,83 @@ class InductionScheduleService(
...
       else -> {
-        updateScheduledStatus(inductionSchedule, prisonNumber)
+        val adjustedInductionDueDate = inductionScheduleDateCalculationService.calculateAdjustedInductionDueDate(inductionSchedule)
+        rescheduleInductionSchedule(inductionSchedule, adjustedInductionDueDate)
       }
     }
   }
 
-  private fun updateExemptStatus(
+  private fun exemptInductionSchedule(
```
nathanrussell-moj-digital, line 114: Some minor refactoring of these private methods, of their names, but also to return the InductionSchedule they have just updated; which means we can return it all the way up the stack (which I wanted to in my new re-schedule method)

domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/induction/service/InductionScheduleStatusTransitionValidator.kt
```diff
@@ -41,12 +41,12 @@ class InductionScheduleStatusTransitionValidator {
       reason = "Cannot transition from COMPLETED to any other status.",
     ),
     ValidationRule(
-      condition = { current, new -> current.isExemptionOrExclusion() && new.isExemptionOrExclusion() },
-      reason = "Cannot transition from one exemption or exclusion status to another.",
+      condition = { current, _ -> current == InductionScheduleStatus.PENDING_INITIAL_SCREENING_AND_ASSESSMENTS_FROM_CURIOUS },
+      reason = "Cannot transition from PENDING_INITIAL_SCREENING_AND_ASSESSMENTS_FROM_CURIOUS to any other status.",
     ),
     ValidationRule(
-      condition = { current, new -> new == InductionScheduleStatus.SCHEDULED && !current.isExemptionOrExclusion() },
```
nathanrussell-moj-digital, line 48: Removed this validation rule (for InductionSchedule status changes) because when re-scheduling an Induction Schedule we _do_ want to set it's status to SCHEDULED

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/PrisonerReceivedDueToAdmissionEventTest.kt
```diff
@@ -0,0 +1,233 @@
+package uk.gov.justice.digital.hmpps.educationandworkplanapi.app.messaging
```
nathanrussell-moj-digital, line 1: This and the next test class look like new test classes; but actually all I've done is split them out of the original `PrisonerReceivedEventTest` (which now looks like a deletion!) because it was getting a bit big!
This test class (to do with prisoner admissions) has some new tests for the new functionality added in this PR

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/PrisonerReceivedDueToAdmissionEventTest.kt
```diff
@@ -0,0 +1,233 @@
...
+      }
+
+    // test that outbound event is also created:
+    val reviewScheduleEvent = inductionScheduleEventQueue.receiveEvent(QueueType.REVIEW)
+    assertThat(reviewScheduleEvent.personReference.identifiers[0].value).isEqualTo(prisonNumber)
+    assertThat(reviewScheduleEvent.detailUrl).isEqualTo("http://localhost:8080/reviews/$prisonNumber/review-schedule")
+  }
+
+  @Test
+  fun `should re-schedule the Induction Schedule and send outbound message given 'prisoner received' (admission) event for prisoner that already has an incomplete Induction Schedule`() {
```
nathanrussell-moj-digital, line 151: This test method (and the next one) are the 2 new tests I've added to support the changes in this PR

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/PrisonerReceivedIntoPrisonEventService.kt
```diff
@@ -57,7 +58,7 @@ class PrisonerReceivedIntoPrisonEventService(
         }
         else -> {
           // The Induction was not completed so need to reschedule it with a new deadline date.
-          // TODO - reschedule it
```
nathanrussell-moj-digital, line 60: This is the TODO that this entire PR is addressing! 🤣

src/test/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/induction/PreviousQualificationsEntityMapperTest.kt
```diff
@@ -53,7 +53,6 @@ class PreviousQualificationsEntityMapperTest {
     // Then
     assertThat(actual)
       .doesNotHaveJpaManagedFieldsPopulated()
-      .hasAReference()
```
nathanrussell-moj-digital, line 56: Unrelated change; but since the changes the other day to PrisonerQualifications (where we got rid of the displayName fields, but also refactored the entity etc to make fields non-nullable), this assertion is no longer valid because it's not possible for the entity to not have a reference. I've removed the method from the assertion class below

stevomcallister: this was on my list to sort too

stevomcallister (dismissed):
Great stuff!

stevomcallister (approved):

PR #529, opened 2025-01-13
Rr 1213 remove mapstruct mapper 1

stevomcallister:


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/entity/induction/WorkOnReleaseEntity.kt
```diff
@@ -55,38 +48,37 @@ class WorkOnReleaseEntity(
...
-
   @Column
   @field:NotNull
   var createdAtPrison: String? = null,
 
+  @Column
+  @field:NotNull
+  var updatedAtPrison: String? = null,
+
+) {
```
nathanrussell-moj-digital, line 59: You _might_ find you can correct the `val` and `var` declarations, and remove the nullability where it makes sense in these constructor args. Something like this:
```
data class WorkOnReleaseEntity(

  @Column(updatable = false)
  val reference: UUID,

  @Column
  @Enumerated(value = EnumType.STRING)
  var hopingToWork: HopingToWork,

  @ElementCollection(targetClass = AffectAbilityToWork::class)
  @Enumerated(value = EnumType.STRING)
  @CollectionTable(name = "affecting_ability_to_work", joinColumns = [JoinColumn(name = "work_on_release_id")])
  @Column(name = "affect")
  val affectAbilityToWork: MutableList<AffectAbilityToWork> = mutableListOf(),

  @Column(name = "affecting_work_other")
  var affectAbilityToWorkOther: String? = null,

  @Column
  val createdAtPrison: String,

  @Column
  var updatedAtPrison: String,

) {
```

I _think_ thats right, in that:
* reference must be set and cannot be changed
* hopingToWork must be set, but can be changed
* affectAbilityToWork is a list where the object itself cannot change, but it's content can
* affectAbilityToWorkOther does not need to be set, and can be changed
* createdAtPrison must be set and cannot be changed
* updatedAtPrison must be set, but can be changed

This should remove some of the `!!`'s still kicking around 👍 , but it _might_ be the type of change that runs away with itself a little, so it if looks like a PITA dont worry about it and we can look at it again at some point in the future

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/induction/WorkOnReleaseEntityMapper.kt
```diff
@@ -1,35 +1,119 @@
...
+
+  fun toEntity(list: List<AffectAbilityToWork>): MutableList<AffectAbilityToWorkEntity> {
+    return list.mapTo(ArrayList(list.size)) { affectAbilityToWorkToAffectAbilityToWork(it) }
+  }
+
+  fun toModel(list: List<AffectAbilityToWorkEntity>?): List<AffectAbilityToWork> {
+    return list?.map { toAffectAbilityToWork(it) } ?: emptyList()
+  }
+
+  fun toAffectAbilityToWork(affectAbilityToWork: uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.induction.AffectAbilityToWork): AffectAbilityToWork {
```
nathanrussell-moj-digital, line 72: Minor, but wonder if it would be worth using an import alias here (eg: `AffectAbilityToWorkEntity`) rather than fully qualifying it?

nathanrussell-moj-digital (approved):
LGTM;

A couple of minor comments/suggestions, but happy to go with it as is if you want to 👍

PR #532, opened 2025-01-14
Fix `GetGoalsTest`

nathanrussell-moj-digital:
PR to fix the `GetGoalsTest` integration test (once and for all with any luck 🤞 )

It does this by forcing the creation order of the goals, but creating them one and a time via the API call. The tests in this test class (that do care about goal order) can then be more definitive about the assertions and goal ordering etc.

### DIFF
--- a/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/IntegrationTestBase.kt
+++ b/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/IntegrationTestBase.kt
@@ -16,7 +16,7 @@ import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.boot.test.autoconfigure.web.reactive.AutoConfigureWebTestClient
 import org.springframework.boot.test.context.SpringBootTest
 import org.springframework.boot.test.context.SpringBootTest.WebEnvironment.RANDOM_PORT
-import org.springframework.http.MediaType
+import org.springframework.http.MediaType.APPLICATION_JSON
 import org.springframework.test.context.ActiveProfiles
 import org.springframework.test.context.DynamicPropertyRegistry
 import org.springframework.test.context.DynamicPropertySource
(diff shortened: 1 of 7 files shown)


src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/IntegrationTestBase.kt
```diff
@@ -273,21 +275,50 @@ abstract class IntegrationTestBase {
     prisonNumber: String,
     createActionPlanRequest: CreateActionPlanRequest = aValidCreateActionPlanRequest(),
     username: String = "auser_gen",
-    displayName: String = "Albert User",
```
nathanrussell-moj-digital, line 276: God rid of this whilst I was here (it is used in the builder for the JWT token, and used to be relevant when we used the displayName field from the JWT when writing the entities .... but we don't do that anymore 😁 )

The real change here though is the new arg `testSensitiveToGoalCreationOrder`
It defaults to false, so all the tests currently using this method can do so as before 👍 
When it is `false` we call the Create Action Plan API call, complete with the goals, exactly as we did before

But when it is `true`, we are saying the test cares about the goal insertion order. In this case we call the Create Action Plan API, but with only the first goal. This creates the Action Plan with 1 goal. Then we iterate through the remaining goals, and call the Create Goals API endpoint, with each goal, one at a time.

It might make for some slightly inefficient/slow tests, but we will only set `testSensitiveToGoalCreationOrder` when we really need it (eg: `GetGoalsTest`), so it is only those tests that are a little slower. The plus side though is that because each Goal is created via it's own API call we can be sure of the order 👍

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/GetGoalsTest.kt
```diff
@@ -90,13 +90,27 @@ class GetGoalsTest : IntegrationTestBase() {
...
     // Then
-    val actual = response.responseBody.blockFirst()!!
-    assertThat(actual.goals).hasSize(5)
-    actual.goals.onEach {
-      assertThat(it).hasStatus(GoalStatus.ACTIVE)
-    }
-    val actualTitles = actual.goals.map { it.title }
-    assertThat(actualTitles).containsExactlyInAnyOrder("Goal 1", "Goal 2", "Goal 3", "Goal 4", "Goal 5")
+    val actual = response.responseBody.blockFirst()
+    assertThat(actual)
```
nathanrussell-moj-digital, line 94: We can now be much clearer and definitive about the assertions of the goals

stevomcallister (approved):

PR #533, opened 2025-01-15
adding notes to induction

stevomcallister:
Almost there, still telemetry to go...


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/induction/InductionEntityMapper.kt
```diff
@@ -89,6 +90,35 @@ abstract class InductionEntityMapper {
       conductedBy = inductionEntity.conductedBy,
       conductedByRole = inductionEntity.conductedByRole,
       completedDate = inductionEntity.completedDate,
+      note = noteEntity?.let { NoteMapper.fromEntityToDomain(noteEntity) },
+    )
+
+  fun fromEntityToDomain(
```
nathanrussell-moj-digital, line 96: Did you need to overload this method, given you add it as a nullable arg with a default value of null in the original copy of the method (line 70)
Or is this for something else?

nathanrussell-moj-digital (approved):
LGTM 👍 

Just one minor question re: whether we needed to overload the `fromEntityToDomain` method ?

PR #545, opened 2025-01-20
Added the extra fields to the induction schedule mapper

stevomcallister:
This should be the last piece of work in our API to get the new fields over to meganexus via the integration API.


nathanrussell-moj-digital (approved):
LGTM 👍

PR #549, opened 2025-01-22
handle prisoner transfer messages wrt inductions

stevomcallister:


domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/induction/service/InductionScheduleService.kt
```diff
@@ -78,7 +78,13 @@ class InductionScheduleService(
         admissionDate = prisonerAdmissionDate,
         prisonId = prisonId,
       ).deadlineDate
-      rescheduleInductionSchedule(this, newInductionDeadlineDate, prisonId)
```
stevomcallister, line 81: Quite a lot of refactoring here - There were three or four methods that updated the induction in various ways then performed the follow on events etc. 

Basically merged them in to one method then updated where each one was called from.

nathanrussell-moj-digital (approved):
LGTM 👍

PR #551, opened 2025-01-24
added etl endpoint to create induction and review schedules

stevomcallister:


domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/induction/InductionSchedule.kt
```diff
@@ -39,12 +39,7 @@ data class InductionSchedule(
 
 enum class InductionScheduleCalculationRule(val existingPrisonerWhenScheduleCreated: Boolean) {
   NEW_PRISON_ADMISSION(false),
-  EXISTING_PRISONER_LESS_THAN_6_MONTHS_TO_SERVE(true),
-  EXISTING_PRISONER_BETWEEN_6_AND_12_MONTHS_TO_SERVE(true),
-  EXISTING_PRISONER_BETWEEN_12_AND_60_MONTHS_TO_SERVE(true),
-  EXISTING_PRISONER_INDETERMINATE_SENTENCE(true),
-  EXISTING_PRISONER_ON_REMAND(true),
-  EXISTING_PRISONER_UN_SENTENCED(true),
+  EXISTING_PRISONER(true),
```
stevomcallister, line 42: As discussed only really need two calculation rules - one for existing prisoners pre KPI rules and one for when new prisoners come on board.

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/ScheduleEtlController.kt
stevomcallister, line 1: This is basically a copy of the review schedule ETL then refactored into methods. Then I've added the logic to create the induction schedules where appropriate. I guess the review ETL can be removed once we are happy that this works but kept it in for now so I can do some testing.

nathanrussell-moj-digital (approved):
LGTM 👍

PR #559, opened 2025-01-28
Fix username resolution in UserPrincipalAuditorAware

nathanrussell-moj-digital:
This PR fixes the username that gets written to JPA fields via `UserPrincipalAuditorAware`, specifically in the case where the JWT is a system token and not one enriched with the DPS username

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/config/AuthAwareTokenConverter.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/config/AuthAwareTokenConverter.kt
@@ -12,7 +12,7 @@ class AuthAwareTokenConverter : Converter<Jwt, AbstractAuthenticationToken> {
   private val jwtGrantedAuthoritiesConverter: Converter<Jwt, Collection<GrantedAuthority>> = JwtGrantedAuthoritiesConverter()
 
   companion object {
-    const val SYSTEM: String = "system"
+    const val SYSTEM_USER: String = "system"
   }
 
   override fun convert(jwt: Jwt): AbstractAuthenticationToken {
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/UserPrincipalAuditorAware.kt
```diff
@@ -3,12 +3,14 @@ package uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa
...
 import org.springframework.stereotype.Component
-import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.config.AuthAwareTokenConverter
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.config.AuthAwareTokenConverter.Companion.SYSTEM_USER
 import java.util.Optional
 
 @Component
 class UserPrincipalAuditorAware : AuditorAware<String> {
   override fun getCurrentAuditor(): Optional<String> = Optional.of(
-    SecurityContextHolder.getContext()?.authentication?.name ?: AuthAwareTokenConverter.SYSTEM,
+    SecurityContextHolder.getContext()?.authentication?.principal?.let {
```
nathanrussell-moj-digital, line 12: This was the core fix - we need to be using the principal (which is a string, as set in `AuthAwareTokenConverter` above)
Using `name` as before effectively pulls out the JWT `sub` claim, which is not what we want

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/config/AuthAwareTokenConverter.kt
```diff
@@ -12,7 +12,7 @@ class AuthAwareTokenConverter : Converter<Jwt, AbstractAuthenticationToken> {
   private val jwtGrantedAuthoritiesConverter: Converter<Jwt, Collection<GrantedAuthority>> = JwtGrantedAuthoritiesConverter()
 
   companion object {
-    const val SYSTEM: String = "system"
+    const val SYSTEM_USER: String = "system"
```
nathanrussell-moj-digital, line 15: Renamed this const, and refactored uses of a similarly named duplicate const to ensure it is only defined once

stevomcallister (approved):

PR #561, opened 2025-01-28
Fix dates on InductionScheduleHistoryEntity and ReviewScheduleHistoryEntity

nathanrussell-moj-digital:
This PR fixes a problem when recording Induction Schedule History or Review Schedule History records, in that the  updatedAt dates do not get correctly updated, which is likely to cause MN a problem when we send them the history and they expect to sort them on updated date 🤣 

image

The problem (I suspect) is that in some cases (the status update methods) we were calling `save` rather than `saveAndFlush` on the Induction/Review Schedule entity classes, and I don't think JPA updates the audit fields via the entiry annotations until it flushes. Therefore, the first history record is correct, but when making a status update, the updatedAt date is copied from the original entity rather than the one that has been flushed with it's new updatedAt value.

I've also been explicit in the History entity classes re: them being immutable, and making the fields all `val` and putting `@Column(updatable = false)` on every field. It might be overly explicit, but hopefully future devs will realise these are history/audit records and are never updated 🤷‍♂️ 🤞

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/JpaReviewSchedulePersistenceAdapter.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/JpaReviewSchedulePersistenceAdapter.kt
@@ -34,11 +34,13 @@ class JpaReviewSchedulePersistenceAdapter(
       if (getActiveReviewSchedule(prisonNumber) != null) {
         throw ActiveReviewScheduleAlreadyExistsException(prisonNumber)
       }
-      val persistedEntity = reviewScheduleRepository.saveAndFlush(
+
+      reviewScheduleRepository.saveAndFlush(
         reviewScheduleEntityMapper.fromDomainToEntity(this),
-      )
-      saveReviewScheduleHistory(persistedEntity)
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/JpaReviewSchedulePersistenceAdapter.kt
```diff
@@ -80,9 +83,10 @@ class JpaReviewSchedulePersistenceAdapter(
       updatedAtPrison = updateReviewScheduleStatusDto.prisonId
     }
 
-    val savedReviewScheduleEntity = reviewScheduleRepository.save(reviewScheduleEntity)
-    saveReviewScheduleHistory(savedReviewScheduleEntity)
-    return reviewScheduleEntityMapper.fromEntityToDomain(savedReviewScheduleEntity)
+    return reviewScheduleRepository.saveAndFlush(reviewScheduleEntity).let {
```
nathanrussell-moj-digital, line 86: And I think this is the other problem line, in that we were creating the history record from the saved (but not flushed) review schedule record

stevomcallister (approved):

PR #564, opened 2025-01-30
RR-1238 - Refactor date calculation tests for Review Schedules

nathanrussell-moj-digital:
This PR refactors the tests that use the `ReviewScheduleDateCalculationService` to better support the changes we are likely going to need to do to account for the "1st April" stuff 😢 

Currently:
* conveniently the `ReviewScheduleDateCalculationService` is the only class responsible for calculating review dates 👍 
* `ReviewScheduleDateCalculationService` takes an optional constructor arg for a date which is the "offset". If the arg is not supplied, "today" is used as the basis for all date calculations. If the arg is supplied, then the later of "today" or the supplied offset date is used as the basis for date calcs.
* The current unit tests ALL new up `ReviewScheduleDateCalculationService` without the optional arg; so all the tests are based on "today". All the tests are essentially what will happen after 1st April
* The current tests are a little bit messy and there is some duplication

This PR refactors the tests so that:
* All code permutation and scenarios of working out `ReviewScheduleCalculationRule`s and `ReviewScheduleWindow`s are properly unit tested in `ReviewScheduleDateCalculationServiceTest`
* `ReviewServiceTest` and `ReviewScheduleServiceTest` now use a mock `ReviewScheduleDateCalculationService` and no longer have the parameterised tests etc that essentially test what is now being tested in `ReviewScheduleDateCalculationServiceTest`

My basic thinking of what we _will_ do (NOT what is in _this_ PR), is that:
* `ReviewServiceTest` and `ReviewScheduleServiceTest` (or their implementations) won't need changing because they use a mock of `ReviewScheduleDateCalculationService`
* We leave `ReviewScheduleDateCalculationServiceTest` as it is now, where it does not provide the optional offset date when newing up `ReviewScheduleDateCalculationServive` - we can leave it as testing the post-April behaviour.
* We can write a new version of `ReviewScheduleDateCalculationServiceTest` that does provide the offset date; and the tests in this class will be based on the review date rules that the BA's have given us for pre-April / "existing prisoner" behaviour
* We make changes in `ReviewScheduleDateCalculationService` as necessary, and ensure that all tests are passing


domain/learningandworkprogress/src/test/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/review/service/ReviewScheduleCalculationServiceTest.kt
nathanrussell-moj-digital, line 1: Not really a delete; it's been renamed to `ReviewScheduleDateCalculationServiceTest` (the previous name was missing the word "Date" in it)
It's also been massively refactored to include new tests predominantly from the 2 deleted test classes, hence it looks like a new file

domain/learningandworkprogress/src/test/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/review/service/ReviewScheduleServiceCreateInitialReviewScheduleTest.kt
nathanrussell-moj-digital, line 1: Deleted; the tests in here are now covered in the above test class

domain/learningandworkprogress/src/test/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/review/service/ReviewServiceCreateReviewTest.kt
nathanrussell-moj-digital, line 1: Deleted; the tests in here are now covered in the test class `ReviewScheduleDateCalculationServiceTest` above

stevomcallister (approved):

PR #569, opened 2025-02-11
Specific rules for existing prisoners wrt induction deaslines

stevomcallister:
PR to address the following:

For prisoners who have a release date that falls prior to the deadline set on the 1st October 2025 then; 

We will set the deadline as their release date minus 7 days. This will reduce the risk of setting deadlines that are already overdue on go-live
If a prisoner is due to be released within 7 days of go live date (i.e. if they are released before 8 April) then they are exempt from the requirement

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/inducton/PefInductionScheduleDateCalculationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/inducton/PefInductionScheduleDateCalculationService.kt
@@ -38,6 +38,7 @@ class PefInductionScheduleDateCalculationService(
     admissionDate: LocalDate,
     prisonId: String,
     newAdmission: Boolean,
+    releaseDate: LocalDate?,
   ): CreateInductionScheduleDto {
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/inducton/PefInductionScheduleDateCalculationService.kt
```diff
@@ -50,14 +51,24 @@ class PefInductionScheduleDateCalculationService(
...
       CreateInductionScheduleDto(
         prisonNumber = prisonNumber,
-        deadlineDate = baseScheduleDate().plusMonths(6),
+        deadlineDate = calculateDeadlineDate(releaseDate),
         scheduleCalculationRule = InductionScheduleCalculationRule.EXISTING_PRISONER,
         scheduleStatus = InductionScheduleStatus.SCHEDULED,
         prisonId = prisonId,
       )
     }
   }
```
stevomcallister, line 61: This is the main meat for this PR

nathanrussell-moj-digital (approved):
LGTM 👍

PR #587, opened 2025-02-24
add prison/person search api spec

stevomcallister:


src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -1777,6 +1800,62 @@ components:
       required:
         - prisonNumbers
 
+    PrisonerSearchCriteria:
```
nathanrussell-moj-digital, line 1803: I know it's a few fields, but I was thinking these would be query string params? 🤔 
I guess the downside is they are not nicely encapsulated in a nice object, but we can create a little data class for that perhaps? 🤔 

What do you think?

nathanrussell-moj-digital (approved):
LGTM 👍

PR #599, opened 2025-02-27
early commit of search with paging and sorting

stevomcallister:


nathanrussell-moj-digital (approved):
Happy to approve this to see if it helps, but we've not had to do anything like this anywhere else in the project so far - we've not had to mess with any kind of date classes, especially not java.sql.Date (I hadn't spotted that in the first PR 🤣 )

Lets go with it to see if it helps, but everywhere else (probably because we are using JPA repository methods that Spring handles for us; rather than native SQL queries ?), we've not had to do anything like this.

We might want to revisit this when you are back from your hols 🤔

PR #603, opened 2025-02-28
Added filtering

stevomcallister:
Implemented Filtering and some basic tests


src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -1909,7 +1909,7 @@ components:
           minimum: 1
           default: 1
           description: Page number for pagination
-        size:
+        pageSize:
```
stevomcallister, line 1912: updated to be inline with the existing searches

nathanrussell-moj-digital (approved):

PR #607, opened 2025-03-05
RR-1270 - fix bug by calling prisoner-search-api once rather than as a paged request

nathanrussell-moj-digital:
This PR fixes a bug we were seeing with the `GET /session/{prisonId}/summary` endpoint, where it would return differing responses (session counts) on different calls.

The way `GET /session/{prisonId}/summary` works is that it returns session counts for prisoners in the given prison.
In our database we store records keyed by prisoner NOMS ID, but we do not hold the prison ID. Therefore in order to get the number of sessions for prisoners in BXI, you first need to know which prisoners (their NOMS ID) are in BXI, and from that you can then query the relevant PLP databases.

To get the list of prisoners in the given prison we call a `prisoner-search-api` endpoint. The [endpoint in question is a paged API](https://prisoner-search-dev.prison.service.justice.gov.uk/swagger-ui/index.html#/Popular/findByPrison), so we have written our service method to call it in a paged manner, with a nominal page size of 250.
A worked example would be Brixton where in `dev` there are 868 prisoners. With a page size of 250 we would call the API 4 times, with 3 x 250 responses plus a final response of 118
The service code keeps appending the responses to a list of prisoners, then returns the list 👍 

We have found that calling prisoner-search-api in this way is not a reliable way of getting all the prisoners in a prison. We suspect that this is because even though it always returns a total of 868 records over the 4 pages, the order of those prisoners is inconsistent, and (more problematic) you can get duplicates across page requests. We suspect this is probably an indexing or similar problem in `prisoner-search-api`:
Given the prisoners A, B, C, D, E, F, G, H, I and J; if you were to call the paged API with a page size of 3 you might expect responses:
1: [A, B, C]
2: [D, E, F]
3: [G, H, I]
4: [J]
The order is irrelevant, just as long as you get all records, and they are only represented once.

What we suspect is happening is that we get back responses such as:
1: [A, J, D]
2: [E, D, F]
3: [B, C, I]
4: [H]
where D is returned twice, and G is not returned at all.

The quick fix (for us) is to use a large page size when calling `prisoner-search-api` which ensures we get all the prisoners in one request. (though a better fix would be in `prisoner-search-api` somewhere ...)

### DIFF
--- a/domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/induction/InductionSchedule.kt
+++ b/domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/induction/InductionSchedule.kt
@@ -71,8 +71,4 @@ enum class InductionScheduleStatus(
   fun isExemptionOrExclusion(): Boolean {
     return isExemption || isExclusion
   }
-
-  fun includeExceptionOnSummary(): Boolean {
-    return (isExemption || isExclusion) && includeExemptionOnSummary
-  }
 }
--- a/domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/review/ReviewSchedule.kt
+++ b/domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/review/ReviewSchedule.kt
@@ -63,10 +63,6 @@ enum class ReviewScheduleStatus(
   fun isExemptionOrExclusion(): Boolean {
     return isExemption || isExclusion
   }
-
-  fun includeExceptionOnSummary(): Boolean {
-    return (isExemption || isExclusion) && includeExemptionOnSummary
-  }
 }
 
 data class ReviewScheduleWindow(
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/PrisonerSearchApiService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/PrisonerSearchApiService.kt
@@ -10,11 +10,12 @@ private val log = KotlinLogging.logger {}
 @Service
 class PrisonerSearchApiService(
   private val prisonerSearchApiClient: PrisonerSearchApiClient,
+// prisoner-search-api appears to be buggy in it's paged API implementation, so safest to request a large page size to get all records in one hit (arguably not the most efficient though)
+  private val pageSize: Int = 9999,
 ) {
 
   fun getAllPrisonersInPrison(prisonId: String): List<Prisoner> {
     var page = 0
-    val pageSize = 250
 
     val prisoners = mutableListOf<Prisoner>()
 
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/SessionSummaryService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/SessionSummaryService.kt
@@ -26,15 +26,19 @@ class SessionSummaryService(
   private val inductionScheduleService: InductionScheduleService,
 ) {
 
-  val today: LocalDate = LocalDate.now()
-
   /**
    * returns due, overdue and on hold counts for Inductions and Reviews
    * for all prisoners within a given prison.
    */
   fun getSessionSummaries(prisonId: String): SessionSummaryResponse {
     val prisoners = prisonerSearchApiService.getAllPrisonersInPrison(prisonId)
-    val prisonerNumbers = prisoners.map { it.prisonerNumber }
+      .also {
+        log.debug { "${it.size} prisoners returned from prisoner-search-api" }
+      }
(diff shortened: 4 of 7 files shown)


domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/induction/InductionSchedule.kt
```diff
@@ -71,8 +71,4 @@ enum class InductionScheduleStatus(
   fun isExemptionOrExclusion(): Boolean {
     return isExemption || isExclusion
   }
-
-  fun includeExceptionOnSummary(): Boolean {
```
nathanrussell-moj-digital, line 75: Not strictly part of the bug fix, but as part of diagnosing it I ended up refactoring elements of `SessionSummaryService` and I've implemented this method as a private extension function in `SessionSummaryService` because that is closer and more relevant to where it is used

domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/review/ReviewSchedule.kt
```diff
@@ -63,10 +63,6 @@ enum class ReviewScheduleStatus(
   fun isExemptionOrExclusion(): Boolean {
     return isExemption || isExclusion
   }
-
-  fun includeExceptionOnSummary(): Boolean {
```
nathanrussell-moj-digital, line 67: As above, not strictly part of the bug fix, but as part of diagnosing it I ended up refactoring elements of `SessionSummaryService` and I've implemented this method as a private extension function in `SessionSummaryService` because that is closer and more relevant to where it is used

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/PrisonerSearchApiService.kt
```diff
@@ -10,11 +10,12 @@ private val log = KotlinLogging.logger {}
 @Service
 class PrisonerSearchApiService(
   private val prisonerSearchApiClient: PrisonerSearchApiClient,
+// prisoner-search-api appears to be buggy in it's paged API implementation, so safest to request a large page size to get all records in one hit (arguably not the most efficient though)
```
nathanrussell-moj-digital, line 13: This is the core fix - set a default initial value of 9999 to get us out of bother here and now, but its now a constructor arg so we can set different values as and when the time comes

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/SessionSummaryService.kt
```diff
@@ -26,15 +26,19 @@ class SessionSummaryService(
...
    * returns due, overdue and on hold counts for Inductions and Reviews
    * for all prisoners within a given prison.
    */
   fun getSessionSummaries(prisonId: String): SessionSummaryResponse {
     val prisoners = prisonerSearchApiService.getAllPrisonersInPrison(prisonId)
-    val prisonerNumbers = prisoners.map { it.prisonerNumber }
+      .also {
+        log.debug { "${it.size} prisoners returned from prisoner-search-api" }
+      }
+    val prisonerNumbers = prisoners.map { it.prisonerNumber }.toSet().sorted().toList()
```
nathanrussell-moj-digital, line 38: Some debug logging, but also remove any duplicates from the prisoner search list (this is part of how I found the root cause of this bug)

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/SessionSummaryService.kt
```diff
@@ -99,49 +103,79 @@ class SessionSummaryService(
...
-          sessionSummaries.exemptInductions.add(schedule)
-        }
-        schedule.scheduleStatus == INDUCTION_SCHEDULED && schedule.deadlineDate < today -> {
-          sessionSummaries.overdueInductions.add(schedule)
-        }
-        schedule.scheduleStatus == INDUCTION_SCHEDULED &&
-          today in schedule.deadlineDate.minusMonths(2)..schedule.deadlineDate -> {
-          sessionSummaries.dueInductions.add(schedule)
-        }
+        schedule.includeInExemptCount() -> sessionSummaries.exemptInductions.add(schedule)
```
nathanrussell-moj-digital, line 114: Refactored the logic into private extension functions to make it easier to grok this code

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/SessionSummaryService.kt
```diff
@@ -26,15 +26,19 @@ class SessionSummaryService(
   private val inductionScheduleService: InductionScheduleService,
 ) {
 
-  val today: LocalDate = LocalDate.now()
```
nathanrussell-moj-digital, line 29: I initially thought this could be the cause of the bug, but it wasn't.
I could however be the cause of other bugs, because `today` is a class member and is only initialised when the bean is instantiated. Therefore, unless the bean is re-instantiated `today` will be wrong tomorrow! 🤣 

I moved it down into the method that uses it 👇

PR #610, opened 2025-03-11
RR-1296 - fix bug where action_plan_created event is not created

nathanrussell-moj-digital:


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/AsyncActionPlanEventService.kt
```diff
@@ -29,9 +27,7 @@ class AsyncActionPlanEventService(
       telemetryService.trackGoalCreatedEvent(it)
     }
 
-    val induction = inductionService.getInductionForPrisoner(actionPlan.prisonNumber)
```
nathanrussell-moj-digital, line 32: This line throws an exemption when there is no induction ...

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/AsyncActionPlanEventService.kt
```diff
@@ -29,9 +27,7 @@ class AsyncActionPlanEventService(
...
-
-    val timelineEvents = timelineEventFactory.actionPlanCreatedEvent(actionPlan, induction)
```
nathanrussell-moj-digital, line 34: ... which means this line never gets executed

stevomcallister (approved):

PR #611, opened 2025-03-11
Handle DB constraint error

stevomcallister:
Not 100% happy with this as it feels like a bit of a bodge.

Basically when we get the constraint issue - pause and then try again. I'm presuming that this is happening because the timeline records are still being written to the database asynchronously and some tests aren't using the await and check the timeline is complete - since they don't need to test this.

### DIFF
--- a/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/IntegrationTestBase.kt
+++ b/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/IntegrationTestBase.kt
@@ -6,10 +6,12 @@ import com.fasterxml.jackson.annotation.JsonProperty
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.module.kotlin.readValue
 import com.microsoft.applicationinsights.TelemetryClient
+import mu.KotlinLogging
 import org.awaitility.Awaitility
 import org.awaitility.kotlin.await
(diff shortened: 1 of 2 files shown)


src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/IntegrationTestBase.kt
```diff
@@ -209,20 +213,30 @@ abstract class IntegrationTestBase {
...
+        reviewScheduleRepository.deleteAll()
+        reviewScheduleHistoryRepository.deleteAll()
+        noteRepository.deleteAll()
+        inductionScheduleRepository.deleteAll()
+        inductionScheduleHistoryRepository.deleteAll()
+
+        clearQueues()
+        wiremockService.resetAllStubsAndMappings()
+
+        return
```
nathanrussell-moj-digital, line 233: I'm assuming this `return` returns out of the whole `repeat` rather than the `try` block; and assuming that it only runs twice when it fails?

nathanrussell-moj-digital (approved):

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/IntegrationTestBase.kt
stevomcallister: yes the return exits the 'loop' `repeat` is literally a wrapper for a for loop

PR #615, opened 2025-03-12
Removed validPrisonNumber and anotherValidPrisonNumber

stevomcallister:
part 2 of this test tidy

### DIFF
--- a/domain/learningandworkprogress/src/testFixtures/kotlin/uk/gov/justice/digital/hmpps/domain/CommonBuilders.kt
+++ b/domain/learningandworkprogress/src/testFixtures/kotlin/uk/gov/justice/digital/hmpps/domain/CommonBuilders.kt
@@ -9,8 +9,6 @@ import java.util.UUID
 
 fun aValidPrisonNumber() = "A1234BC"
 
-fun anotherValidPrisonNumber() = "B5678CD"
-
 fun randomValidPrisonNumber(): String {
   val letters = ('A'..'Z')
   val numbers = ('0'..'9')
--- a/domain/learningandworkprogress/src/testFixtures/kotlin/uk/gov/justice/digital/hmpps/domain/CommonBuilders.kt
+++ b/domain/learningandworkprogress/src/testFixtures/kotlin/uk/gov/justice/digital/hmpps/domain/CommonBuilders.kt
@@ -8,16 +8,9 @@ import java.util.UUID
  */
 
 fun randomValidPrisonNumber(): String {
-  val letters = ('A'..'Z')
-  val numbers = ('0'..'9')
-
-  val firstLetter = letters.random()
-  val secondLetter = letters.random()
-  val thirdLetter = letters.random()
-
-  val numberPart = numbers.random().toString().padStart(4, '0')
-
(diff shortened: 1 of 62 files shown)


domain/learningandworkprogress/src/testFixtures/kotlin/uk/gov/justice/digital/hmpps/domain/CommonBuilders.kt
```diff
@@ -7,10 +7,6 @@ import java.util.UUID
  * prison numbers, times and dates etc.
  */
 
-fun aValidPrisonNumber() = "A1234BC"
-
-fun anotherValidPrisonNumber() = "B5678CD"
```
nathanrussell-moj-digital, line 12: One for another day perhaps, but I was wondering recently about the randomValidPrisonNumber function, and specifically line 22:
```
val numberPart = numbers.random().toString().padStart(4, '0')
```
As I read that (and I could be wrong), I think that generates a 4 digit string, where the first 3 characters are `0`, and only the 4th is a random digit? IE. Its creating a random string number of between `0000` and `0009`
(I think thats what it does anyway 🤷‍♂️ )

I wonder if it should be something as pedestrian as: 
```
val numberPart = "${numbers.random()}${numbers.random()}${numbers.random()}${numbers.random()}"
```
(though theres probably something cleverer than that 😁 )

domain/personallearningplan/src/testFixtures/kotlin/uk/gov/justice/digital/hmpps/domain/CommonBuilders.kt
```diff
@@ -7,7 +7,7 @@ import java.util.UUID
  * prison numbers, times and dates etc.
  */
 
-fun aValidPrisonNumber() = "A1234BC"
+fun randomValidPrisonNumber() = "A1234BC"
```
nathanrussell-moj-digital, line 10: One for a future change to make the impl return a valid one? (its not very random at the moment 🤣 )

src/testFixtures/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/CommonBuilders.kt
```diff
@@ -7,8 +7,17 @@ import java.util.UUID
...
+fun randomValidPrisonNumber(): String {
+  val letters = ('A'..'Z')
+  val numbers = ('0'..'9')
 
-fun anotherValidPrisonNumber() = "B5678CD"
+  val firstLetter = letters.random()
+  val secondLetter = letters.random()
+  val thirdLetter = letters.random()
+
+  val numberPart = numbers.random().toString().padStart(4, '0')
```
nathanrussell-moj-digital, line 18: As per my previous comment re: whether the numbers part is random enough? 🤷‍♂️

nathanrussell-moj-digital (approved):
LGTM 👍 
A couple of minor comments, but nothing I cant live with 😁 👍

nathanrussell-moj-digital (approved):
👍 LGTM

PR #630, opened 2025-03-18
added workaround for app insights issue

stevomcallister:
As discussed in [here](https://mojdt.slack.com/archives/CTLQ5TCNN/p1739353601712389) there is an issue with [App insights java](https://github.com/microsoft/ApplicationInsights-Java/issues/4079) which causes this error to be displayed in the logs. This will be fixed at some point but they recommend putting the full connection details and not just the instrumentation key - which is the change in this PR.


nathanrussell-moj-digital (approved):
LGTM 👍 

Out of interest, I dont recall doing this specifically in the UI, but it looks like we do, so agree it's good to be consistent 👍 
https://github.com/ministryofjustice/hmpps-education-and-work-plan-ui/blob/main/helm_deploy/hmpps-education-and-work-plan-ui/values.yaml#L54

PR #631, opened 2025-03-18
Fix for null date

stevomcallister:
update so that the contextual item COMPLETED_INDUCTION_CONDUCTED_IN_PERSON_DATE is not added when the completedDate is null.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/AsyncInductionEventService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/AsyncInductionEventService.kt
@@ -46,10 +46,9 @@ class AsyncInductionEventService(
       prisonId = createdAtPrison,
       actionedBy = induction.createdBy!!,
       timestamp = induction.createdAt!!,
-      contextualInfo = mapOf(
+      contextualInfo = mutableMapOf(
         TimelineEventContext.COMPLETED_INDUCTION_ENTERED_ONLINE_AT to createdAt.toString(),
         TimelineEventContext.COMPLETED_INDUCTION_ENTERED_ONLINE_BY to userService.getUserDetails(createdBy!!).name,
-        TimelineEventContext.COMPLETED_INDUCTION_CONDUCTED_IN_PERSON_DATE to completedDate.toString(),
         TimelineEventContext.COMPLETED_INDUCTION_NOTES to noteContent,
         *conductedBy
           ?.let {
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/AsyncInductionEventService.kt
```diff
@@ -58,7 +57,11 @@ class AsyncInductionEventService(
               TimelineEventContext.COMPLETED_INDUCTION_CONDUCTED_IN_PERSON_BY_ROLE to conductedByRolePerson,
             )
           } ?: arrayOf(),
-      ),
+      ).apply {
+        completedDate?.let {
+          this[TimelineEventContext.COMPLETED_INDUCTION_CONDUCTED_IN_PERSON_DATE] = it.toString()
+        }
+      },
```
nathanrussell-moj-digital, line 64: I can see your change works 👍 , but I was thinking something along these lines perhaps:
```
      contextualInfo = mapOf(
        TimelineEventContext.COMPLETED_INDUCTION_ENTERED_ONLINE_AT to createdAt.toString(),
        TimelineEventContext.COMPLETED_INDUCTION_ENTERED_ONLINE_BY to userService.getUserDetails(createdBy!!).name,
        *conductedBy
          ?.let {
            val noteContent = note?.content ?: ""
            val conductedByPerson = conductedBy ?: ""
            val conductedByRolePerson = conductedByRole ?: ""
            arrayOf(
              TimelineEventContext.COMPLETED_INDUCTION_CONDUCTED_IN_PERSON_BY to conductedByPerson,
              TimelineEventContext.COMPLETED_INDUCTION_CONDUCTED_IN_PERSON_BY_ROLE to conductedByRolePerson,
              TimelineEventContext.COMPLETED_INDUCTION_CONDUCTED_IN_PERSON_DATE to completedDate.toString(),
              TimelineEventContext.COMPLETED_INDUCTION_NOTES to noteContent,
            )
          } ?: arrayOf(),
```
where it's all done in the `*conductedBy` spread operator thingy ....

My thinking is that the `*conductedBy` spread operator thingy was used to say "add these fields to the contextual info if conductedBy is set" , and the things we want to add in this case are the contextual info fields all relating to who conducted the induction and their notes

I wonder if this approach makes it a little neater/clearer to read in the future?

(I dont know if what I've proposed works, or even compiles, but you get the idea 😁 )

src/main/resources/db/migration/common/V2025.03.18.0001__fix_null_event_context.sql
```diff
@@ -0,0 +1,3 @@
+DELETE FROM timeline_event_contextual_info
```
nathanrussell-moj-digital, line 1: Good shout; yes, we defo want to do this 👍 

On the assumption that we move the notes field into the conditional block (my other comment), what do you think about also deleting the `COMPLETED_INDUCTION_NOTES` record where the value is empty string?

I know we support no/empty notes; but at the moment where the feature toggle is not enabled, there shoudnt be any records set this way anyway 🤷

nathanrussell-moj-digital (approved):

PR #636, opened 2025-03-21
Timeline filter fixes RR-1321 and RR-1322

stevomcallister:
Basically was trying to be too smart for my own good with the original version. The prisonId or eventsSince filters didn't work unless there was one of the boolean filters (goals, reviews etc) set.

Went back to the drawingboard and the process is now:

If eventSince is set then create a list of all events after that date otherwise include all. 
if prisonId is set then filter the previous list on prisonId
lastly apply the boolean filters. 

Also added specific tests for the conditions that were failing. 

This code can probably be refined further but I quite like the do this then this then this logic rather than some complex filters that can easily miss conditions if the testing isn't 100%.


nathanrussell-moj-digital (approved):
LGTM 👍

PR #643, opened 2025-03-24
check ETL endpoint

stevomcallister:


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/ScheduleEtlController.kt
```diff
@@ -146,6 +147,38 @@ class ScheduleEtlController(
     return response
   }
 
+  @ResponseStatus(HttpStatus.CREATED)
+  @PreAuthorize(HAS_EDIT_REVIEWS)
+  @GetMapping(value = ["/action-plans/schedules/etl-check/{prisonId}"])
```
nathanrussell-moj-digital, line 152: Minor, but is it also worth changing line 85 back to a GET? (not that I mind, but seeing as you are editing this class anyway 🤷‍♂️ )

nathanrussell-moj-digital (approved):
LGTM 👍

PR #648, opened 2025-03-27
Demo prisoner-search-api bug

nathanrussell-moj-digital:
## Background
PR to demo the apparent bug in prisoner-search-api when calling the `/prisoner-search/prison/$prisonId` endpoint as a paged operation.

## Context
We need to call prisoner-search-api to get a list of all prisoners in a given prison. We previously called the API endpoint with a page size of 250 as it felt like an appropriate balance between number of requests vs. response data size vs. "spamming" prisoner-search-api
We found intermittent problems with this and after a bit of debugging in our client and service methods felt that the problem was with prisoner-search-api not returning the correct prisoners when calling it as a paged operation.
[This PR](https://github.com/ministryofjustice/hmpps-education-and-work-plan-api/pull/607) worked around it for us, in that we changed our call to a single non-paged call, but this from the description explains our thinking:

> Given the prisoners A, B, C, D, E, F, G, H, I and J; if you were to call the paged API with a page size of 3 you might expect responses:
> 1: [A, B, C]
> 2: [D, E, F]
> 3: [G, H, I]
> 4: [J]
> The order is irrelevant, just as long as you get all records, and they are only represented once.
>
> What we suspect is happening is that we get back responses such as:
> 1: [A, J, D]
> 2: [E, D, F]
> 3: [B, C, I]
> 4: [H]
> where D is returned twice, and G is not returned at all.

## Failing test
This PR adds a test that demonstrates the problem.
To run the test:
* Uncomment the test annotation `@Disabled` (it is disabled because other wise it would run as part of the CI checks for the branch and would fail!)
* Add an appropriate client ID and secret for in `application

### DIFF
--- a/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/PrisonerSearchApiClientTest.kt
+++ b/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/PrisonerSearchApiClientTest.kt
@@ -1,0 +1,93 @@
+package uk.gov.justice.digital.hmpps.educationandworkplanapi.app
+
+import mu.KotlinLogging
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Disabled
+import org.junit.jupiter.api.RepeatedTest
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.test.context.ActiveProfiles
+import org.springframework.test.context.DynamicPropertyRegistry
+import org.springframework.test.context.DynamicPropertySource
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.client.prisonersearch.Prisoner
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.client.prisonersearch.PrisonerSearchApiClient
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.testcontainers.LocalStackContainer
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.testcontainers.LocalStackContainer.setLocalStackProperties
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.testcontainers.PostgresContainer
+
+private val log = KotlinLogging.logger {}
+
+/**
+ * Test class to demonstrate possible problem with prisoner-search-api when calling it with paged requests
+ */
+@SpringBootTest
+@ActiveProfiles("demo-prisoner-search-api-paging-bug")
+class PrisonerSearchApiClientTest {
+
+  companion object {
+    private const val PRISON_ID = "WDI"
+
+    private val postgresContainer = PostgresContainer.instance
+
+    private val localStackContainer = LocalStackContainer.instance
+
+    @JvmStatic
+    @DynamicPropertySource
+    fun configureTestContainers(registry: DynamicPropertyRegistry) {
+      postgresContainer?.run {
+        registry.add("spring.datasource.url", postgresContainer::getJdbcUrl)
+        registry.add("spring.datasource.username", postgresContainer::getUsername)
+        registry.add("spring.datasource.password", postgresContainer::getPassword)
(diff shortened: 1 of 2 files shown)


src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/PrisonerSearchApiClientTest.kt
```diff
@@ -0,0 +1,93 @@
...
+        registry.add("spring.flyway.password", postgresContainer::getPassword)
+      }
+      localStackContainer?.also { setLocalStackProperties(it, registry) }
+    }
+  }
+
+  @Autowired
+  private lateinit var prisonerSearchApiClient: PrisonerSearchApiClient
+
+  @Disabled
```
nathanrussell-moj-digital, line 54: Comment out this annotation to run the test

I have it set as a repeated test because the original problem was intermittent, but WDI seems to be failing every time at the moment

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/PrisonerSearchApiClientTest.kt
```diff
@@ -0,0 +1,93 @@
...
+  @RepeatedTest(100)
+  fun `test paged operation`() {
+    // First call prisoner-search-api to get all prisoners in BXI as a single page
+    val allPrisonersInOneCall = getAllPrisonersInPrison(PRISON_ID, 9999).map { it.prisonerNumber }
```
nathanrussell-moj-digital, line 58: This is basically our test control sample - we call the prisoner-search-api with a large page size, so it has the effect of calling it once as a non-paged operation.
It returns a list (which could contain duplicates)
To ensure all elements in the list are unique we assert that the list size is equal to the size of the list as a set.

This is our happy path / control sample.

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/PrisonerSearchApiClientTest.kt
```diff
@@ -0,0 +1,93 @@
...
+
+    assertThat(allPrisonersInOneCall.size)
+      .describedAs { "The number of elements in the returned list should equal the number of elements when turning the list into a set, proving a unique set of prisoners being returned" }
+      .isEqualTo(allPrisonersInOneCall.toSet().size)
+
+    // Now get all prisoners via a series of paged calls
+    val allPrisonersViaPagedCall = getAllPrisonersInPrison(PRISON_ID, 250).map { it.prisonerNumber }
```
nathanrussell-moj-digital, line 65: Now we call the same API endpoint for the same prison, but as a paged operation; 250 records at a time.

As before we get a list of records, and the first thing we do is assert this size of this list (from the paged operation) is the same size as the list from the non-paged operation.
This seems to pass every time 👍

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/PrisonerSearchApiClientTest.kt
```diff
@@ -0,0 +1,93 @@
...
+
+    assertThat(allPrisonersViaPagedCall.size).isEqualTo(allPrisonersInOneCall.size)
+
+    assertThat(allPrisonersViaPagedCall.size)
```
nathanrussell-moj-digital, line 69: The bit that fails is this assertion, that is asserting the size of the paged list is equal to the size of the paged list as a set (ie. are there any duplicates in the list)
This is failing every time at the moment for WDI, and would seem to suggest that calling prisoner-search-api as a paged operation can return duplicates somehow

src/integrationTest/resources/application-demo-prisoner-search-api-paging-bug.yml
```diff
@@ -0,0 +1,86 @@
+prisoner-search-api:
+  client:
+    id: < prisoner-search-api-id >
```
nathanrussell-moj-digital, line 3: To run the test, put a client ID and secret in here

PR #676, opened 2025-04-28
Set updated at prison when a goal is archived/unarchived/completed

stevomcallister:
Following the previous update to the API spec for archived/unarchived/completed goals, this PR sets the updated at prison when a goal is archived, unarchived or completed.


nathanrussell-moj-digital (approved):
LGTM 👍 Nice

PR #680, opened 2025-05-07
Initial PR to check wiring in dev

stevomcallister:
PR to process inbound assessment events from Meganexus. Currently only logs the message so the tests don't test anything except that the process doesn't fail. At least the framework is there for when we do add more functionality.

### DIFF
--- a/helm_deploy/hmpps-education-and-work-plan-api/values.yaml
+++ b/helm_deploy/hmpps-education-and-work-plan-api/values.yaml
@@ -69,6 +69,10 @@ generic-service:
       HMPPS_SQS_QUEUES_EDUCATIONANDWORKPLAN_QUEUE_NAME: "sqs_queue_name"
     education-and-work-plan-domain-events-sqs-dl-instance-output:
       HMPPS_SQS_QUEUES_EDUCATIONANDWORKPLAN_DLQ_NAME: "sqs_queue_name"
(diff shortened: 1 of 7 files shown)


helm_deploy/hmpps-education-and-work-plan-api/values.yaml
```diff
@@ -69,6 +69,10 @@ generic-service:
       HMPPS_SQS_QUEUES_EDUCATIONANDWORKPLAN_QUEUE_NAME: "sqs_queue_name"
     education-and-work-plan-domain-events-sqs-dl-instance-output:
       HMPPS_SQS_QUEUES_EDUCATIONANDWORKPLAN_DLQ_NAME: "sqs_queue_name"
+    education-and-work-plan-assessments-queue:
```
nathanrussell-moj-digital, line 72: I'm assuming the CPE stuff (new queue?) has been rolled out in preprod and prod? Otherwise we'll need to be careful about promoting other API changes forward into preprod and prod?

nathanrussell-moj-digital (approved):
👍 LGTM

helm_deploy/hmpps-education-and-work-plan-api/values.yaml
stevomcallister: yes it's in preprod and prod - much easier to do them all at once

PR #689, opened 2025-05-27
added message publisher for induction messages

stevomcallister:
Probably won't need this one very often but this is an endpoint to generate induction schedule messages in the event we have to manually adjust data that has gone out of sync.


nathanrussell-moj-digital (approved):
Looks like a good idea. Nice 👍

PR #690, opened 2025-05-27
Added data correction script

stevomcallister:


src/main/resources/db/postgres/prod/V2025.05.27.0001__Induction-schedule-corrections-RR-1519.sql
```diff
@@ -0,0 +1,9 @@
+update induction_schedule set schedule_status = 'COMPLETED', updated_at = now(), updated_by = 'STEPHENMCALLISTER_GEN' where prison_number = 'A1092FF';
```
nathanrussell-moj-digital, line 1: I wonder if `updated_by` (in all of these update and inserts) should be `system` rather than yourself (and perhaps a comment explaining why we are doing this?)

The reason I suggest `system` is that I had a conversation with Catriona Reader last week re: how she needs to query the schedule data , and what the data means etc. And part of that was that any schedule record that was `COMPLETED` would have an `updated_by` value of a real user (because it was the user who completed the induction/review in the UI) and they should be counted in her stats/reports.
But any schedule record that is `COMPLETED` that has an `updated_by` value of `system` was the subject of one of our data correct scripts/ETLs, and she should not count them.

What do you think?

PR #712, opened 2025-07-04
change review calculation window rules

stevomcallister:


nathanrussell-moj-digital (commented):
LGTM, but do you think we need a couple of unit tests that specifically have a release date AND each of the 3 sentence types?
I think it would be good to codify the intent; and will mean that if anyone changes the order it should break some tests?

What do you think?

PR #730, opened 2025-09-16
Allow toPrisonId to be null

stevomcallister:


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/client/prisonapi/PrisonApiMapper.kt
```diff
@@ -39,14 +39,16 @@ class PrisonApiMapper {
     return admissionsAndReleases
   }
 
-  private fun buildTransfers(transfers: List<TransferDetail>): List<PrisonMovementEvent> = transfers.map {
+  private fun buildTransfers(transfers: List<TransferDetail>): List<PrisonMovementEvent> = transfers.map { t ->
```
stevomcallister, line 42: It is possible for a transfer to not have a dateInToPrison or a toPrisonId, however from the limited evidence I have there should be a dateOutOfPrison. 

the PrisonMovementEvent.date is the date the transfer in to prison happened but since there is no transfer in only out then I'm setting the date to be the date the out transfer happens only when the transferInDate is null.

PR #893, opened 2026-04-10
RR-2510 - ETL to correct inductions and reviews

nathanrussell-moj-digital:


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/ScheduleEtlController.kt
```diff
@@ -28,48 +22,40 @@ private val log = KotlinLogging.logger {}
...
+  private val inductionScheduleService: InductionScheduleService,
 ) {
-  @ResponseStatus(HttpStatus.OK)
+
+  @PostMapping("/action-plans/schedules/reschedule-inductions-following-transfer")
+  @ResponseStatus(HttpStatus.CREATED)
   @PreAuthorize(HAS_EDIT_REVIEWS)
-  @GetMapping(value = ["/action-plans/schedules/etl-check/{prisonId}"])
   @Transactional
-  fun checkSchedulesForPrisonersInPrison(
```
nathanrussell-moj-digital, line 43: I dont think we need this ETL endpoint anymore? I think it was to check schedules of a previous ETL fix we had to do? Is that right @stevomcallister ? Safe to delete this endpoint now?

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/ScheduleEtlController.kt
```diff
@@ -28,48 +22,40 @@ private val log = KotlinLogging.logger {}
...
-    @PathVariable("prisonId") prisonId: String,
-  ): CheckSchedulesEtlResponse {
-    log.info("Check ETL process for prison ID: $prisonId")
-
-    val allPrisoners = prisonerSearchApiService.getAllPrisonersInPrison(prisonId).also {
-      log.info("Total prisoners in prison $prisonId: ${it.size}")
+  fun rescheduleInductionsFollowingTransfer(
+    @RequestBody prisonNumbersRequest: PrisonNumbersRequest,
+  ) {
```
nathanrussell-moj-digital, line 38: New API endpoint `/action-plans/schedules/reschedule-inductions-following-transfer` to fix Induction schedules following a prisoner transfer (specifically for this bug) - I'm planning on a similar endpoint for reviews but I've not started that one yet.

What this does is to take a list of PRNs from the request body, and for each one:
* Check it is still a problem / suitable to fixing via this ETL by:
  * getting the induction schedule history in reverse order (newest first)
  * check the first (most recent) record is SCHEDULED and it's deadline date is before the buggy feature went live (ie. overdue)
  * check the immediately previous record is EXEMPT_PRISONER_TRANSFER
* call the existing service method to exempt and reschedule due to transfer

My thinking is that:
*  we should do the check first to make sure the induction is still in a state by which it needs to be fixed (we know some users are doing people's inductions and/or setting manual exemptions to get them off their Overdue lists; and this fix wont work for those cases)
* Use the same service method that the regular listener uses to exempt and reschedule due to transfer, which will set the correct dates (based on "today" being the date of transfer, but Oli has already agreed that 👍 ), and it will send the messages to MN 👍 

My only worry is that it will look like they've had a second transfer but to the same prison! (ie. it will exempt due to transfer, and then re-schedule it). I suspect we'll need to explain this to MN !

Does this sound sensible? @stevomcallister ?

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/ScheduleEtlController.kt
```diff
@@ -28,48 +22,40 @@ private val log = KotlinLogging.logger {}
...
-
-    log.info("ETL check process completed for prison ID: $prisonId. Response: ${response.prisonersWithoutPLPData.size}")
-    return response
   }
 
-  data class PrisonNumbersRequest(
-    val prisonNumbers: List<String>,
-  )
-
   @PostMapping("/action-plans/schedules/publish-review-messages")
```
nathanrussell-moj-digital, line 96: Just as a side note, and not one for this PR, but these ETL methods to re-publish the messages have proved to be quite handy in the past 👍 
I'm thinking about making these less "temporary", and moving them into an "admin" controller or similar, writing tests for them, and documenting them in the swagger spec etc

One for another day though

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/ScheduleEtlController.kt
stevomcallister: yeah not needed

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/ScheduleEtlController.kt
stevomcallister: This does sound a sensible approach - ie let the code reprocess it applying the rules as if it was done 'today' 

MN will get a new set of EXEMPT_PRISONER_TRANSFER and then SCHEDULED 'messages' BUT they will not know it's for the same prison per se. They could work that out from the updated at prison I guess. But in either case I don't think they really care. They only take note of the deadline date and then ultimately compare that to when the induction is finally complete in order to calculate the KPI.

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/ScheduleEtlController.kt
stevomcallister: yeah there is also the ReviewRepairController which you may want to remove OR perhaps look at it for part 2 of this fix as that endpoint is all about updating review deadline dates...

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/EducationAssessmentBackfillTest.kt
nathanrussell-moj-digital, line 1: I have also deleted the "education assessment back fill" ETL controller, so this test for it is also deleted

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/EducationAssessmentBackfillController.kt
nathanrussell-moj-digital, line 1: Deleted the "education assessment back fill" ETL controller

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/ReviewRepairController.kt
nathanrussell-moj-digital, line 1: Deleted the previous "review repair" ETL controller

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/AsyncInductionScheduleEventService.kt
```diff
@@ -38,19 +38,20 @@ class AsyncInductionScheduleEventService(
...
-        timelineEventFactory.inductionScheduleStatusUpdatedEvent(this),
-      )
-    } catch (e: DataIntegrityViolationException) {
-      log.warn(e) { "DataIntegrityViolationException occurred, retrying recordTimelineEvent once more for prisoner [$prisonNumber]" }
-      // Retry once
-      timelineService.recordTimelineEvent(
-        prisonNumber,
-        timelineEventFactory.inductionScheduleStatusUpdatedEvent(this),
-      )
+    run {
```
nathanrussell-moj-digital, line 41: This logic was missing in the `AsyncReviewScheduleEventService` (below), and when I went to add it as per this class, I found it was wrong. So I have improved it in both classes (though I'm still not 100% happy with it - further improvements for another day).

There is a race condition that the previous try/catch block was trying to protect against, and it is because these event classes are async. If you had 2 things that called this method for the same prisoner in very quick succession, there is a race condition creating the timeline. This is particular prevalent in the code that deals with schedules, because typically a schedule can be created then updated in quick succession, and each change to the schedule will call this method to record the timeline, telemetry and send MN messages.
The call to `timelineService.recordTimelineEvent(...)` tries to create the prisoner's timeline if one does not already exist, before adding the timeline event. In the case of the prisoner not having a timeline to begin with there is a race condition where multiple calls to this method can attempt to create the timeline. And there is a DB constraint in place to ensure a prisoner only has 1 timeline. I suspect this only happened in tests .... but ... we should still protect against it.

The previous code attempted to record the timeline event event in a try block. If that failed the catch block would attempt it again.
But the 2nd attempt in the catch block could still fail, and _that_ call is not in a try block (its in the catch). In this case the entire method fails and the calls to the telemetry service and event publisher do not happen. Worse still!, because it's async, the method throws the exception, but the main thread knows nothing of it so carries on.
I saw this A LOT when writing the integration tests for the new ETL endpoints.

Anyway ... I have improved it by using a  `run` block with a `repeat` block. The `repeat` is set to 3 (though this should probs be a const somewhere. 3 seemed OK/sufficient for the tests to pass). Immediately after the call to `timelineService.recordTimelineEvent(...)` within the `try` block,  we return to the outer `run` (ie. `return@run`) which breaks us out of the `repeat`
I guess I could have done it with a `while` loop of some kind, but I would have had to track both the count and success 🤷‍♂️ . This approach seemed more "kotlin"; there's probably not much in it 🤷‍♂️

src/test/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/EducationAssessmentBackfillControllerTest.kt
nathanrussell-moj-digital, line 1: I have also deleted the "education assessment back fill" ETL controller, so this test for it is also deleted

stevomcallister (approved):

PR #915, opened 2026-05-20
RR-2532 - Refactor how InductionEntity has it's modified timestamp updated

nathanrussell-moj-digital:
PR to change how the `InductionEntity` gets it's updated_at field updated.

Under normal circumstances, where JPA has updated the `InductionEntity` (and the entity is therefore dirty), fields like `updated_at` will automatically get updated by virtue of the hibernate lifecycle annotation on the field (eg: `@UpdateTimestamp`)

There are scenarios though where a child entity is updated (eg: `FutureWorkInterestsEntity`). When the `FutureWorkInterestsEntity` is persisted by JPA we also need the parent `InductionEntity` to have it's `updated_at` field is updated. This is because even though they are all separate entities, as far as the business is concerned they are all part of the Induction, and we need to track when the Induction was updated.

Until now this was done via a method on the `InductionEntity` that updates the `updated_at` field with `Instant.now()`, and that method was called as part of the persistence adapter when it updates anything to do with the Induction:
```
  fun updateLastUpdatedAt() {
    updatedAt = Instant.now()
  }
```

We are moving to use a `Clock` instance in all calls to `.now()` ... and injecting a `Clock` bean into an entity class is not possible. So we have to use a different approach to update this entity.

This approach that this PR implements is actually a much better solution - not sure why we didnt do it this way first time! 🤣 

This PR injects the spring `AuditingHandler` into the persistence adapter, and then uses that to mark the `InductionEntity` as dirty (instead of calling the `updateLastUpdatedAt()` method). The `updateLastUpdatedAt()` method is deleted.
This all works by virtue of marking the entity as dirty. JPA will commit the entity because it is dirty. And as part of committing it, it will update the `updated_at` field via the `@UpdateTimestamp` annotation 👍


.snyk
```diff
@@ -0,0 +1,14 @@
+# WARNING - THIS FILE WAS GENERATED BY THE hmpps-gradle-spring-boot GRADLE PLUGIN
```
nathanrussell-moj-digital, line 1: Added this file at the same time.
It's nothing to do with this PR, but a previous bump to one of the HMPPS spring boot dependencies now generates this file as part of the build.
We need to commit the file at some point, so doing it as part of this PR seems sensible.

ayoMoj (approved):

PR #916, opened 2026-05-21
RR-2532 - Refactor GoalEntity and StepEntity away from using ParentEntity and KeyAwareEntity

nathanrussell-moj-digital:
This PR is a stepping stone / enabler to get us to the point where we can refactor how `GoalEntity` gets its `updated_at` field set (currently it uses a method which uses `.now()`, and we want to move away from that)

_This_ PR doesn't change that (it still uses `.now()`) but it does some important refactoring which will allow us to make that change in the next PR.

---

Back when we first wrote LWP, one of the original design approaches was that `GoalEntity` and `StepEntity` knew about each other via the `ParentEntity` and `KeyAwareEntity` superclasses. These superclasses basically defined a relationship between the parent and child, and visa versa.
We then had generic mapper classes that operated on instances of `ParentEntity` and `KeyAwareEntity` (as the parent and child) and had methods to add children to the parent, and for the children to be able to update the `updated_at` field of the parent.
Because we felt there would be several cases of this parent/child relationship, and that they would need to operate on each other in this way, we defined these superclasses and used generics etc; with the view that we would use it in a number of places. The reality is that we used it in a very small number of places, and that it is overly complicated for what it delivers.

This PR unpicks all of that (in the context of `GoalEntity` and `StepEntity - we might need to do it separately for other entity types in a subsequent PR). The code is now much simpler, and will allow changing how `GoalEntity` has its `updated_at` field set in my next PR.


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/entity/actionplan/GoalEntity.kt
```diff
@@ -80,10 +79,17 @@ data class GoalEntity(
   @LastModifiedBy
   var updatedBy: String? = null
 
-  override fun childEntityUpdated() {
+  fun childEntityUpdated() {
```
nathanrussell-moj-digital, line 82: This will be replaced in the next PR, which will use the `AuditingHandler` to dirty the entity (and therefore update the updated_at field), rather than this approach

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/actionplan/GoalEntityMapper.kt
```diff
@@ -127,4 +121,45 @@ class GoalEntityMapper(
     GoalStatusEntity.ARCHIVED -> GoalStatusDomain.ARCHIVED
     GoalStatusEntity.COMPLETED -> GoalStatusDomain.COMPLETED
   }
+
+  private fun updateExistingSteps(
```
nathanrussell-moj-digital, line 125: These methods are essentially what were originally in `GoalEntityListManager`

src/testFixtures/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/entity/actionplan/GoalEntityAssert.kt
```diff
@@ -41,14 +41,6 @@ class GoalEntityAssert(actual: GoalEntity?) : AbstractObjectAssert<GoalEntityAss
     return this
   }
 
-  fun isEqualToComparingAllFieldsExceptNotes(expected: GoalEntity): GoalEntityAssert {
```
nathanrussell-moj-digital, line 44: These assertion methods are no longer used/relevant, so I've removed them (comparing the entity "except Notes" was always a bit weird, and was actually masking a different problem which has been fixed in the mapping test changes in this PR)

src/testFixtures/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/entity/actionplan/GoalEntityBuilder.kt
```diff
@@ -20,6 +20,7 @@ fun aValidGoalEntity(
   updatedBy: String? = "bjones_gen",
   archiveReason: ReasonToArchiveGoal? = null,
   archiveReasonOther: String? = null,
+  notes: String? = null,
```
nathanrussell-moj-digital, line 23: This was essentially why the above assertion class had a method to "compare except Notes" ... it's because the test data builder didnt previously support setting the Notes field! 🤣

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/actionplan/GoalEntityListManager.kt
nathanrussell-moj-digital, line 1: The original idea/design was that `GoalEntity` and `StepEntity` knew about each other via the `ParentEntity` and `KeyAwareEntity` superclasses. These superclasses basically defined a relationship between the parent and child, and visa versa.

This class contained methods to add steps to a goal, delete steps from a goal, and edit inc re-order steps within a goal.
But ... it was written using generics and the `ParentEntity` and `KeyAwareEntity` classes, which meant that removing `ParentEntity` and `KeyAwareEntity` classes (which ultimately is what this PR is doing) meant that this class no longer compiled.
Given that it was used nowhere else, the simplest solution was to move the methods into `GoalEntityMapper` and reference `GoalEntity` and `StepEntity` rather than trying to be clever with generics etc.

See `GoalEntityMapper` below for the methods that have been moved into it

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/entity/actionplan/StepEntity.kt
```diff
@@ -76,12 +74,10 @@ data class StepEntity(
     parent?.childEntityUpdated()
   }
 
-  override fun associateWithParent(parent: ParentEntity) {
-    this.parent = parent as GoalEntity
+  fun associateWithParent(parent: GoalEntity) {
```
nathanrussell-moj-digital, line 77: In a few PRs time (once we've changed how the parent GoalEntity get it's updated_at field updated), I _think_ we'll be able to remove this `parent` field in the child StepEntity, and just use a regular collection on the parent GoalEntity.
I _think_ we'll be able to do that ... but we'll know for sure in a few PRs time ....

src/test/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/mapper/actionplan/GoalEntityMapperTest.kt
nathanrussell-moj-digital, line 1: This is a large diff, but essentially it now contains all of the tests that were originally in `GoalEntityMapperUpdateEntityFromDtoTest` below 👇

ayoMoj (approved):

PR #917, opened 2026-05-21
RR-2532 - Refactor how GoalEntity has it's modified timestamp updated

nathanrussell-moj-digital:
PR to change how the `GoalEntity` gets it's updated_at field updated.

Under normal circumstances, where JPA has updated the `GoalEntity` (and the entity is therefore dirty), fields like `updated_at` will automatically get updated by virtue of the hibernate lifecycle annotation on the field (eg: `@UpdateTimestamp`)

There are scenarios though where the child entity `StepEntity` is updated. When the `StepEntity` is persisted by JPA we also need the parent `GoalEntity` to have it's `updated_at` field is updated. This is because even though they are all separate entities, as far as the business is concerned they are all part of the Goal, and we need to track when the Goal was updated.

Until now this was done via a method on the `GoalEntity` that updates the `updated_at` field with `Instant.now()`, and that method was called as part of the JPA lifecycle methods of the `StepEntity` being updated
```
  fun childEntityUpdated() {
    updatedAt = Instant.now()
  }
```

We are moving to use a `Clock` instance in all calls to `.now()` ... and injecting a `Clock` bean into an entity class is not possible. So we have to use a different approach to update this entity.

This approach that this PR implements is actually a much better solution - not sure why we didnt do it this way first time! 🤣 

This PR injects the spring `AuditingHandler` into the persistence adapter, and then uses that to mark the `GoalEntity` as dirty (instead of calling the `childEntityUpdated()` method). The `childEntityUpdated()` method is deleted.
This all works by virtue of marking the entity as dirty. JPA will commit the entity because it is dirty. And as part of committing it, it will update the `updated_at` field via the `@UpdateTimestamp` annotation 👍

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/entity/actionplan/StepEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/entity/actionplan/StepEntity.kt
@@ -10,9 +10,6 @@ import jakarta.persistence.GeneratedValue
 import jakarta.persistence.Id
 import jakarta.persistence.JoinColumn
 import jakarta.persistence.ManyToOne
-import jakarta.persistence.PrePersist
-import jakarta.persistence.PreRemove
-import jakarta.persistence.PreUpdate
 import jakarta.persistence.Table
 import org.hibernate.Hibernate
 import org.hibernate.annotations.CreationTimestamp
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/database/jpa/entity/actionplan/StepEntity.kt
```diff
@@ -67,13 +64,6 @@ data class StepEntity(
   @JoinColumn(name = "goal_id")
   var parent: GoalEntity? = null
 
-  @PrePersist
-  @PreUpdate
-  @PreRemove
-  fun onChange() {
-    parent?.childEntityUpdated()
```
nathanrussell-moj-digital, line 74: This is why the `StepEntity` class had to know about the parent. I _think_ this is the only use case for it, so now that this is deleted I _think_ we can delete the `parent` field (here, and in the DB etc). But thats one for another day ....

ayoMoj (approved):

PR #924, opened 2026-05-26
RR-2626 - Default `isIndeterminateSentence` and `isRecall` to false when missing from Prisoner Search response

ayoMoj:
The `defaultValue` annotation on `Boolean` fields in `Prisoner` DTO was not correctly handling cases where the corresponding JSON fields (`indeterminateSentence`, `recall`) were entirely missing from the Prisoner Search API response.

This change updates the `Prisoner` data class to use Kotlin default parameter values, ensuring that `isIndeterminateSentence` and `isRecall` are correctly defaulted to `false` during deserialization if they are omitted from the upstream payload. New unit and integration tests have been added to verify this behavior.


nathanrussell-moj-digital (approved):
LGTM 🙌

PR #943, opened 2026-06-04
RR-1464 - Schedule Induction on admission when assessments already complete (PES)

ayoMoj:
Closes the out-of-order gap where a prisoner's "all relevant S&As completed"
message is received before they have any Induction Schedule. Previously that
message no-op'd (no schedule to update), and a later admission created a PENDING
schedule that would never be scheduled.

Now, on admission, after creating the Induction Schedule we schedule it straight
away if the prisoner's S&As are already complete (today + 10 days); otherwise it
is left pending as before. The same check is applied via a shared helper for both
the newly-created and the existing-PENDING schedule cases. Dormant under PEF.


nathanrussell-moj-digital (approved):
LGTM 👍

PR #946, opened 2026-06-08
RR-1464: Add PES integration tests for remaining induction-scheduling

ayoMoj:
Adds integration tests (test-only, no production change) for PES scenarios whose behaviour already works but was not yet covered end-to-end:

- Prison admission with no Induction Schedule and no S&A completion record creates a schedule with status PENDING_INITIAL_SCREENING_AND_ASSESSMENTS_FROM_CURIOUS.
- S&A completed message for a prisoner with no Induction Schedule records the event but does not create a schedule (out-of-sequence messages).
- A duplicate S&A completed message is idempotent on the Induction Schedule: it is scheduled once (today + 10) and the second message is a no-op as it is no longer pending.


nathanrussell-moj-digital (approved):
LGTM 👍

PR #947, opened 2026-06-09
RR-1464: Re-gate automatically-exempt Induction on the prisoner's S&As on admission (PES)

ayoMoj:
Under the PES contract, when a prisoner with an automatically-exempt Induction Schedule (release, transfer, temporary absence, hospital, merge) is re-admitted, the Induction is re-gated on their Screening & Assessments rather than rescheduled straight to SCHEDULED: it is set back to PENDING_INITIAL_SCREENING_AND_ASSESSMENTS_FROM_CURIOUS, then scheduled immediately (today + 10 days) if their S&As are already complete.

Gated on the PES contract (ciag-kpi-processing-rule), so PEF behaviour is unchanged: an automatically-exempt schedule on admission continues to be rescheduled as today. Adds InductionScheduleStatus.isAutomaticExemption to distinguish system-applied exemptions from manual ones, and integration tests for both the S&As-complete and S&As-not-complete cases.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/PrisonerReceivedIntoPrisonEventService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/PrisonerReceivedIntoPrisonEventService.kt
@@ -1,6 +1,7 @@
 package uk.gov.justice.digital.hmpps.educationandworkplanapi.app.messaging
 
 import mu.KotlinLogging
+import org.springframework.beans.factory.annotation.Value
 import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.domain.learningandworkprogress.induction.InductionScheduleAlreadyExistsException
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/PrisonerReceivedIntoPrisonEventService.kt
```diff
@@ -49,7 +50,10 @@ class PrisonerReceivedIntoPrisonEventService(
   private val scheduleAdapter: ScheduleAdapter,
   private val educationAssessmentEventService: EducationAssessmentEventService,
   private val clock: Clock,
+  @param:Value("\${ciag-kpi-processing-rule:PEF}") private val ciagKpiProcessingRule: String,
 ) {
```
ayoMoj, line 54: PES-gated as to not change production behaviour. Auto-exempt schedule aslo exist under PEF (released prisoner re-admitted), and PEF must keep rescheduling them to SCHEDULED

nathanrussell-moj-digital (approved):
LGTM 👍

PR #950, opened 2026-06-11
fix: make JVM heap config effective and right-size memory to mitigate search OOMs

ayoMoj:
The prison search endpoint (GET /search/prisons/{prisonId}/people) was throwing java.lang.OutOfMemoryError under load. Investigation showed two compounding issues.

1. The intended `-Xmx512m` was never applied. The Docker ENTRYPOINT is exec-form, which does not expand $JAVA_OPTS, and the `java` binary does not read JAVA_OPTS. The live JVM ran with MaxHeapSize=250Mi (container default of 25% of the 1000Mi limit) — roughly half the intended heap. The pod that OOMed peaked at only ~848Mi RSS, well under the 1000Mi container limit, confirming it ran out of JVM heap rather than container memory. Moved the setting to JAVA_TOOL_OPTIONS (read natively by the JVM) at -Xmx768m.

   Right-sized the container from the live JVM: non-heap committed is a large, stable ~700Mi (metaspace ~250Mi, ~290 thread stacks, ~240Mi reserved code cache, App Insights agent native, Netty direct buffers). With a 768m heap that puts worst-case RSS at ~1.45Gi, so the limit is raised 1000Mi -> 1800Mi and the request to 1024Mi (matching observed ~950Mi steady-state RSS). No HeapDumpOnOutOfMemoryError: main already sets -XX:+ExitOnOutOfMemoryError so a dump on the ephemeral fs is lost on restart and risks an ephemeral-storage eviction — capturing dumps needs an emptyDir volume (follow-up).

2. The endpoint paginates entirely in memory — it loads the whole prison roster (pageSize=9999) and discards all but one page. This is the underlying design flaw and the real per-request heap consumer. Left TODO markers documenting the proper fix (push filtering/sorting/pagination down to prisoner-search + the DB) rather than attempting that larger, higher-risk change in this hotfix.

Mitigation only; the proper paginated-search rework is tracked by the TODOs.

### DIFF
--- a/helm_deploy/hmpps-education-and-work-plan-api/values.yaml
+++ b/helm_deploy/hmpps-education-and-work-plan-api/values.yaml
@@ -23,9 +23,36 @@ generic-service:
   # Used to access resources like SQS queues and SNS topics
   serviceAccountName: hmpps-education-and-work-plan
 
+  # Container resources.
+  # Sized from the live prod JVM (jcmd on the highest pod): non-heap committed is large and steady —
+  # ~250Mi metaspace, ~150-290Mi thread stacks (~290 threads), ~240Mi reserved code cache, plus the
+  # Application Insights agent's native memory and Netty direct buffers — giving a ~700Mi non-heap
+  # baseline before any heap. With the now-effective 768m heap (see JAVA_TOOL_OPTIONS) worst-case RSS is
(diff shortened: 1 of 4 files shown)


rickchoijd (approved):
LGTM

PR #956, opened 2026-06-16
RR-2462 - Move SAR Report Generation Test to specialtests

nathanrussell-moj-digital:
PR to move `SarReportGenerationTest` into `specialtests`

This change is an enabler for the work we are about to get into re: LWP SAR changes, and splitting it out into its own PR just make life easier.

We made this same change a while ago in SAN (moving `SarReportGenerationTest` into `specialtests`). The reason we had to do it in SAN was because the SAN SAR report included the timestamps at which certain things were created (IE. our standard createdAt/updatedAt audit fields).
In the SAN `SarReportGenerationTest` we had a problem because every time the test ran it setup new data/records which were created "now". Therefore the comparison document that the SAR test uses was always out of date, and the test would always fail.
We fixed this in SAN by using a spring context that uses a fixed clock.
There are a few other integration tests that need to use a slightly different spring context for various reasons. We refer to these as "special tests", and live in the `specialtests` package.

Whilst the LWP SAR report does not currently contain any audit timestamps, so we've never had this problem in LWP, we always knew it was coming.
Ticket RR-2462 is to add the Employability Skills data to the SAR report, and as part of that we need to include the createdAt audit fields. When we do that we will have the same problem that we had in SAN (because the createdAt timestamp will be different on every run)

We _could_ move these files into `specialtests` with a fixed clock when we do the main work in RR-2462, but it would make for a larger PR, and would be difficult to explain and reason about this part of the change. Therefore I'm making this change now as an enabler for the main work.


src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/config/ClockConfig.kt
```diff
@@ -5,21 +5,21 @@ import org.springframework.context.annotation.Bean
...
 import org.springframework.context.annotation.Primary
 import java.time.Clock
-import java.time.LocalDate
+import java.time.LocalDateTime
 import java.time.ZoneId
 
 @Configuration
 class ClockConfig(
-  @param:Value("\${app.clock.fixed-date:}") private val fixedDate: String?,
+  @param:Value("\${app.clock.fixed-date-time:}") private val fixedDateTime: String?,
```
nathanrussell-moj-digital, line 13: Updated the test clock config so that it sets a fixed datetime, rather than just a fixed date
(the changes to the SAR report to include Employability Skills will need time precision)

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/specialtests/resource/sar/SarReportGenerationTest.kt
```diff
@@ -20,6 +21,7 @@ import java.time.LocalDate
 // Run test with env var SAR_GENERATE_ACTUAL=true to generate actual SAR report content for the test data created in this test,
 // which can be used to verify the format and content of the report generated by the Subject Access Request service.
 
+@ActiveProfiles("integration-test", "fixed-clock")
```
nathanrussell-moj-digital, line 24: Move the existing test into the `specialtests` package, and use the `fixed-clock` profile

src/integrationTest/resources/application-fixed-clock.yml
```diff
@@ -0,0 +1,3 @@
+app:
```
nathanrussell-moj-digital, line 1: Here's the application.yml for the fixed-clock profile.
It doesnt really matter what the fixed timestamp is; just that it is a fixed known value.

ayoMoj (approved):

PR #960, opened 2026-06-17
RR-2682 - Added 'Completed induction' section to the SAR report

nathanrussell-moj-digital:
PR to add the 'Completed induction' section to the SAR report as per the design
See attached PDF for example report
[sar-generated-report.pdf](https://github.com/user-attachments/files/29042807/sar-generated-report.pdf)


src/integrationTest/resources/sar/expected-report.html
```diff
@@ -259,10 +291,6 @@ <h3>Review</h3>
     <td class="data-column-30">Deadline date</td>
     <td>10 August 2025</td>
   </tr>
-  <tr>
```
nathanrussell-moj-digital, line 262: Removed 'note type' from the Reviews section as agreed with Aliki, as the Note Type will always be `REVIEW` for reviews, so this adds no value to the prisoner at all
(noteType is an internal field that is used to link and categories notes to their parent entity - it has never been a user facing field)

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/resource/mapper/induction/InductionResourceMapper.kt
```diff
@@ -77,6 +78,12 @@ class InductionResourceMapper(
         personalSkillsAndInterests = personalSkillsAndInterests?.let { skillsAndInterestsMapper.toPersonalSkillsAndInterestsResponse(it) },
         futureWorkInterests = futureWorkInterests?.let { workInterestsMapper.toFutureWorkInterestsResponse(it) },
         employabilitySkills = employabilitySkills.map { employabilitySkillsResourceMapper.fromModelToResponse(it) },
+        inductionCompletionMetadata = InductionCompletionMetadataResponse(
```
nathanrussell-moj-digital, line 81: Update the `InductionResponseMapper` function to map in the new meta data fields that describe the induction completion data (who did it, job role, notes etc)

src/main/resources/static/openapi/EducationAndWorkPlanAPI.yml
```diff
@@ -1215,6 +1215,8 @@ components:
           description: A List of employability skills.
           items:
             $ref: '#/components/schemas/GetEmployabilitySkillsResponse'
+        inductionCompletionMetadata:
```
nathanrussell-moj-digital, line 1218: Update the `InductionResponse` object in the swagger spec to add in a new `InductionCompletionMetadataResponse` property, which describes the induction completion data (who did it, job role, notes etc)

Updating the swagger spec like this means that the SAR endpoint in the swagger spec will accurately describe the SAR endpoint response, including this new property/fields.

Updating the swagger spec like this (adding new properties to a response object) is a non-breaking change for the UI, so we do not need to reimport the swagger spec into the UI project. The UI will happily ignore the new property/fields

ayoMoj (approved):

PR #967, opened 2026-06-24
RR-2682 - Add a Ticking Clock for time sensitive tests

nathanrussell-moj-digital:
PR to add a "ticking clock" to the integration test config for tests that are sensitive to the time / rely on timestamp ordering.

`MillisecondTickingClock` is an implementation of `Clock` which we inject into the Spring context anyway.

Up until now we had the concept of a fixed clock that could be used in integration tests. As the name suggests it was fixed and the timestamp never changed. We use this in the tests that test the behaviour over Christmas 2025, though strictly speaking that clock does not need to be fixed, it just needs to set the date to a known fixed date.

Anyway, we also use the fixed clock in the SAR Report Generation tests.This is more important that we use a fixed clock here, because the generated report contains the date and time (to second precision) for when entities where created - eg:
Screenshot 2026-06-24 at 08 30 59
The entity timestamps are generated using the injected `Clock`, which is why we use a fixed clock for these integration tests - the fixed clock is set at `2026-03-21T09:43:27.126`

In respect of the SAR tests, this approach works fine ... at the moment. But what it means is that all entities created as part of the test setup (Induction + all child entities, Induction Schedules + histories, Goals, Steps, Review Schedules + histories, timeline events etc) ALL get created with the creation and update timestamps of `2026-03-21T09:43:27.126`
This works for the SAR Report Generation test as it currently stands, because the "Created At" / "Updated At" timestamps will all be `21 March 2026, 9:43:27 am` in the generated report.

But for entries on the report that rely on being sorted by timestamp, the resultant order on the report will be indeterminate (because they all have exactly the same timestamp), which could result in a fragile test. 
This is particular true of the piece of work I am about to start, which is to add the Review Schedules to the report. There will be several Review Schedules in the report, and they need to be correctly ordered by date.

This PR is an enabler for that work. It replaces the fixed clock with a 'ticking' clock. The ticking clock is initialised with a known date and time (`2026-03-21T09:43:27.000`), and for every call to get the current time (calls to `.instant()`), it increments the millisecond value. Therefore it will return:

* 2026-03-21T09:43:27.000
* 2026-03-21T09:43:27.001
* 2026-03-21T09:43:27.002
* 2026-03-21T09:43:27.003
* etc

regardless of the actual time, or how long has passed between calls to get the current time.
Therefore every entity created will have a different, sequential and incremental, timestamp.

The difference in timestamp is down to millisecond level, and because our flyway schema defines timestamp columns as 3 digit millisecond precision, it means a given test can effect up to 999 calls to get the current time before the current time rolls over to the next second.
For most of our tests this will be fine - we dont really care when things were created, and certainly not precise to the second.
But the SAR Report Generation test does render times precise to the second. As long as the SAR test does not create/update more that 999 entities, this will be fine.


src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/config/ClockConfig.kt
```diff
@@ -5,21 +5,61 @@ import org.springframework.context.annotation.Bean
...
+import java.time.Duration
+import java.time.Instant
 import java.time.LocalDateTime
 import java.time.ZoneId
+import java.time.temporal.ChronoUnit
 
 @Configuration
 class ClockConfig(
-  @param:Value("\${app.clock.fixed-date-time:}") private val fixedDateTime: String?,
+  @param:Value("\${app.ticking-clock.initial-date-time:}") private val initialDateTime: String?,
```
nathanrussell-moj-digital, line 17: Change the clock config for the integration tests to setup a ticking clock, rather than a fixed clock

src/integrationTest/resources/sar/expected-report.html
```diff
@@ -889,9 +889,9 @@ <h3>Employability skill</h3>
 <h2>Goals</h2>
 <h3>Goal</h3>
 <table class="summary-list">
-  <tr><td>Title</td><td>Goal 1</td></tr>
+  <tr><td>Title</td><td>Goal 3</td></tr>
```
nathanrussell-moj-digital, line 892: This change to the expected report actually demonstrates how the use of the fixed clock was wrong (for the SAR test), and the ticking clock is better.
Previously all 3 goals were recorded as being created at the exact same timestamp - `2026-03-21T09:43:27.126`
The API returns goals sorted by creation date, but because they all had the same timestamp the return order was indeterminate. Conveniently for the test they are returned in a consistent order of Goal 3, Goal 2, Goal 1; but strictly speaking that order is incorrect. Even though the execution order of the test setup method creates them in the order of Goal 1, Goal 2, Goal 3; they were not being returned in that order.

Now that we have a ticking clock, each goal is created with an incremental timestamp. The creation timestamp of these goals will be somewhere within the time/second 09:43:27. We dont know or care what the millisecond value is, but they will now be in sequence. Therefore the goals are now being returned by the API in the correct order of Goal 1, Goal 2, Goal 3

ayoMoj (dismissed):

ayoMoj (approved):

PR #970, opened 2026-06-25
RR-2682 - Fix SAR report test

nathanrussell-moj-digital:
Fix SAR report test by using the clock in the date calculation implementation classes (which I clearly forgot to do in the earlier PR that introduced the `Clock` as a spring bean)
Basically a couple of the "schedule date calculation" classes were still using `.now()` to generate the current date/time, rather than using an injected clock. 
This caused a problem in the SAR test because the Review Schedule dates were being calculated based on "today" rather than based on the `TickingClock` (which we use to generate predictable dates and times! 🤣 )

Therefore the bug fix for the SAR report test is actually in the main implementation code, specifically in the "schedule date calculation" classes, rather than SAR code itself

### DIFF
--- a/domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/induction/service/InductionScheduleDateCalculationService.kt
+++ b/domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/induction/service/InductionScheduleDateCalculationService.kt
@@ -3,6 +3,7 @@ package uk.gov.justice.digital.hmpps.domain.learningandworkprogress.induction.se
 import uk.gov.justice.digital.hmpps.domain.learningandworkprogress.induction.InductionSchedule
 import uk.gov.justice.digital.hmpps.domain.learningandworkprogress.induction.InductionScheduleStatus
 import uk.gov.justice.digital.hmpps.domain.learningandworkprogress.induction.dto.CreateInductionScheduleDto
+import java.time.Clock
 import java.time.LocalDate
 import java.time.ZoneId
 
@@ -14,6 +15,7 @@ import java.time.ZoneId
  * The only exception to this are the abstract methods which are intended to be implemented by subclasses.
  */
 abstract class InductionScheduleDateCalculationService(
+  private val clock: Clock,
   private val scheduleDateNotBefore: LocalDate? = null,
   private val propertiesProvider: InductionSchedulePropertiesProvider,
 ) {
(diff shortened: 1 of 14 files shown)


domain/learningandworkprogress/src/main/kotlin/uk/gov/justice/digital/hmpps/domain/learningandworkprogress/induction/service/InductionScheduleDateCalculationService.kt
```diff
@@ -14,6 +15,7 @@ import java.time.ZoneId
  * The only exception to this are the abstract methods which are intended to be implemented by subclasses.
  */
 abstract class InductionScheduleDateCalculationService(
+  private val clock: Clock,
```
nathanrussell-moj-digital, line 18: Basically we need to inject and use the `Clock` in these "schedule date calculation" classes. I'd forgotten to do that the other week when we added the `Clock` bean 😬

src/integrationTest/resources/sar/expected-report.html
```diff
@@ -549,7 +549,7 @@ <h3>Schedule</h3>
   </tr>
   <tr>
     <td class="data-column-30">Schedule calculation rule</td>
-    <td>BETWEEN_6_AND_12_MONTHS_TO_SERVE</td>
+    <td>BETWEEN_12_AND_60_MONTHS_TO_SERVE</td>
```
nathanrussell-moj-digital, line 552: The expected SAR report has a few changes; not only the dates (now they are based on the injected `TickingClock` rather than "today"), but also the calculation rules (because the dates that derive the rules are now "correct")

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/service/TimelineEventFactory.kt
```diff
@@ -43,13 +43,15 @@ import uk.gov.justice.digital.hmpps.domain.timeline.TimelineEventType.INDUCTION_
...
+import java.time.Clock
+import java.time.Instant
 import java.util.UUID
 
 /**
  * Responsible for identifying which [TimelineEvent]s have occurred, for example following an update to a [Goal].
  */
 @Component
-class TimelineEventFactory(private val userService: ManageUserService) {
+class TimelineEventFactory(private val clock: Clock) {
```
nathanrussell-moj-digital, line 54: Not obviously related to the SAR report (but it indirectly is) .... I needed to also inject and use the `Clock` in the timeline event data records

rickchoijd (approved):
LGTM

PR #977, opened 2026-07-10
RR-2785 - Set PES as the CIAG KPI rule for integration tests

nathanrussell-moj-digital:
Set PES as the CIAG KPI rule for integration tests. Set up separate integration tests for PEF

### DIFF
--- a/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/PrisonerReceivedDueToAdmissionEventTest.kt
+++ b/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/PrisonerReceivedDueToAdmissionEventTest.kt
@@ -12,6 +12,8 @@ import org.junit.jupiter.params.provider.CsvSource
 import uk.gov.justice.digital.hmpps.domain.randomValidPrisonNumber
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.IntegrationTestBase
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.client.prisonersearch.aValidPrisoner
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.educationassessment.EducationAssessmentEventEntity
+import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.educationassessment.EducationAssessmentEventStatus.ALL_RELEVANT_ASSESSMENTS_COMPLETE
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.induction.InductionScheduleCalculationRule.EXISTING_PRISONER
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.database.jpa.entity.induction.InductionScheduleStatus
 import uk.gov.justice.digital.hmpps.educationandworkplanapi.app.messaging.AdditionalInformation.PrisonerReceivedAdditionalInformation.Reason.ADMISSION
@@ -44,7 +46,7 @@ class PrisonerReceivedDueToAdmissionEventTest : IntegrationTestBase() {
 
   // New prison admission, prisoner has never had a PLP
   @Test
-  fun `should create new Induction Schedule given prisoner does not already have an Induction`() {
+  fun `should create new Induction Schedule given prisoner does not already have an Induction and has not had their Screenings and Assessments`() {
     // Given
     val prisonNumber = randomValidPrisonNumber()
     with(aValidPrisoner(prisonerNumber = prisonNumber, prisonId = "MDI")) {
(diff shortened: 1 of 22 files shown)


src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/PrisonerReceivedDueToAdmissionEventTest.kt
```diff
@@ -116,6 +120,111 @@ class PrisonerReceivedDueToAdmissionEventTest : IntegrationTestBase() {
...
+    with(aValidPrisoner(prisonerNumber = prisonNumber, prisonId = "MDI")) {
+      createPrisonerAPIStub(prisonNumber, this)
+    }
+
+    assertThat(runCatching { getInduction(prisonNumber) }.getOrNull()).isNull()
+    assertThat(runCatching { getInductionSchedule(prisonNumber) }.getOrNull()).isNull()
+
+    val earliestDateTime = OffsetDateTime.now()
+
+    educationAssessmentEventRepository.save(
```
Mary-ford, line 136: I think this implies that the prisoner has had their S&A completed, and then we get an event saying they have been admitted to prison.  I would have expected the S&A would happen after we get the admission event? How does this scenario happen?

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/specialtests/PrisonerReceivedDueToAdmissionEventDuringChristmasTest.kt
```diff
@@ -82,13 +89,119 @@ class PrisonerReceivedDueToAdmissionEventDuringChristmasTest : IntegrationTestBa
...
+        }
+
+      val inductionScheduleEvents = inductionScheduleEventQueue.receiveEventsOnQueue(QueueType.INDUCTION)
+      assertThat(inductionScheduleEvents)
+        .hasNumberOfEvents(2)
+        .allEvents {
+          it.hasDetailUrl("http://localhost:8080/inductions/$prisonNumber/induction-schedule")
+            .hasNumberOfPersonReferenceIdentifiers(1)
+            .personReferenceIdentifier(1) { it.hasValue(prisonNumber) }
+        }
```
Mary-ford, line 204: Should this send an outbound event to notify that the induction schedule has been created?

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/PrisonerReceivedDueToAdmissionEventTest.kt
nathanrussell-moj-digital: Brilliant question, thats exactly what it means, and yes your expectation of "normal" events is correct 👍 

> I think this implies that the prisoner has had their S&A completed, and then we get an event saying they have been admitted to prison

...
> How does this scenario happen?

It's basically an edge case, but technically is possible.
All of our processing in this area is message driven. We receive and process events from NOMIS/Domain Events telling us stuff has happened about the prisoner (such as admissions), and we also get messages from Curious saying that the S&As are complete.

Messaging is asynchronous. There is no guarantee when we'll get a message, or in what order messages will arrive (but common sense / logic states that we _should_ get the admission message from NOMIS before the S&A message from Curious)
But sometimes things don't always go to plan:
* Perhaps NOMIS/Domain Events didnt send the admission message at all? (rare, and would be an exceptional case, also also MN listen to the same event so in theory would not have setup the prisoner record at their end)
* Perhaps LWP failed to process the admission event for some reason. Maybe we retried it 3 times, and then it landed on the DLQ
* Maybe LWP entirely was down for a weekend and nobody knew until Monday. During that time NOMIS/Domain Events sent the admission message, and then at some point later Curious sent the S&A message. Both messages are sat on the LWP queues (a different queue for each), but not processed because LWP is down. Then on Monday when we restart LWP, the messages get processed ... but the order of receipt and processing by LWP is not guaranteed.
* and probably some other weird and whacky scenarios ...

It's technically possible, but is very much an edge case. Having a test around it at least shows we've thought about it and have a predictable outcome in this scenario.

> I would have expected the S&A would happen after we get the admission event?

100% yes, that is the "normal" sequence of events that we would expect 👍

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/specialtests/PrisonerReceivedDueToAdmissionEventDuringChristmasTest.kt
nathanrussell-moj-digital: Yes, thats what one of the 2 messages are. The first is sent when the `InductionSchedule` was created with a status of `PENDING_INITIAL_SCREENING_AND_ASSESSMENTS_FROM_CURIOUS` and the 2nd was sent when the `InductionSchedule` was updated with a status of `SCHEDULED`

The assertions just above this on the `InductionScheduleHistory` assert that there are 2 `InductionSchedules` and that they have the correct statuses/deadline etc.

The assertions here to do with the message sent to MN is the right / appropriate way to do it.
The semantics of the message sent to MN is that every time an `InductionSchedule` is changed in any way (inc. initial creation), we send a message to MN saying "the prisoner's InductionSchedule has changed; come and get the history records if you want the detail"
The message itself does not contain anything about the InductionSchedule, so we cant write assertions here to that effect.
All we can really do is assert that the correct number of messages was sent for the prisoner in question.

Mary-ford: Oh yes, sorry.  I'd missed that this was doing that check. Makes sense, thanks.

Mary-ford (approved):
LGTM

PR #992, opened 2026-07-22
RR-2767: 17 day rule adjustment

ayoMoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/PrisonerReceivedIntoPrisonEventService.kt
```diff
@@ -196,19 +194,25 @@ class PrisonerReceivedIntoPrisonEventService(
...
+    if (scheduleAdapter.isInductionComplete(nomsNumber)) {
       val prisoner = prisonerSearchApiService.getPrisoner(nomsNumber)
-      if (prisoner.releaseDate != null && prisoner.releaseDate.isAfter(LocalDate.now(clock).plusDays(16))) {
-        reviewScheduleService.handle17DayTransferRule(
-          prisonNumber = nomsNumber,
-          prisonTransferredTo = prisonId,
-          releaseDate = prisoner.releaseDate,
-        )
+      val releaseDate = prisoner.releaseDate
+      if (releaseDate != null) {
```
nathanrussell-moj-digital: What happens here if there is no release date? Whilst it might sound odd, that is very much a valid scenario (though I wish the data from prisoner-search was more structured/clear in this regard)
A prisoner can have no release date for a few reasons, such as being on remand, or having an indeterminate sentence (I think there are some other cases as well)

Can you check with Aliki what is expected to happen if the prisoner has no release date?

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/PrisonerReceivedEventDueToTransferTest.kt
```diff
@@ -1133,6 +1133,127 @@ class PrisonerReceivedEventDueToTransferTest : IntegrationTestBase() {
...
+    @Test
+    fun `prisoner who never had a review but has completed their induction is transferred with 17 or more days left to serve new review schedule should be created`() {
+      // Given
+      // Induction is completed within the last 3 months so no initial Review Schedule is ever created (RR-2711 scenario)
+      val prisonNumber = setUpRandomPrisoner(releaseDate = LocalDate.now().plusDays(60))
+      createInduction(prisonNumber, aValidCreateInductionRequestForPrisonerNotLookingToWork(prisonId = ORIGINAL_PRISON))
+      createActionPlan(prisonNumber)
+
+      // the prisoner has no review schedule at all and has never had a review
+      assertThat(reviewScheduleRepository.getAllByPrisonNumber(prisonNumber)).isEmpty()
```
nathanrussell-moj-digital: Can you use the REST API methods for this instead of direct calls to the repository please?
Having the tests interacting directly with the repository is quite a tight coupling that I'd rather we didnt have. In LWP we deliberately have persistence adapters to abstract the main service code away from the repositories. And in SAN we currently have a hidden problem that I am unpicking where we are making an assertion against a table/repository that we want to delete (we want to delete the table, but not the concept of what the table is storing; we are storing that data elsewhere)

Anyway, I'd much rather we didnt use the repositories like this in tests. I know we have in some places, but we do already have the helper methods that call the REST API in the IntegrationTestBase class for these, and we also have the relevant assertJ classes, so I'd much rather we use them instead.

```
val reviewSchedules = getReviewSchedules(prisonNumber)
assertThat(reviewSchedules).hasNumberOfReviewSchedules(0)
val reviews = getActionPlanReviews(prisonNumber)
assertThat(reviews).hasNumberOfCompletedReviews(0)
```

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/PrisonerReceivedEventDueToTransferTest.kt
```diff
@@ -1133,6 +1133,127 @@ class PrisonerReceivedEventDueToTransferTest : IntegrationTestBase() {
...
+      // When
+      sendDomainEvent(sqsMessage)
+
+      // Then
+      await untilCallTo {
+        domainEventQueueClient.countMessagesOnQueue(domainEventQueue.queueUrl).get()
+      } matches { it == 0 }
+
+      // a new review schedule is created with a deadline of today + 10 days
+      val reviewSchedulesAfter = reviewScheduleRepository.getAllByPrisonNumber(prisonNumber)
```
nathanrussell-moj-digital: As above, please can we use the REST API via the helper methods in IntegrationTestBase, plus the assertJ assertions 🙏

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/PrisonerReceivedEventDueToTransferTest.kt
```diff
@@ -1133,6 +1133,127 @@ class PrisonerReceivedEventDueToTransferTest : IntegrationTestBase() {
       val schedule = reviewSchedulesAfter.findLast { it.scheduleStatus == SCHEDULED }
       assertThat(schedule!!.latestReviewDate).isEqualTo(LocalDate.now().plusDays(10))
     }
+
+    @Test
+    fun `prisoner who never had a review but has completed their induction is transferred with 17 or more days left to serve new review schedule should be created`() {
+      // Given
+      // Induction is completed within the last 3 months so no initial Review Schedule is ever created (RR-2711 scenario)
```
nathanrussell-moj-digital: I'm struggling with these 2 lines a little bit ....
What is the relevance of 3 months? And where are we setting something up that sets a date or similar that aligns with 3 months?

And then on the next line ... what is the significance of a release date 60 days away? I know the test scenario is about the prisoner having more than 17 days left to serve, and yes, I would agree that 60 is greater than 17 🤣 , but is there some significance with this number?
Why did you use 60? Is the scenario the same and would it work the same way if you set it to 600 ?
Would 18 be a better number here? As it is clearly related to the scenario of "more than 17"

nathanrussell-moj-digital: In fact, the more I read the comment `// Induction is completed within the last 3 months so no initial Review Schedule is ever created (RR-2711 scenario)`, whilst I understand it's trying to add some context, I think its actually adding some unnecessary confusion.
I think the test method name describes it well (`prisoner who never had a review`), and the comment on line 1145 `the prisoner has no review schedule at all and has never had a review` describes it well.

I think the significance of this scenario is that the prisoner has had an induction, but did not get a review scheduled (because they were due for release within 3 months, but _that_ nuance is related to another part of the code and test. That nuance is nothing to do with transfers). When they transfer, if they 17 days or more left to serve, then we create a review schedule for them.
And I think the test method name and comment on L1145 describe this sufficiently 👍 

I wonder if we can help future readers of this test by removing the comment on L1140 altogether?

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/PrisonerReceivedEventDueToTransferTest.kt
```diff
@@ -1133,6 +1133,127 @@ class PrisonerReceivedEventDueToTransferTest : IntegrationTestBase() {
...
+      val prisonNumber = setUpRandomPrisoner(releaseDate = LocalDate.now().plusDays(60))
+      createInduction(prisonNumber, aValidCreateInductionRequestForPrisonerNotLookingToWork(prisonId = ORIGINAL_PRISON))
+      createActionPlan(prisonNumber)
+
+      // the prisoner has no review schedule at all and has never had a review
+      assertThat(reviewScheduleRepository.getAllByPrisonNumber(prisonNumber)).isEmpty()
+      assertThat(reviewRepository.getAllByPrisonNumber(prisonNumber)).isEmpty()
+
+      // the prisoner's release date has since moved into the past
+      wiremockService.stubGetPrisonerFromPrisonerSearchApi(
```
nathanrussell-moj-digital, line 1221: I slightly confused here.
I get it that you are setting the wiremock stub for this prisoner to set their release date in the past 👍  ... but ...
* Why does L1215 set a release date of 60 days? Should they not be the same?
* Why do the other tests not need to set the wiremock stub?
* Does `setUpRandomPrisoner` not already setup the wiremock stub internally? (which would answer why the other tests don't need to do this)

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/PrisonerReceivedIntoPrisonEventService.kt
```diff
@@ -196,19 +194,25 @@ class PrisonerReceivedIntoPrisonEventService(
       },
     )
 
-    val reviews = reviewService.getCompletedReviewsForPrisoner(nomsNumber)
```
nathanrussell-moj-digital, line 199: Is this right? Are we removing this functionality? or is it somehow covered in your new code?
Or does the new business rule effectively supersede this?

src/test/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/PrisonerReceivedIntoPrisonEventServiceTest.kt
nathanrussell-moj-digital, line 1: Can we make more use of this unit test class please? Otherwise its becoming a bit redundant 🤔 

I know you've written some scenarios as integration tests, but I'm a bit concerned with this approach (more so in SAN to be fair) that the integration tests are masquerading as unit tests. I think we are probably missing scenarios/combination/permutations (this is definitely true in SAN)

I know there is one (basic) unit test in this case for handling a basic transfer. I wasn't aware that we'd not been maintaining this and were not writing new unit tests as we've developed the service out, and instead put more (but not enough IMHO) tests as integration tests (thats my fault for not being on top of this)

We want to be more aligned with the testing triangle/pyramid, where we have lots of unit tests (because they are cheap and fast), and fewer integration tests

src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/PrisonerReceivedEventDueToTransferTest.kt
ayoMoj: Have removed the comment at L1140 and changed the number to 18

src/test/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/PrisonerReceivedIntoPrisonEventServiceTest.kt
ayoMoj: added unit tests to this class for the new transfer logic covering the permutations

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/PrisonerReceivedIntoPrisonEventService.kt
ayoMoj: Checked with Aliki, a prisoner with no release date should still get a review schedule at today + 10

src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/PrisonerReceivedIntoPrisonEventService.kt
ayoMoj: Superseded, not removed. The new isInductionComplete gate is a strict superset of the old preRelease one (you can't have a pre-release review without a completed induction), so everyone the old code caught is still caught, plus the RR-2711 case.

The only thing actually gone is the getCompletedReviewsForPrisoner lookup, which left reviewService unused so I removed it from the constructor.

nathanrussell-moj-digital (approved):
LGTM 👍

PR #995, opened 2026-07-27
RR-2767 apply 17day rule on transfer that already have a review schedule

ayoMoj:
RR-2767 applied the transfer "17 day rule" only to prisoners with no active review schedule. The 17-day check now runs on every transfer.


nathanrussell-moj-digital (approved):
LGTM 👍

PR #997, opened 2026-07-28
RR-2767: Apply 17 day rule to readmissions

ayoMoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/educationandworkplanapi/app/messaging/PrisonerReceivedIntoPrisonEventService.kt
```diff
@@ -193,10 +193,7 @@ class PrisonerReceivedIntoPrisonEventService(
     // isInductionComplete also covers the "prisoner already has an active Review Schedule" case.
     if (scheduleAdapter.isInductionComplete(nomsNumber)) {
       val releaseDate = prisonerSearchApiService.getPrisoner(nomsNumber).releaseDate
-      val today = LocalDate.now(clock)
-      val hasLessThan17DaysLeftToServe = releaseDate != null &&
-        !releaseDate.isBefore(today) &&
-        releaseDate.isBefore(today.plusDays(17))
+      val hasLessThan17DaysLeftToServe = hasLessThan17DaysLeftToServe(releaseDate)
```
nathanrussell-moj-digital, line 196: I'm not sure I fully understand this code (not this line, the whole stuff about handling reviews and the 17 day thing)
It's not that I don't understand the business requirement; it's that this class and the code within it seems to have run away with itself over time
What I mean by that is that _this_ class is `PrisonerReceivedIntoPrisonEventService` - it should be reasonably lightweight, doing not much more than looking at the event type/subtype, and delegating calls to specialised services.

We now have loads of logic in this class relating to creating/exempting/rescheduling reviews, and the nuances of the 17 day rule. I kinda feel that all of that logic should live in the ReviewService (ie. the business talk about the 17 day rule and whether reviews should be created/schedules etc . They don't talk about the 17 day rule in conjunction with a message handler/service 🤔 )

Don't worry about it for now; I think we can approve and merge this PR, with a view to perhaps refactoring this some other time. I think your integration test will help us and and when that refactoring is done.

But I think this class has grown out of its responsibilities, and should probably be refactored at some point 👍

nathanrussell-moj-digital (approved):
1 comment for some future refactoring, but happy to approve

PR #998, opened 2026-07-30
RR-2557 Add config for LWP DLQ alerting in dev

awatson-moj:
Added threshold for SQS message count alert and the names of the queues to monitor


rickchoijd (approved):
LGTM

PR #999, opened 2026-07-30
RR-2591 - Correct the role permission checks on the Employability Skills endpoints

nathanrussell-moj-digital:
This PR corrects the role permission checks on the Employability Skills endpoints

The steer we were given a long time ago from the HAAR team is that permission checks on API endpoints should be granular, and should take the form:

`ROLE_<service>__<resource>__[RO|RW]`

EG: 
`ROLE_EDUCATION_AND_WORK_PLAN__INDUCTIONS__RO`
`ROLE_EDUCATION_AND_WORK_PLAN__GOALS__RW`
etc

When we implemented the Employability Skills endpoint we clearly didnt do that!, and instead re-used the roles for the Action Plan resource!

This has gone unnoticed till now because the only client of the Employability Skills endpoint has been the LWP UI, which has all the roles anyway. IE. it already has the role `ROLE_EDUCATION_AND_WORK_PLAN__ACTIONPLANS__RW`

In RR-2591 we need Prisoner Profile to be able to retrieve Employability Skills via the GET endpoint. It does not currently have access because it does not have the `ROLE_EDUCATION_AND_WORK_PLAN__ACTIONPLANS__RO` role, and we dont want to give it that role because it will get access to too much.

We need this PR to update the endpoints so that they allow traffic from the existing ACTIONPLANS roles (for the current LWP client), and also the new EMPLOYABILITYSKILLS roles.
Once this PR is merged I can get HAAR to add the new EMPLOYABILITYSKILLS roles to both the LWP and Prisoner Profile clients.
And then we can come back to this PR and remove the ACTIONPLAN roles from the Employability Skills endpoint checks.


rickchoijd (approved):
LGTM

